PR #4, opened 2015-10-16
Refactoring

syohex:
- Use bound-and-true-p
- Refactoring sample code block

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,14 +10,14 @@
 * Tide is available in [melpa](http://melpa.org/#/tide). You can
   install tide via package-install <kbd>M-x package-install [ret] tide</kbd>
 
-````cl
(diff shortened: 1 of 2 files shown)


ananthakumaran:
Thanks for the PR. Why do you prefer +1 instead of t?

syohex:
> Thanks for the PR. Why do you prefer +1 instead of t?

`define-minor-mode` documentations says

```
The defined mode command takes one optional (prefix) argument.
Interactively with no prefix argument, it toggles the mode.
A prefix argument enables the mode if the argument is positive,
and disables it otherwise.
```

So it is better to use positive number instead of `t`.

ananthakumaran:
Thanks again.

PR #21, opened 2016-04-22
Support for unpatched tsserver, other minor changes.

jbms:
This makes the tsserver patch unnecessary, which makes it easier to use development versions of typescript.

### DIFF
--- a/tide.el
+++ b/tide.el
@@ -504,11 +504,12 @@ With a prefix arg, Jump to the type definition."
   (setq tide-buffer-dirty t))
 
 (defun tide-sync-buffer-contents ()
-  (setq tide-buffer-dirty nil)
-  (when (not tide-buffer-tmp-file)
-    (setq tide-buffer-tmp-file (make-temp-file "tide")))
-  (write-region (point-min) (point-max) tide-buffer-tmp-file nil 'no-message)
-  (tide-send-command "reload" `(:file ,buffer-file-name :tmpfile ,tide-buffer-tmp-file)))
+  (when tide-buffer-dirty
+    (setq tide-buffer-dirty nil)
+    (when (not tide-buffer-tmp-file)
+      (setq tide-buffer-tmp-file (make-temp-file "tide")))
+    (write-region (point-min) (point-max) tide-buffer-tmp-file nil 'no-message)
+    (tide-send-command "reload" `(:file ,buffer-file-name :tmpfile ,tide-buffer-tmp-file))))
 
 
 ;;; Auto completion
--- a/tide.el
+++ b/tide.el
@@ -77,6 +77,15 @@
   :group 'tide
   )
 
+(defcustom tide-tsserver-executable nil
+  "Name of tsserver executable to run instead of the bundled tsserver.
+This may either be a path or a name to be looked up in the PATH.
+Note that some versions of TypeScript may not be compatible with
+this package."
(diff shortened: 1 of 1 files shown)


ananthakumaran:
Thanks for taking time to write this PR. I just skimmed through the changeset. I had tried to use their event api https://github.com/ananthakumaran/tide/commit/e68eb9971715733bb8457f08c2d01ea32d16694b in the past, but gave up and decided to patch the tsserver because of the complexity introduced by it in tide. Event api breaks request -> response model. If you make two consecutive geterr request, you might get back two responses or one response. This model might work well if tide controls the error display by itself. But flycheck api design is bit different. flycheck will call tide whenever it decides to redo the error check and it expects the callback to be called, no matter what happens. Sometimes, flycheck will call the callback `tide-flycheck-start` even before the last callback has returned (may be the buffer content has changed since then). I am not sure your PR handles this situation. This is the reason I gave up last time. If we make another geterr request to tsserver, we might get back two or one event response back. We would have to use a timer or some other mechanism to decide which is the latest response.

In the end, I have decided to not use the event mechanism and just patch the tsserver.Yes, It is ugly, but less complex and less buggy. May be, you can propose a patch to add a immediate response type for geterr request.

jbms:
Yes, I did observe that the events that are received are a bit unpredictable --- if the file is modified while the diagnostics are pending, tsserver may cancel the request and not send any event, and I think also multiple requests while the diagnostics are still pending can also result in just a single event response.

However, my patch tries to take care of that, and appears to be robust in practice --- I've only been using it for several hours, but I haven't hit any cases where flycheck gets stuck.

It retries the geterr request every 2 seconds until both a syntaxDiag and semanticDiag event are seen for the file; if no event is seen after 10 seconds, it gives flycheck an error response.  Almost all of the time the events are received from the first request, and given the nature of flycheck a delay even of 2 seconds doesn't give a bad user experience.

Certainly upstreaming your patch would probably be the better solution, but until that patch makes it into a stable release of TypeScript, I think there is significant value in not tying tide to a particular version of typescript.

(There are also some independent patches in this pull request unrelated to flycheck.)

tide.el
```diff
@@ -849,35 +874,101 @@ number."
...
-         (tide-flycheck-send-response callback checker response)
-       (funcall callback 'errored (plist-get response :message))))))
+  (tide-command:geterr)
+  (setq tide--pending-flycheck-diagnostics-plist nil)
+  (setq tide--pending-flycheck (list :checker checker
+                                     :callback callback
+                                     :deadline (+ (float-time) tide-flycheck-timeout)
+                                     :timer (tide-flycheck-make-retry-timer)
+                                     ))
+  )
```
ananthakumaran: >      A syntax checker _must_ call CALLBACK at least once with a
>      STATUS that finishes the current syntax checker.  Otherwise
>      Flycheck gets stuck at the current syntax check with this
>      syntax checker.

I don't see any check for whether any pending callback is there. It just seems to overwrite the pending-flycheck, effectively losing the last callback reference if any.

ananthakumaran:
can you move the independent patches to a separate PR?

The trade off seems to between a complex implementation (that is very likely to be buggy because of the way event request-response works) that supports multiple typescript version or less buggy one but tied to the lastest version of typescript.

Given that most of the typescript releases so far is backward compatible, I am leaning towards the second option.

PR #22, opened 2016-04-22
Support unpatched tsserver

jbms:
This includes only the relevant changes.

Per your comment in pull request #21, I added code to handle the case of there already being a pending flycheck when tide-flycheck-start is called: it now calls the flycheck callback with 'interrupted, and properly cancels the active timer.  In fact per looking at flycheck.el the flycheck callback is a no-op if another checker has already been started, but calling it anyway with 'interrupted per the documentation may better handle future changes to flycheck.

### DIFF
--- a/tide.el
+++ b/tide.el
@@ -204,9 +204,15 @@ LINE is one based, OFFSET is one based and column is zero based"
         (apply (cdr callback) (list response)))
       (remhash request-id tide-response-callbacks))))
 
+(defun tide-dispatch-event (response)
+  (let ((event-type (intern (plist-get response :event))))
+    (case event-type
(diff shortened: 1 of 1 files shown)


martine:
A change like this is important for us at Google because we track different versions of the TypeScript services, so we have our users `(set-variable 'tide-tsserver-directory "/path/to/the/google-wide-typescript-dir")`.  This lets us ensure everyone is on the same version of TypeScript.

sandersn:
This would also be useful for developing TypeScript itself. I prefer to use a tsserver that I just built, eg `~/ts/built/local/`, so that I can test that changes work with the language service.

PR #26, opened 2016-05-10
Fixed conflict with typescript-tslint checker from Flycheck 27

Simplify:
As of Flycheck version 27 (on melpa and melpa-stable now), TSLint checker is built-in. This checker conflicts with tide-mode somehow, only errors from `typescript-tide` checker are shown in errors list. Only solution that I found working is to run `typescript-tslint` checker after `typescript-tide` checker.

### DIFF
--- a/tide.el
+++ b/tide.el
@@ -6,7 +6,7 @@
 ;; URL: http://github.com/ananthakumaran/tide
 ;; Version: 1.8.10
 ;; Keywords: typescript
-;; Package-Requires: ((dash "2.10.0") (flycheck "0.23") (emacs "24.1") (typescript-mode "0.1"))
+;; Package-Requires: ((dash "2.10.0") (flycheck "27") (emacs "24.1") (typescript-mode "0.1"))
 
 ;; This program is free software: you can redistribute it and/or modify it
 ;; under the terms of the GNU General Public License as published by
@@ -977,6 +977,7 @@ number."
   :predicate (lambda () (and tide-mode (tide-current-server) (not (file-equal-p (tide-project-root) tide-tsserver-directory)))))
 
 (add-to-list 'flycheck-checkers 'typescript-tide)
+(flycheck-add-next-checker 'typescript-tide '(warning . typescript-tslint) 'append)
 
 ;;; Utility commands
 


ananthakumaran:
Thanks for the PR. A couple of questions. I have been using a hand rolled tslint with the following config `(flycheck-add-next-checker 'typescript-tide '(warning . tslint))` for quite some time. 

Should we add a level field? and this can be moved to the checker definition itself. The :next-checkers seems to do the same thing.

Whose responsibility is it to setup the chaining properly? I don't want tide to do it unless it is the responsibility of the top-level checker.

Simplify:
I just added level in PR. According to the TSLint documentation ts file needs to compile in order to run TSLint. However, while adding TSLint into Flycheck, I tried everything to crash it. Even renaming HTML or Ruby files to ts did not crash TSLint.

Also I can't get `:next-checkers` working in the definition, however `flycheck-add-next-checker` works well.

Flycheck policy is that `syntax` checkers have higher priority than `lint` or `style` checkers. So, even without this problem with cleaning of errors list, `typescript-tide` have higher priority and must run before `typescript-tslint`.

This is temporally solution, until new checker chaining system in implemented in Flycheck. `:next-checkers` and `flycheck-add-next-checker` are going to be removed and new system will sort checkers automatically. IMO, `tide-mode` in only good place where this can be sorted out. Other option is that you add line with `flycheck-add-next-checker` to your setup instructions in README and not in the code as in this PR.

nrehm:
Sorry if i hijacks this but im currently hunting down a problem i have with tide and ts-lint.

Is it intended behavior that when you have tide errors tslint does not get executed and warnings do not show up in the flycheck window?
Even with this PR nothing changed for me.

It would be nice to have both as they tangle separate concerns.

Im using  spacemacs with this [https://github.com/syl20bnr/spacemacs/blob/develop/layers/+lang/typescript/packages.el](url) layer

@Simplify i could make tslint crash when there is no tslint.json available on project/home path and only later learned on your old project pre merge that i need to add one. 
Maybe catch this case with a message or ship with a sensible default config?
> Error from syntax checker typescript-tslint: End of file during parsing

Simplify:
@nrehm If you have old package `flycheck-typescript-tslint`, please remove it.

"Please ensure that the TypeScript source files compile correctly before running the linter." is stated on https://www.npmjs.com/package/tslint That is the reason why `typescript-tslint` is set not to run if there are compile errors detected by `typescript-tide`.

In Flycheck documentation is stated that you need `tslint.json` in order to use `typescript-tslint` checker as described on http://www.flycheck.org/en/latest/languages.html#typescript. That was the case before, but now, after your comment, I see that TSLint have default configuration since December last year, and that is used unless `tslint.json` is found. See https://github.com/palantir/tslint/blob/master/src/configuration.ts. Same config is generated if you run `tslint --init`. I just tested if linter works in Emacs without config, and that is the case. I had no crashes with TSLint without config, so my guess is that problem is related to your `ts` file. 

Please open an issue about `typescript-tslint` checker on Flycheck repo. Please specify version of TSLint that you are using, and ts file that creates problems (if possible). Is the file that crashes tslint with `.ts` extension?

ananthakumaran:
@Simplify I just merged #28 which adds support for js file. So basically next checker needs to be switched based on the current major mode. The only way I can think of is to define multiple checkers in tide like `typescript-tide`, `javascript-tide`, `web-mode-tide` etc and set up the next checkers appropriately. Let me know if you know any other simpler way.

Simplify:
@ananthakumaran Yes, that is best approach. There is a PR for adding `stylelint` into Flycheck that does similar thing, `flycheck-css-stylelint`, `flycheck-scss-stylelint` and `flycheck-less-stylelint`.

Convention is to use `flycheck-<language>-<checker-name>`. 

Current modes supported by JS checkers in Flycheck are: `js-mode js-jsx-mode js2-mode js2-jsx-mode js3-mode`. You can add those modes to `javascript-tide`. Also, errors from `tsserver` will be better than `ESLint` errors, they came from compiler, so you have two choices: or `:prepand` javascript-tide before `javascript-eslint` or explain that in README to users, they can then make that choice in Emacs init files or per project in `.dir-locals`. 

Not sure why you need with `web-mode`, it's not a programming language. Maybe you can add that in README, explaining to users that they can add `tide` to `web-mode` if they want that with:
```
(with-eval-after-load 'flycheck
  (flycheck-add-mode 'javascript-tide 'web-mode))`.
```

Personally I was never happy with `tern-mode`and #28 makes sense to me. 👍

ananthakumaran:
@Simplify  Thanks for the info, much appreciated. `web-mode` is the only emacs mode that I know which does a decent job in syntax highlighting `jsx` and `tsx` files. All the other modes break when they encounter xml tags.

PR #36, opened 2016-06-24
Update README.md

z4139jq:
Make tide-format-before-save local to typescript-mode to avoid Error: (void-function tide-format-before-save)

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,10 @@
 (setq company-tooltip-align-annotations t)
 
 ;; formats the buffer before saving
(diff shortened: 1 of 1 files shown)


ananthakumaran:
I guess adding autoload cookie would fix the problem https://github.com/ananthakumaran/tide/commit/46ae4a86f0960ae24ce402214dddf3a363d501f0. Could you test it?

z4139jq:
Yeah, it works, thanks

PR #38, opened 2016-06-25
Remove geterr.patch and add option to configure the tsserver executable

ananthakumaran:
This will be merged into master once TypeScript 2.0 is released. Thanks @jbms @sandersn. see #35 #22


jbms:
Great to see that your synchronous error interface got implemented in Typescript!  Thanks for pushing that forwarding so that tide can be used with other versions of TypeScript more easily.

PR #56, opened 2016-08-16
Make `tide-references` jump directly to a single reference

### LINKED ISSUE #55: Make `tide-references` jump directly to a single usage
If there is a single usage of a symbol, then `tide-references` should jump directly to it instead of creating a \*tide-references\* buffer. Then `tide-jump-to-definition` jumps back to the definition. For example:

```ts
/*1*/function usedOnce() { }
/*2*/usedOnce();
```

When the cursor is at (1), `tide-references` should jump to (2). When the cursor is at (2), `tide-jump-to-definition` should jump to (1). Otherwise (in case there is more than one usage or more than one declaration), \*tide-references\* should be created normally.

Note that this requires TypeScript 2.0, which adds `isDefinition` to `ReferenceEntry` to support this feature.  e1128b7 adds this feature if you want to take a look -- I'll create a PR for it soon.

sandersn:
If there is a single usage of a symbol, then tide-references jumps directly to it instead of creating a \*tide-references\* buffer. Then tide-jump-to-definition jumps back to the definition. For example:

```ts
/*1*/function usedOnce() { }
/*2*/usedOnce();
```

When the cursor is at (1), `tide-references` jumps to (2). When the cursor is at (2), `tide-jump-to-definition` jumps to (1). Otherwise (in case there is more than one usage or more than one declaration), \*tide-references\* is created normally.

Note that this requires TypeScript 2.0, which adds isDefinition to ReferenceEntry to support this feature.

Fixes #55

### DIFF
--- a/tide.el
+++ b/tide.el
@@ -778,12 +778,37 @@ number."
       (goto-char (point-min))
       (current-buffer))))
 
+(defun tide-is-identical-reference (original second)
+  (and (equal (plist-get original :file) (plist-get second :file))
+       (eq (tide-plist-get original :start :line) (tide-plist-get second :start :line))))
+(defun tide-find-single-usage (references)
+  (let ((definition nil)
+        (usage nil)
+        (multiple nil))
+    (-each references
+      #'(lambda (reference)
+          (if (eq t (plist-get reference :isDefinition))
+              (if (or (eq definition nil) (tide-is-identical-reference definition reference))
+                  (setq definition reference)
+                (setq multiple t))
+            (if (or (eq usage nil) (tide-is-identical-reference usage reference))
+                (setq usage reference)
+              (setq multiple t)))))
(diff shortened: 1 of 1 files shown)


ananthakumaran:
The second scenario you have mentioned(cursor at 2) doesn't work for me. The cursor just stays in reference and doesn't jump to the definition

sandersn:
Are you running `tide-references` at (2) or `tide-jump-to-definition`? For me, `tide-references` just prints the message "This is the only usage." at (2), but `tide-jump-to-definition` jumps back to (1). (I didn't change `tide-jump-to-definition` so I didn't expect it to be broken.)

Sorry for the delay -- I was busy with 2.0/2.1 work.

ananthakumaran:
yeah, looks like I used `tide-references` at (2). Merged. Thanks

PR #58, opened 2016-08-31
tide-import prototype

TatriX:
I've made a prototype for (auto) import functionality.
It has obvious flaws and just a prototype for a full featured support of autoimport.

Currently `tide-import` (bound to `C-c C-a` by default) works as follows:
It will ask a user for a search string (defaults to symbol at point) then asks tsserver for a list of possible occurrences of this symbol. Then it keeps only project related symbols and sorts them by match kind (exact match = good).

Suppose you have a code like this:
```typescript
import { foo} from "../bar";
log.Error(error);
```
but symbol `log` is not defined. If you call `tide-import` when point is at `log` it will search for global symbols containing "log" allowing you to select which one is correct and then add it  to the imports block of the file:
```typescript
import { foo} from "../bar";
import log from "../log";
log.Error(error);
```

Currently there is a lot of hardcoded things and I'm not even  sure it's a correct way to do autoimport but it works for me so I hope it will be useful as a start point in further tide development.

### DIFF
--- a/tide.el
+++ b/tide.el
@@ -35,6 +35,7 @@
 (require 'dash)
 (require 'flycheck)
 (require 'imenu)
+(require 'thingatpt)
 
 (defgroup tide nil
   "TypeScript Interactive Development Environment."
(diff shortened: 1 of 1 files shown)


ananthakumaran:
@TatriX thanks for your interest. It's a nice idea and I would really like this feature, but tide is not the best place to implement it.  You can try to get it added in TypeScript repo, once it is available there, it can be added in tide. Tide is just a thin UI wrapper around tsserver and I would like to keep it that way.

TatriX:
I think I can make it as a separate package for now.

KevinVillela:
Was there ever a separate package created for this, or does TSServer support it? VSCode and IntelliJ plugins both have auto-importing, it would be really nice to have it for Tide :)

Issue #61, opened 2016-09-22
Make `tide-references` behave like occur and compilation-mode


josteink:
When you use `occur` or compile things using `compilation-mode`, you get the results presented in a buffer which where you can navigate between the results using `previous-error` and `next-error` (even though not all things are errors).

I have these commands conveniently bound to F7 and F8 and it works pretty much for all things Emacs. All things except tide's `tide-references`.

Would this be possible to add?

ananthakumaran:
There are already couple of methods that provide navigation (bound to `n` & `p` by default). https://github.com/ananthakumaran/tide/blob/master/tide.el#L726. Could you use those?

josteink:
Those are *different* from what works everywhere else, so it's not really optimal. It also has the disadvantage that the buffer has to be open to use those keys.

`next-error` and `last-error` are global functions accessible from any buffer, meaning you can easily jump from one to the other, without having to go back to the Occur/Build/References-buffer to choose the next one.

It really improves the flow of code navigation.

I'll take a look at how other similar packages for other languages (Racer for Rust, Omnisharp for C#, etc) does this to make it work with the error-system for seamless navigation.

Maybe it's a simple fix, and in that case, I might send a patch back to fix this myself.

josteink:
Looking at `omnisharp-emacs` it seems it's just a matter of [creating a new buffer, use `compilation-mode` for it](https://github.com/OmniSharp/omnisharp-emacs/blob/master/src/omnisharp-utils.el#L37-L68), and ensure [your lines are formatted correctly](https://github.com/OmniSharp/omnisharp-emacs/blob/master/src/omnisharp-utils.el#L70-L87) to trigger the navigation effect.

The format is something we can define ourselves, or like in the case of `omnissharp-emacs`, just reuse an existing one.

It really doesn't seem like an impossible job, but it might break compatibility with the existing implementation. Any opinion? Should I prepare a patch? :)

josteink:
This is obviously not finished code, and it can use some refinements (and probably some cleanup if any obsolete code is left behind as well), but it shows how simple a thing like this can be.

https://github.com/josteink/tide/commit/1d050517e3cd8bd9ba8cfbe88ab6741b6b852866

This will leverage everything emacs already has for these purposes, and present the results in a more standard way.

Obviously the formatting will take some work, but you get the idea. What do you think?

josteink:
[...]

PR #64, opened 2016-09-29
Make references support next-error and previous-error.

josteink:
After looking into this, I found out that setting text-properties in a `compilation-mode` buffer is troublesome. Basically they will get constantly wiped, as `compilation-mode` tries to update its own properties (and handle the fact that it can be used in a comint).

Working around this would require quite some cumbersome hacks, and possibly have to rely on Emacs' current implementation of `compilation-mode`. Leaving aside the stability and compatibility issues that would entail, it also wouldn't really create any nicer code than what we already have.

And since you insisted on the formatting to stay pretty much exactly as it already is... I decided to try another approach which did not rely on `compilation-mode` directly: I looked into what `next-error` and `previous-error` does and how `compilation-mode` and `occur-mode` hooks into them.

And that was actually very simple. Using your existing references-functions I was able to just add the following snippet, and now it works.

The results are in this PR. This closes https://github.com/ananthakumaran/tide/issues/61

### DIFF
--- a/tide.el
+++ b/tide.el
@@ -695,6 +695,16 @@ With a prefix arg, Jump to the type definition."
 
 ;;; References
 
+(defun tide-next-error-function (n &optional reset)
+  "Override for `next-error-function' for use in tide-reference-mode buffers."
+  (interactive "p")
+
+  (with-current-buffer (get-buffer-create "*tide-references*")
+    (if (> n 0)
+        (tide-find-next-reference (point) n)
+      (tide-find-previous-reference (point) (- n)))
+    (tide-goto-reference)))
+
 (defun tide-find-next-reference (pos arg)
   "Move to next reference."
   (interactive "d\np")
@@ -823,7 +833,9 @@ number."
               (progn
                 (message "This is the only usage.")
                 (tide-jump-to-filespan usage nil t))
-            (display-buffer (tide-insert-references references))))
+            (progn
+              (display-buffer (tide-insert-references references))
+              (setq next-error-function #'tide-next-error-function))))
       (message (plist-get response :message)))))
 
 
--- a/tide.el
+++ b/tide.el
@@ -699,11 +699,15 @@ With a prefix arg, Jump to the type definition."
   "Override for `next-error-function' for use in tide-reference-mode buffers."
   (interactive "p")
 
-  (with-current-buffer (get-buffer-create "*tide-references*")
-    (if (> n 0)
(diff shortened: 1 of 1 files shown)


josteink:
I think it's worth nothing that this implementation (unlike the earlier one) doesn't incur any compatibility issues with previous usage or previous work-flows and just adds another option for how to interface with the references system.

It really should be a win-win for everyone :)

tide.el
```diff
@@ -695,6 +695,16 @@ With a prefix arg, Jump to the type definition."
 
 ;;; References
 
+(defun tide-next-error-function (n &optional reset)
+  "Override for `next-error-function' for use in tide-reference-mode buffers."
+  (interactive "p")
+
+  (with-current-buffer (get-buffer-create "*tide-references*")
```
ananthakumaran: do we need `get-buffer-create` here? I don't think we should process further if the buffer is not available

tide.el
```diff
@@ -695,6 +695,16 @@ With a prefix arg, Jump to the type definition."
 
 ;;; References
 
+(defun tide-next-error-function (n &optional reset)
```
ananthakumaran: should we also add support for `reset` param? from the description, it looks like we need to go to the start of the buffer before moving to the next/previous `n` references if `reset` is non nil. `first-error` function expects this functionality.

tide.el
```diff
@@ -823,7 +833,9 @@ number."
               (progn
                 (message "This is the only usage.")
                 (tide-jump-to-filespan usage nil t))
-            (display-buffer (tide-insert-references references))))
+            (progn
```
ananthakumaran: `progn` is not necessary, as `-if-let` already accepts multiple else expressions.

tide.el
josteink: Fair enough. I thought about doing a as minimal as possible implementation for the first time around. I've never used the reset functionality personally, so I didn't consider it pressing.

tide.el
josteink: Indeed. I'll make sure the function exit cleanly when buffer doesn't exist.

tide.el
josteink: Will fix this up!

josteink:
I'll also add a keybinding "q" for the find-references, buffer and then I'll consider this PR done :)

josteink:
That should fix all issues. Let me know what you!

ananthakumaran:
Thanks for PR

PR #65, opened 2016-10-05
Fix screencast in README

josteink:
Use inline image-reference.

Original GIF replaced with an optimized GIF, because original exceeded Github's maximum content-length.


ananthakumaran:
This is deliberate decision. Adding a gif in readme would be too distracting & just wastes the viewer's bandwidth.

josteink:
Ok. Fair enough.

PR #70, opened 2016-10-24
add tide-hl-identifier-mode

dominikh:
Add a minor mode for automatically highlighting occurences of an ID.

tide-hl-mode mkv

### DIFF
--- a/tide.el
+++ b/tide.el
@@ -1117,6 +1117,105 @@ number."
 
 (add-to-list 'flycheck-checkers 'tsx-tide)
 
+;;; Identifier highlighting
+
+(defun tide-command:documentHighlights ()
+  (tide-send-command-sync
+   "documentHighlights"
+   `(:file ,buffer-file-name :line ,(tide-line-number-at-pos) :offset ,(tide-current-offset) :filesToSearch (,buffer-file-name))))
+
+(defface tide-hl-identifier-face
+  '((t (:inherit highlight)))
+  "Face used for highlighting identifiers in `tide-hl-identifier'."
+  :group 'tide)
+
+(defcustom tide-hl-identifier-idle-time 0.1
+  "How long to wait after user input before highlighting the current identifier."
+  :type 'float
+  :group 'tide)
+
+(defvar tide--current-hl-identifier-idle-time
+  0
+  "The current delay for hl-identifier-mode.")
+
+(defvar tide--hl-identifier-timer
+  nil
+  "The global timer used for highlighting identifiers.")
+
+;;;###autoload
+(defun tide-unhighlight-identifiers ()
+  "Remove highlights from previously highlighted identifier."
+  (remove-overlays nil nil 'tide-overlay 'sameid))
+
+;;;###autoload
+(defun tide-hl-identifier ()
+  "Highlight all instances of the identifier under point. Removes
+highlights from previously highlighted identifier."
+  (interactive)
+  (tide-unhighlight-identifiers)
+  (tide--hl-identifier))
+
+(defun tide--hl-identifier ()
+  "Highlight all instances of the identifier under point."
+  (let ((response (tide-command:documentHighlights)))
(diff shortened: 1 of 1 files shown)


dominikh:
Is there a way to limit the tide-command:references query to references in the current file? For types like Event, the current implementation of this PR is too slow, because of all the references in other files and modules.

ananthakumaran:
Sorry for being late. Don't get much free time nowadays.

* Is this code based on any minor-mode? 

* `references` might be a wrong command to use. There is another command named [`documentHighlights`](https://github.com/Microsoft/TypeScript/blob/master/lib/protocol.d.ts#L366) which might be suitable.

``` lisp
(defun tide-command:documentHighlights (cb)
  (tide-send-command
   "documentHighlights"
   `(:file ,buffer-file-name :line ,(tide-line-number-at-pos) :offset ,(tide-current-offset) :filesToSearch (,buffer-file-name))
   cb))
```

Don't use the sync version, because it will block emacs till tsserver responds. If we use the async version, we don't have to worry much about the latencies.

dominikh:
> Is this code based on any minor-mode?

I have written similar code for go-guru, to highlight identifiers in Go code. I have the copyright to that code.  The same code does not exist as a generic minor mode, however.

> Don't use the sync version, because it will block emacs till tsserver responds. If we use the async version, we don't have to worry much about the latencies.

The main reason I went with the sync version is that it is much simpler to implement, as we will not have to worry about the user having moved point somewhere else before the result comes in, nor about cancelling outstanding requests when the user moves point. In go-guru, that worked fine because of millisecond response times.

Also, while using the async version would solve the hangs, it would still require fast response times to have a good user experience. Sometimes having to wait a second for a highlight would feel very inconsistent.

> references might be a wrong command to use. There is another command named documentHighlights which might be suitable

I have no experience with tsserver and am having a hard time finding documentation on its API. Where can I look up what exactly that function does?

ananthakumaran:
> I have written similar code for go-guru, to highlight identifiers in Go code. I have the copyright to that code. The same code does not exist as a generic minor mode, however.

sounds great.

> Don't use the sync version, because it will block emacs till tsserver responds. If we use the async version, we don't have to worry much about the latencies.
The main reason I went with the sync version is that it is much simpler to implement, as we will not have to worry about the user having moved point somewhere else before the result comes in, nor about cancelling outstanding requests when the user moves point. In go-guru, that worked fine because of millisecond response times.

> Also, while using the async version would solve the hangs, it would still require fast response times to have a good user experience. Sometimes having to wait a second for a highlight would feel very inconsistent.

Agree, async version might be hard to implement than sync. But it also makes editing experience much smoother. Let's not worry about canceling outstanding requests. We can simply not do anything in the callback if the user has moved off the identifier. tsserver usually response in milliseconds and because we limit the scope of the search to single file, I hope it won't take too long to respond

> I have no experience with tsserver and am having a hard time finding documentation on its API. Where can I look up what exactly that function does?

best bet is to look the source code. [https://github.com/Microsoft/TypeScript/blob/master/src/server/protocol.ts](protocol.ts) contains the interface (request and response format). [session.ts](https://github.com/Microsoft/TypeScript/blob/master/src/server/session.ts#L1571) contains the implementation.

dominikh:
Would you be fine with merging the sync version first, and changing it to an async version in a separate PR? It'd make the change simpler to review. Though I still believe that an async version isn't really necessary. The sync version with millisecond response times has no noticeable delay in Emacs (we've used that in go-guru for months).

I'll give documentHighlights a go and update the PR accordingly.

dominikh:
@ananthakumaran I've pushed a new version of the PR, now using documentHighlights – which works like a charm. It's still using the sync code for now, but it's blazing fast, only throttled by the configurable delay (which defaults to 0.1s)

ananthakumaran:
I will do a proper review soon. Agree, we can convert to async later.

PR #76, opened 2016-11-08
Make doc-view mode read-only and easily closed.

josteink:
With current implementation `tide-documentation-at-point` returns a buffer where you can edit it's content and you have to manually kill the buffer when you're done working with it.

This patch makes the documentation-buffer read-only, and makes it easily closable according to existing Emacs-conventions by simply pressing "q".

### DIFF
--- a/tide.el
+++ b/tide.el
@@ -270,10 +270,13 @@ LINE is one based, OFFSET is one based and column is zero based"
 
 (defun tide-doc-buffer (string)
   (with-current-buffer (get-buffer-create "*tide-documentation*")
-    (erase-buffer)
-    (when string
-      (save-excursion
-        (insert string)))
+    (setq buffer-read-only t)
+    (let ((inhibit-read-only t))
+      (erase-buffer)
+      (when string
+        (save-excursion
+          (insert string))))
+    (local-set-key (kbd "q") #'quit-window)
     (current-buffer)))
 
 ;;; Server


tide.el
```diff
@@ -270,10 +270,15 @@ LINE is one based, OFFSET is one based and column is zero based"
...
+    (read-only-mode -1)
     (erase-buffer)
     (when string
       (save-excursion
         (insert string)))
+
+    (read-only-mode t)
+    (local-set-key (kbd "q") #'quit-window)
+
     (current-buffer)))
```
ananthakumaran: I am not sure lisp program is supposed to use `read-only-mode`

> Do not call this from a Lisp program unless you really intend to
do the same thing as the \\[read-only-mode] command, including
possibly enabling or disabling View mode.  Also, note that this
command works by setting the variable `buffer-read-only', which
does not affect read-only regions caused by text properties.  To
ignore read-only status in a Lisp program (whether due to text
properties or buffer state), bind `inhibit-read-only' temporarily
to a non-nil value.

tide uses `(setq buffer-read-only t)` and `(let ((inhibit-read-only t) ..)` in reference buffer to achieve the same effect

josteink:
Good catch. Fixed.

josteink:
And there it should all be fixed properly. Anything else you think should be fixed?

ananthakumaran:
Thanks

PR #86, opened 2016-12-02
Rename: prettify symbol-name presented when doing rename-refactoring

josteink:
Currently the current code *may* produce something like the following
in the minibuffer:

````
Rename "d:/repo/src/ui/viewBase".ViewBase to: "d:/repo/src/ui/viewBase".ViewBase
````

With this commit, we get the symbol-name only, as expected:

````
Rename ViewBase to: ViewBase
````

### DIFF
--- a/tide.el
+++ b/tide.el
@@ -981,7 +981,7 @@ number."
     (tide-on-response-success response
       (if (eq (tide-plist-get response :body :info :canRename) :json-false)
           (message "%s" (tide-plist-get response :body :info :localizedErrorMessage))
-        (let* ((old-symbol (tide-plist-get response :body :info :fullDisplayName))
+        (let* ((old-symbol (tide-plist-get response :body :info :displayName))
                (new-symbol (tide-read-new-symbol old-symbol))
                (locs (tide-plist-get response :body :locs))
                (count 0))

### CODE REVIEW

**Comment by josteink** [comment]
@ananthakumaran: any news on this one? :)

**Comment by ananthakumaran** [comment]
going over the spec of RenameInfo looks like we could use `displayName`.

```typescript
    /**
      * Information about the item to be renamed.
      */
    export interface RenameInfo {
        /**
          * True if item can be renamed.
          */
        canRename: boolean;

        /**
          * Error message if item can not be renamed.
          */
        localizedErrorMessage?: string;

        /**
          * Display name of the item to be renamed.
          */
        displayName: string;

        /**
          * Full display name of item to be renamed.
          */
        fullDisplayName: string;

        /**
          * The items's kind (such as 'className' or 'parameterName' or plain 'text').
          */
        kind: string;

        /**
          * Optional modifiers for the kind (such as 'public').
          */
        kindModifiers: string;
    }

```

Could you check whether that field gives the short name?

josteink:
I agree that looks like a much better approach. I'll check it out.

josteink:
Just tested and it seems to do what you would expect. I've updated the PR to just change this key instead.

ananthakumaran:
thanks

Issue #93, opened 2016-12-17
keyof is unknown with bundled tsserver (tide 20161205.2257)


asmundg:
I'm not 100% certain of how the toolchain works here, so sorry in advance if this is adressed to the wrong place. The typescript-tide flychecker appears to fail when using tide's bundled tsserver. The following compiles fine with tsc 2.1.4, but the checker fails with "Cannot find name 'keyof'":

```
interface Foo {
    a: string;
}

export function foo(foo: keyof Foo) {
    return foo;
}

foo("a");
//foo("b");
```

ananthakumaran:
can we close this?

asmundg:
I believe so. Thanks!

PR #94, opened 2016-12-17
Add configurable workarond for missing shebangs and env on windows

asmundg:
Since the #!/usr/bin/env node header in typescript's tsserver doesn't
work too well on windows, this introduces a new config variable that
allows you to force the executable to run your tsserver with.

This is sort of related to #93, since typescript's tsserver supports the new syntax.

### DIFF
--- a/tide.el
+++ b/tide.el
@@ -75,6 +75,13 @@ above."
   :type '(choice (const nil) string)
   :group 'tide)
 
+(defcustom tide-tsserver-executor nil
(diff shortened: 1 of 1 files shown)


ananthakumaran:
Could you test the master and see if it solves your problem.

asmundg:
It works, great stuff! I wasn't sure if it was safe to assume that tsserver is always a node executable, but that's your call. :)

PR #100, opened 2017-01-09
feat(tide): Add configurable ability to insert completion template fo…

### LINKED ISSUE #84: Add templates after autocomplete selection
After selecting an autocomplete suggestion, Company-mode has the ability to "templatify" the resulting signature, like in the following screencast:
https://www.google.com/imgres?imgurl=http%3A%2F%2Fi.stack.imgur.com%2F4gKHP.gif&imgrefurl=http%3A%2F%2Femacs.stackexchange.com%2Fquestions%2F9591%2Fhow-to-display-type-of-c-variable-in-a-curses-type-menu&docid=cSBQM6CfuxI7nM&tbnid=yBHCkLQMqoIVDM%3A&vet=1&w=1100&h=900&bih=2049&biw=1910&ved=0ahUKEwjo6ca98MzQAhUhqVQKHYIVDs4QMwgfKAIwAg&iact=mrc&uact=8

This would be a nice (configurable) addition.

KevinVillela:
…r methods.

Company-mode has the ability to templatify completion strings,
so this commit integrates with that.

Fixes #84


mzgnr:
@ananthakumaran friendly ping for this PR 😄

tide.el
```diff
@@ -44,6 +44,7 @@
 (defvar web-mode-code-indent-offset)
 (defvar sgml-basic-offset)
 (defvar company-backends)
+(defvar company-gtags-insert-arguments)
```
ananthakumaran:
```elisp
(defcustom tide-completion-insert-arguments t
  "When non-nil, insert function arguments as a template after completion."
  :type 'boolean
  :group 'tide)
```

can you declare something like the above? depending on the config of gtags completion backend doesn't sound well

tide.el
```diff
@@ -778,7 +796,11 @@ With a prefix arg, Jump to the type definition."
     (ignore-case tide-completion-ignore-case)
     (meta (tide-completion-meta arg))
     (annotation (tide-completion-annotation arg))
-    (doc-buffer (tide-completion-doc-buffer arg))))
+    (doc-buffer (tide-completion-doc-buffer arg))
+    (post-completion  (-when-let (template (tide-completion-template arg))
+                        (when company-gtags-insert-arguments
+                          (insert template)
+                          (company-template-c-like-templatify template))))))
```
ananthakumaran: `c-like-templatify` doesn't seem to work in some cases. example `document.addEventListener(type: "MSContentZoom", listener: (this: Document, ev: UIEvent) => any, useCapture?: boolean)`.

tide.el
KevinVillela: Yes definitely, for some reason I thought that this was generic to company, but it's actually just for the gtags package.

tide.el
KevinVillela: This was because `document.addEventListener` defined some generics. I fixed this by modifying the regexp, but it now has a new problem: c-like-templatify does not handle the function argument well. It also doesn't know anything about optional arguments (making deleting them annoying), so what we should really do is write our own templatify function for TypeScript, since it is quite different from c. Unfortunately, I don't think my elisp-foo is good enough for that at the moment, so I may need to close this up for a while :\

tide.el
```diff
@@ -468,6 +490,7 @@ LINE is one based, OFFSET is one based and column is zero based"
   (tide-send-command "configure" `(:hostInfo ,(emacs-version) :file ,buffer-file-name :formatOptions ,(tide-file-format-options))))
 
 (defun tide-command:projectInfo (cb &optional need-file-name-list)
+  (message "hi")
   (tide-send-command "projectInfo" `(:file ,buffer-file-name :needFileNameList ,need-file-name-list) cb))
```
josteink: This looks like leftover which probably should be removed.

josteink:
@ananthakumaran I'm not sure how we would do that. Do you have any suggestions?

The way I see it, there are at least three options:

* Show optional parameters and let the user remove them (like now).
* Create some fancy system which let's us partially complete on parameters as we type them. This sounds like a boat-load of work :)
* Don't show optional parameters at all. Let them be discovered through `eldoc`/quickinfo.

So for a "quick fix", I would assume the third option is the most realistic. Any opinions?

ananthakumaran:
(required: type, required: type, |optional: type, optional: type)

Providing a keyboard shortcut for removing all the rest of args would be ideal I guess. So once you reach an optional args, you could decide to remove all the rest of the args with one action. In the current flow you would have to type something and then delete it.  This is a minor annoyance.

The major problem with this PR is `company-template-c-like-templatify` only works for a subset of type specification returned by tsserver.

josteink:
Closing due to lack of activity and feedback.

@KevinVillela: Feel free to re-open if you wish to address any of the comments raised.

PR #115, opened 2017-02-16
Remove newlines from command fixes #114

### LINKED ISSUE #114: Stopped working after upgrade to tide-20170210.1932
After upgrading to 20170210.1932, I'm constantly seeing:
Error running timer ‘tide--hl-identifiers-function’: (error "Server does not exist. Run M-x tide-restart-server to start it again")
If I restart the server then I can see the tsserver process start and then immediately exit. I can't see an error message.
I'm on Windows 7, GNU Emacs 25.1.1 (x86_64-w64-mingw32) of 2016-09-17

ksjogo:


ananthakumaran:
@ksjogo can you post your log file without this fix? I am trying to understand where the new line is inserted?

ksjogo:
The newline is Inserated After each key value pair. I suppose something is pretty printing the json. 
Might be related to Emacs master and not the same bug as here . 
But the change is definitely fixing something :D

ananthakumaran:
what is your emacs version? instead of replacing newlines, can you add `(json-encoding-pretty-print false)` in the enclosing let* and see if that fixes the problem?

ksjogo:
`emacs-version` is `GNU Emacs 26.0.50.22 (x86_64-apple-darwin16.1.0, NS appkit-1504.60 Version 10.12.1 (Build 16B2555)) of 2017-02-17`
The pp encoding change also fixes the the current problem.

Issue #116, opened 2017-02-23
Add support for code-actions/code-fixes


josteink:
`tsserver` supports doing various "code actions" (or [code fixes](https://github.com/Microsoft/TypeScript/blob/master/src/server/protocol.ts#L94) depending on whether you read the TypeScript source or the TypeScript blog). These can be anything the server deems useful.

Starting with [TypeScript 2.2](https://blogs.msdn.microsoft.com/typescript/2017/02/22/announcing-typescript-2-2/) there's now a bunch of new and **very interesting** code actions available:

* Adding missing imports
* Adding missing properties
* Adding forgotten `this.` to variables
* Removing unused declarations
* Implementing abstract members

If getting all these features into `tide` is only a matter of adding a new command or two, that seems like a very worthwhile effort.

josteink:
Scimming the source it seems the following needs to be done:

* We need to issue a [`CodeFixesRequest`](https://github.com/Microsoft/TypeScript/blob/master/src/server/protocol.ts#L400-L403) with our current location in the solution.
* The [response](https://github.com/Microsoft/TypeScript/blob/master/src/server/protocol.ts#L1393-L1396) will contain zero or several [`CodeAction`](https://github.com/Microsoft/TypeScript/blob/master/src/server/protocol.ts#L1398-L1403)-instances. We'll need to present these to the user. Good for us each action has a server-provided `description` field :)
* Once a particular action is chosen by the user, we need to execute the associated [`FileCodeEdits`](https://github.com/Microsoft/TypeScript/blob/master/src/server/protocol.ts#L1388-L1391) (note: not one! plural!)
* Each `FileCodeEdit` contains the changes needed to be made to a single file. These are encoded in a array of [`CodeEdit`](https://github.com/Microsoft/TypeScript/blob/master/src/server/protocol.ts#L1363-L1386)-instances.
* The `CodeEdit`-instance basically tells us to replace the text between `start` and `end` with a new one (`newText`), which should be trivial.

All in all, this looks like more than a 2-liner, but nothing fundamentally hard. This is definitely doable. So who's volunteering for eternal Emacs fame and glory? :D

josteink:
To anyone who wants to doodle further at this, this was my doodle-code:

````elisp
(defun tide-command:getCodeFixes ()
  (tide-send-command-sync "getCodeFixes" `(:file ,buffer-file-name
                                  :startLine ,(tide-line-number-at-pos)
                                  :startOffset ,(tide-current-offset)
[...]

PR #117, opened 2017-02-24
Add support for code-fixes.

josteink:
This closes https://github.com/ananthakumaran/tide/issues/116.

### DIFF
--- a/tide.el
+++ b/tide.el
@@ -692,10 +692,7 @@ With a prefix arg, Jump to the type definition."
     (save-excursion
       (dolist (file-change file-changes)
         (with-current-buffer (find-file-noselect (plist-get file-change :fileName))
-          (tide-apply-edits (plist-get file-change :textChanges))))
-      (dolist (file-change file-changes)
-        (with-current-buffer (find-file-noselect (plist-get file-change :fileName))
-          (tide-format)
+          (tide-format-regions (tide-apply-edits (plist-get file-change :textChanges)))
           (basic-save-buffer))))))
 
 
@@ -1089,12 +1086,14 @@ number."
 (defun tide-apply-edit (edit)
   (goto-char (tide-location-to-point (plist-get edit :start)))
   (delete-region (point) (tide-location-to-point (plist-get edit :end)))
-  (insert (plist-get edit :newText)))
+  (let ((start (point-marker)))
+    (insert (plist-get edit :newText))
+    (cons start (point-marker))))
 
 (defun tide-apply-edits (edits)
   (save-excursion
-    (-each (nreverse edits)
-      (lambda (edit) (tide-apply-edit edit)))))
+    (-map (lambda (edit) (tide-apply-edit edit))
+          (nreverse edits))))
 
 (defun tide-format-region (start end)
   (let ((response (tide-send-command-sync
@@ -1106,6 +1105,13 @@ number."
                   :endOffset ,(tide-offset end)))))
     (tide-on-response-success response
       (tide-apply-edits (plist-get response :body)))))
+
+(defun tide-format-regions (ranges)
+  (let ((positions (->>
+                    ranges
+                    (-mapcat (lambda (range) (list (marker-position (car range)) (marker-position (cdr range)))))
+                    (-sort '<))))
+    (tide-format-region (-min positions) (-max positions))))
 
 ;;; Mode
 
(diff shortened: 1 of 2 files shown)


tide.el
```diff
@@ -651,6 +651,57 @@ With a prefix arg, Jump to the type definition."
...
+(defun tide-get-fix-from-description (desc fixes)
+  (car
+   (-filter (lambda (fix) (string-equal desc
+                                        (tide-get-fix-description fix)))
+            fixes)))
+
+(defun tide-apply-codefix (fix)
+  "Apply a single `FIX' (which may apply to several files)!"
+
+  (setq my-tide-applied-fix fix)
```
ananthakumaran: left over?

josteink: Oops. Yes.

tide.el
```diff
@@ -651,6 +651,55 @@ With a prefix arg, Jump to the type definition."
...
+                                                 :startOffset ,(tide-current-offset)
+                                                 :endLine ,(tide-line-number-at-pos)
+                                                 :endOffset ,(+ 1 (tide-current-offset))
+                                                 :errorCodes ,(tide-get-flycheck-errors-ids-at-point))))
+
+(defun tide-get-fix-description (fix)
+  (plist-get fix :description))
+
+(defun tide-get-fix-from-description (desc fixes)
+  (car
```
fbergroth: You can use `-first` instead of `car` and `-filter`.

tide.el
```diff
@@ -1020,10 +1069,18 @@ number."
       (tide-format-region (region-beginning) (region-end))
     (tide-format-region (point-min) (point-max))))
 
+(defun tide-normalize-line-end (text)
+  (while (string-match "\r" text)
```
fbergroth: Wrap this with `(save-match-data ...)`

fbergroth:
Cool! Is the plan to also highlight locations where suggestions can be applied?

Intero has some support for this: http://commercialhaskell.github.io/intero/#suggestions

josteink:
> Is the plan to also highlight locations where suggestions can be applied?

That sounds like a "nice to have" version 2 feature, and I'm not actively going to be working towards that goal. I just wanted to have the basic feature in place, because it seems like tsserver is going some really cool places now :)

If/once this is accepted, feel free to improve the code with highlight/preview of the relevant fixes.

tide.el
```diff
@@ -1020,10 +1067,19 @@ number."
...
+      (setq text (replace-match "" nil nil text))))
+  text)
+
 (defun tide-apply-edit (edit)
   (goto-char (tide-location-to-point (plist-get edit :start)))
   (delete-region (point) (tide-location-to-point (plist-get edit :end)))
-  (insert (plist-get edit :newText)))
+  (let* ((newtext (plist-get edit :newText))
+         ;; we often get ^M chars in our returns! kill them with fire!
+         (normalized (tide-normalize-line-end newtext)))
```
ananthakumaran: I am not sure about removing `^M`. Probably, we would need to figure out why tsserver returns this chars when not expected. But removing won't fix this problem for all, because some might use different newline character.

tide.el
```diff
@@ -651,6 +651,53 @@ With a prefix arg, Jump to the type definition."
...
+(defun tide-apply-codefix (fix)
+  "Apply a single `FIX' (which may apply to several files)!"
+
+  (let ((file-changes (plist-get fix :changes)))
+    (save-excursion
+      (dolist (file-change file-changes)
+        (find-file (plist-get file-change :fileName))
+        (tide-apply-edits (plist-get file-change :textChanges)))))
+
+  (tide-format))
```
ananthakumaran: Running tide-format, when they asked for codeFixes is bit unexpected, especially, format applies for the whole buffer. I am not sure about how to fix this problem either, the codeEdits returned by tsserver doesn't leave the buffer in a good formatted state either

tide.el
```diff
@@ -651,6 +651,53 @@ With a prefix arg, Jump to the type definition."
...
+
+
+(defun tide-apply-codefixes ()
+  "Query the server for possible code-fixes and apply them."
```
ananthakumaran: Should we consider better name for this function? This is going to be a public api and we probably can't change once commited in master. `tide-quick-fix` or any others?

tide.el
ananthakumaran: I did some experiments, if we set the `newLineCharacter` format to `\r\n`, then tide returns `^M`, otherwise not. So, probably, best to let the user fix the format options rather than to replace it here.

ananthakumaran:
Looks like there is an open issue to support code formatting https://github.com/Microsoft/TypeScript/issues/12249. We would have to manually call format till this is fixed I guess.

ananthakumaran:
Merged. Thanks for the PR.

Issue #107, opened 2017-02-03
Add ability to navigate to named type


josteink:
One feature found in many IDEs is the ability to search for a type-name and navigate directly to it, often with a `completing-read` like interface which is populated with all known types.

Would this be possible to add?

I've fooled around with the following code, which *almost* but not quite works:

````emacs lisp
(defun tide-filter-types (body)
  (let ((result))
    (dolist (item body)
      (let ((kind (plist-get item :kind)))
        (when (or
               (string-equal kind "interface")
               (string-equal kind "type"))
          (setq result (append result (list (plist-get item :name)))))))
    result))

(defun tide-get-types ()
  "Gets all known types (by cheating and using the completions system)"

  (save-excursion
    (goto-char (point-max))
    (let* ((command-args `(:file ,(buffer-file-name) :line ,(tide-line-number-at-pos) :offset ,(tide-current-offset)))
           (command-response (tide-send-command-sync "completions" command-args))
           (response-body (plist-get command-response :body))
           (type-names (tide-filter-types response-body)))
      type-names)))

(defun tide-test-get-types ()
  (interactive)
  (let ((types (tide-get-types)))
    (setq my-test-get-types types)
    (completing-read "Select type to navigate to: " types)))
````

What it does is this:

* Move cursor to end of file to ensure we're not in any completion-context which will limit our results
* Request a completion from tsserver (which outside a completion prefix, should include "all" possible completions)
* We take the response and filter for types.
* We let the user choose a type.

This approach has several drawbacks and problems:

* We ask tsserver for way more than we need. So its slow.
* It will only show *imported* types.
* Even worse, it seems to only include types which has static members (because why on earth would we want to complete a type-name if we can't use it for a invocation in this otherwise empty statement-context?)

Checking the [tsserver protocol](https://github.com/Microsoft/TypeScript/blob/master/src/server/protocol.ts) I found no such command. The closest seems to be `navbar` or `navtree`, but they only return types from the currently loaded file. Iterating over all known files in the project and aggregating the results would probably work, but I'm going to guess it will be so slow it won't be close to usable.

Do you have any idea of another possible approach here? Or will we have to bugger Microsoft and request a new command to be added to tsserver before this can be implemented?

josteink:
Looking at [Visual Studio Code](https://github.com/Microsoft/vscode), pressing `Ctrl P` while in a TypeScript-project and then prefixing your search with `#` seems to search all types in the project.

[...]

Issue #97, opened 2017-01-04
Jump-to-implementation


AndyMoreland:
Hi!

I work on a project called [tsunami](https://github.com/andymoreland/tsunami) that is symbiotic with `tide`. I use a lot of your private functions and add some extra stuff. Recently I implemented support for the new TS2.1 "goto implementations" language service API here: https://github.com/AndyMoreland/tsunami/commit/70073d55dac7311794ee86019560148fb37b29cf#diff-6ecbb72ef0ea102717543dab4c97cb8eR46.

It's pretty simple, so perhaps it should go into `tide` directly. I use a popup menu UI for displaying the options inline in the buffer, but I don't think `tide` takes a dependency on that library yet.

![screen shot 2017-01-04 at 11 06 16 am](https://cloud.githubusercontent.com/assets/37327/21655188/d5d71c72-d26d-11e6-824f-710f6562ff83.png)

Feel free to use any of the code if you want.

Andy

ananthakumaran:
Cool. Feel free to open PRs if you feel something could be done better in tide. As for goto implementation, tide has no dependency on autocomplete. It has optional dependency on company. Perhaps we could use ido style selection for multiple choices. I will think about it.

josteink:
If we use `completing-read` whatever the user has set as his preference (ido, ivy, helm whatever) will be used.

I agree popups *look* nice, but I think `completing-read` should be preferred over it because popup lacks the fuzzy matching you you can use with helm/ido, and you are constrained to up/down/select style navigation only.

That said: I'd love to see a PR for this functionality in any form, and then worst case scenario we can improve upon it later :)

josteink:
I think I'll take a stab at implementing this myself. Looks easy enough and we have most of the bits we need already.

But in order to avoid excess duplicate code, I'll hold of submitting this until https://github.com/ananthakumaran/tide/pull/118 is merged and mainlined.

PR #118, opened 2017-02-24
Implement support for basic type-search with navigation.

josteink:
This closes https://github.com/ananthakumaran/tide/issues/107.

### DIFF
--- a/tide.el
+++ b/tide.el
@@ -533,6 +533,38 @@ With a prefix arg, Jump to the type definition."
 
 (defalias 'tide-jump-back 'pop-tag-mark)
 
+;;; Navigate to named member
+
+(defun tide-nav ()
+  "Search and navigate to named types."
+  (interactive)
+  (let ((completion-ignore-case t)
+        (last-completions nil)
+        ;; helm does not work well with dynamically/incrementally
+        ;; generated tables, so disable it for this function!
+        (helm-completing-read-handlers-alist '()))
+    (-when-let (completion
+                (completing-read
+                 "Search: "
+                 (completion-table-dynamic
+                  (lambda (prefix)
+                    (let ((response (tide-command:navto prefix)))
+                      (tide-on-response-success response
+                        (when-let (navto-items (plist-get response :body))
+                          (setq navto-items
+                                (-filter
+                                 (lambda (navto-item) (member (plist-get navto-item :kind) '("class" "interface" "type" "enum")))
+                                 navto-items))
+                          (setq last-completions navto-items)
+                          (-map (lambda (navto-item) (plist-get navto-item :name))
+                                navto-items)))))
+                  t) nil t))
+      (let ((navto-item (-find (lambda (navto-item) (string-equal completion (plist-get navto-item :name))) last-completions)))
+        (tide-jump-to-filespan navto-item)))))
+
+(defun tide-command:navto (type)
+  (tide-send-command-sync "navto" `(:file ,buffer-file-name :searchValue ,type :maxResultCount 100)))
+
 ;;; Eldoc
 
 (defun tide-annotate-display-part (display-part &optional highlight)


josteink:
This PR should be ready for review.

I've done some work to reduce code-duplication in general, and specifically code-duplication I know will be worsened even more when I take a stab at solving https://github.com/ananthakumaran/tide/issues/97.

Let me know if you have any objections :)

tide.el
```diff
@@ -533,6 +533,54 @@ With a prefix arg, Jump to the type definition."
...
+               (or (string-equal "class" kind)
+                   (string-equal "interface" kind)
+                   (string-equal "type" kind)
+                   (string-equal "enum" kind))))
+           body))
+
+(defun tide-get-type-name (item)
+  (plist-get item :name))
+
+(defun tide-go-to-type (type)
```
ananthakumaran: I guess `tide-jump-to-filespan` does the same things

tide.el
```diff
@@ -533,6 +533,54 @@ With a prefix arg, Jump to the type definition."
...
+  (tide-send-command-sync "navto" `(:file ,buffer-file-name :searchValue ,type)))
+
+(defun tide-parse-types-response (response)
+  (tide-filter-types
+   (plist-get response :body)))
+
+(defun tide-filter-types (body)
+  (-filter (lambda (item)
+             (let ((kind (plist-get item :kind)))
+               (or (string-equal "class" kind)
```
ananthakumaran: could use member function

tide.el
josteink: Indeed. I missed that one. And it seems to do a much better job too :)

tide.el
josteink: Good suggestion. Will do.

josteink:
Fixed up PR based on feedback so far. Looking shorter. Looking better :)

ananthakumaran:
`completing-read` supports dynamic function instead of a static collection. I put together the following based on your code and seems to work ok

```elisp
;;; Navigation

(defun tide-nav ()
  "Search and navigate to named types."
  (interactive)
  (let ((completion-ignore-case t)
        (last-completions nil))
    (-when-let (completion
                (completing-read
                 "Search: "
                 (completion-table-dynamic
                  (lambda (prefix)
                    (let ((response (tide-command:navto prefix)))
                        (when-let (navto-items (plist-get response :body))
                          (setq navto-items
                                (-filter
                                 navto-items))
                          (setq last-completions navto-items)
                                navto-items)))))
                  t) nil t))
      (let ((navto-item (-find (lambda (navto-item) (string-equal completion (plist-get navto-item :name))) last-completions)))
        (tide-jump-to-filespan navto-item)))))

(defun tide-command:navto (type)
  (tide-send-command-sync "navto" `(:file ,buffer-file-name :searchValue ,type :maxResultCount 100)))
```
You can see the documentation [here](https://www.gnu.org/software/emacs/manual/html_node/elisp/Programmed-Completion.html#Programmed-Completion). The way we convert the completion item to navbar is kind of hacky now. But, completing-read function even strips the text-properties associated with the candidate collection. So, I couldn't find any way to attach the navbar-item to the candidate string.

josteink:
Thats awesome!

Do you want me to look at cleaning that code and including it in the PR, or do you want to do it yourself?

ananthakumaran:
Give it a try and see if it works properly. If there are no issues, I can commit my changes directly. Let's not worry about minor duplication much, especially related to plucking data out of the structure returned by tsserver. All those two line helper functions don't seem to add any value and make code reading difficult. I had to read the function definition everytime to understand what it does....

josteink:
I'll look at it later this evening!

This is shaping up to be a seriously productive week with regard to getting tide up to date :)

josteink:
I tried it and it has some issues. Some obvious ones (which I figured out). Some less obvious ones (which I didn't).

First things first:

````elisp
(defun tide-command:navto (type)
  (tide-send-command-sync "navto" `(:file ,buffer-file-name :searchValue ,type :maxResultCount 100)))
````

This command is dispatched inside the minibuffer context... when we're not really in a file-based buffer at all. So `buffer-file-name` is `nil` and `tsserver` errors out. This was easily solved by grabbing this value early and passing it as a parameter to the command-function.

Now it no longer crashes, but I only get results if I disable `helm-mode`. When I do that I can type things into the search field and press `<tab>` to get completions. Trying the same with `helm-mode` on doesn't work at all.

Why it fails... I have no idea. Which brings me to your comment :)

> Let's not worry about minor duplication much, especially related to plucking data out of the structure returned by tsserver. All those two line helper functions don't seem to add any value and make code reading difficult.

I subscribe to the "school" of Robert C. Martin and his philosophy on functions: They should be like Unix-programs. They should be short, not deeply nested. They should do one thing only, and do them well. And it should be simple enough for its correctness to be self-evident. They should be named well enough so that you don't need to read them to understand what they do. They should read like english prose. (And I guess I failed there :) )

When I deconstruct "big" functions like this into small ones... At least for *me* it makes it easier to observe where things go wrong and why. It makes it easier to hook in `edebug` the only place you suspect it to be needed.

So in this case, since this is failing, I would I disagree about deconstructing the function not adding value. But ultimately, it's your project and you get to decide style-guide.

So right now we have two options as far as I can see:

* Rewrite the function to work in all popular `completing-read` extensions (like `helm`). Maybe we can look in similar functions, like `helm-projectile-grep`, to find out how it's done?
* Rewrite the function to temporarily *disable* `completing-read` extensions (like `helm`)

I'm not sure which one is easiest (or even possible), but I do know which one I would ultimately prefer.

**Comment by ananthakumaran** [comment]
> I tried it and it has some issues. Some obvious ones (which I figured out). Some less obvious ones (which I didn't).

> First things first:

> (defun tide-command:navto (type)
  (tide-send-command-sync "navto" `(:file ,buffer-file-name :searchValue ,type :maxResultCount 100)))
> This command is dispatched inside the minibuffer context... when we're not really in a file-based buffer at all. So buffer-file-name is nil and tsserver errors out. This was easily solved by grabbing this value early and passing it as a parameter to the command-function.

> Now it no longer crashes, but I only get results if I disable helm-mode. When I do that I can type things into the search field and press <tab> to get completions. Trying the same with helm-mode on doesn't work at all.

> Why it fails... I have no idea. Which brings me to your comment :)

`completion-table-dynamic` accepts a second argument, which is specifically designed to fix this issue. I guess helm-mode is at fault here, somehow mixing up the current buffer. I tested it without helm-mode enabled and worked properly.

> Let's not worry about minor duplication much, especially related to plucking data out of the structure returned by tsserver. All those two line helper functions don't seem to add any value and make code reading difficult.
> I subscribe to the "school" of Robert C. Martin and his philosophy on functions: They should be like Unix-programs. They should be short, not deeply nested. They should do one thing only, and do them well. And it should be simple enough for its correctness to be self-evident. They should be named well enough so that you don't need to read them to understand what they do. They should read like english prose. (And I guess I failed there :) )

> When I deconstruct "big" functions like this into small ones... At least for me it makes it easier to observe where things go wrong and why. It makes it easier to hook in edebug the only place you suspect it to be needed.

> So in this case, since this is failing, I would I disagree about deconstructing the function not adding value. But ultimately, it's your project and you get to decide style-guide.

I agree with the principle, indeed there are many small functions in tide.el already, the problem here is, the small function is not self-evident when you read them in isolation, and their functionality is not apparent from their name either. You need to have the whole context to understand what that small function does.

```elisp
(defun tide-get-type-name (item)
   (plist-get item :name))
```

But yeah, we could argue back and forth. Let's not get sidetracked.

So right now we have two options as far as I can see:

> Rewrite the function to work in all popular completing-read extensions (like helm). Maybe we can look in similar functions, like helm-projectile-grep, to find out how it's done?
> Rewrite the function to temporarily disable completing-read extensions (like helm)
I'm not sure which one is easiest (or even possible), but I do know which one I would ultimately prefer.

My guess is option 2 would be the easiest to do. I would argue the problem is with helm because it is supposed to replicate the functionality/interface of completing-read. Given that it is used by many, it's probably prudent to temporarily disable it.

josteink:
Hmmm. Looking at this, it seems getting it to work properly with `helm` will be more complex than I anticipated.

I'd love to avoid *too* hacky solutions, but if I make one simple change to your implementation, I'll have it disable `helm` for people who use it:

````elisp
;;; Navigation

(defun tide-nav ()
  "Search and navigate to named types."
  (interactive)
  (let ((completion-ignore-case t)
        (last-completions nil)
        ;; helm does not work well with dynamically/incrementally
        ;; generated tables, so disable it for this function!
        (helm-completing-read-handlers-alist '()))
    (-when-let (completion
                (completing-read
                 "Search: "
                 (completion-table-dynamic
                  (lambda (prefix)
                    (let ((response (tide-command:navto prefix)))
                        (when-let (navto-items (plist-get response :body))
                          (setq navto-items
                                (-filter
                                 navto-items))
                          (setq last-completions navto-items)
                                navto-items)))))
                  t) nil t))
      (let ((navto-item (-find (lambda (navto-item) (string-equal completion (plist-get navto-item :name))) last-completions)))
        (tide-jump-to-filespan navto-item)))))

(defun tide-command:navto (type)
  (tide-send-command-sync "navto" `(:file ,buffer-file-name :searchValue ,type :maxResultCount 100)))
````

The other options I've looked into (emulate isearch, use run-with-idle-timer, etc) has the complexity-factor going through the roof pretty quickly, so I've decided to not chase that goal for now.

For now, this is certainly much better than nothing.

As for my other refactorings... I guess I'll keep them around for when I implement"go to implemention" :)

ananthakumaran:
@josteink can you revert other commits except the last one so that I can merge this branch?

josteink:
I tried doing that initially, but something must have messed up in magit... No idea what, but I just discovered the same error.

Consider it fixed :)

josteink:
Awesome. Thanks :)

PR #120, opened 2017-02-27
Implement support for "jump to implementation".

josteink:
This closes https://github.com/ananthakumaran/tide/issues/97.

The reponse from "Jump to implementation" is not as rich as I'd like it to be: [We only get filename and position back, and no class-names](https://github.com/Microsoft/TypeScript/blob/master/src/server/protocol.ts#L585-L590).

It's not entirely uncommon to have several implementations of the same interface in the same file, so for that reason I decided that showing file-name alone was probably not specific enough to get a "unique" index we could use to look up the user's selection.

Let me know if you can think of a better way to solve this :)

Thanks to @AndyMoreland for giving me the push to provide all these patches :)

### DIFF
--- a/tide.el
+++ b/tide.el
@@ -296,6 +296,16 @@ LINE is one based, OFFSET is one based and column is zero based"
     (local-set-key (kbd "q") #'quit-window)
     (current-buffer)))
 
+(defun tide-get-item-from-list (list textfn text)
+  (-first (lambda (item)
+            (string-equal text (funcall textfn item)))
+          list))
+
+(defun tide-select-item-from-list (prompt list textfn)
+  (let* ((texts (-map textfn list))
+         (selected-text (completing-read prompt texts)))
+    (tide-get-item-from-list list textfn selected-text)))
+
 ;;; Events
 
 (defvar tide-event-listeners (make-hash-table :test 'equal))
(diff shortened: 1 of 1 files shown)


AndyMoreland:
Looks sane to me. Thanks for writing this in a generic-enough way to fit into `tide`. Glad I provided some inspiration.

I think that typescript upstream would probably accept a patch to include the type name in the response -- seems odd that they don't.

josteink:
Good point about TypeScript upstream might be able/want to fix this. I'll consider issuing a patch for just that.

For now, for compatibility's sake, I guess we'll have to do with this though.

josteink:
Hmm. Or maybe not. 

Looking at [the sources](https://github.com/Microsoft/TypeScript/blob/master/src/server/session.ts#L1587-L1592) (once again), there's both a `GetImplementations` and a `GetImplementations-Full`.

I remember now seeing it when I implemented some of the other new protocol commands, but it's worth noting that `GetImplementations-Full` is [marked `internal` in the protocol declaration](https://github.com/Microsoft/TypeScript/blob/master/src/server/protocol.ts#L22-L24).

For that reason I don't feel entirely comfortable just using it. Not being a public API, there's no saying if it's going to be working in the future.

I guess I'll have to raise an issue and ask what their policy on this matter is.

ananthakumaran:
FYI, we already use one such api. Look at `tide-command:quickinfo` for details. It uses `quickinfo-full` and falls back to `quickinfo` if it is not available. The full version provides enough information to syntax highlight the types

josteink:
Thanks for the tip! If that's so I'll try to do the same for this functions.

josteink:
After investigating further, I've found that the `full` command returns mostly the same data in a different format... It's not really helping us the slightest:

````elisp
;; item-value for "implementation"
(:file "d:/Git/repo/test.ts" :start (:line 13 :offset 5) :end (:line 15 :offset 6))

;; item-value for "implementation-full"
(:textSpan (:start 92 :length 67) :fileName "d:/Git/repo/test.ts")
````

With that cleared up, I'll leave this as is.

PR #121, opened 2017-02-27
Fix tide-nav for helm.

josteink:
After doing (better) testing of `tide-nav`, I discovered the following:

* my hacky "disable helm" workaround doesn't actually seem to work reliably. I must have messed something up during testing.
* Getting "something" which works in helm wasn't that hard.

This patch lets you type something, then press `<TAB>`(just as a regular completing-read would have you do) and then you get a helm-popup with the candidates, ready for selection. Much better than it used to be.

To avoid more errors in this department I've tested the following configurations:

* helm
* ido
* ivy
* none of the above

From what I can tell it all works as expected now.

### DIFF
--- a/tide.el
+++ b/tide.el
@@ -541,7 +541,10 @@ Noise can be anything like braces, reserved keywords, etc."
 
   (when (not (member (face-at-point) '(font-lock-keyword-face
                                        font-lock-comment-delimiter-face)))
-    (let* ((symbol (region-str-or-symbol))
+    ;; we could have used symbol-at-point here, but that leaves us unable to
+    ;; differentiate between a symbol named nil and no symbol at all.
+    ;; thing-at-point returns a string OR nil, which means we don't get this problem.
+    (let* ((symbol (thing-at-point 'symbol))
            (value (substring-no-properties (if (equal nil symbol) "" symbol))))
       (save-match-data
         (when (not (string-match "{|}|;|[|]" value))
--- a/tide.el
+++ b/tide.el
@@ -535,11 +535,24 @@ With a prefix arg, Jump to the type definition."
 
 ;;; Navigate to named member
 
-(defun tide-nav ()
+(defun tide-get-symbol-at-point ()
+  "Returns the symbol found at point, if not deemed 'noise'.
+Noise can be anything like braces, reserved keywords, etc."
+
+  (when (not (member (face-at-point) '(font-lock-keyword-face
+                                       font-lock-comment-delimiter-face)))
+    (let* ((symbol (region-str-or-symbol))
+           (value (substring-no-properties (if (equal nil symbol) "" symbol))))
+      (save-match-data
+        (when (not (string-match "{|}|;|[|]" value))
+          value)))))
+
+(defun tide-nav (arg)
   "Search and navigate to named types."
-  (interactive)
+  (interactive "P")
   (let ((completion-ignore-case t)
-        (last-completions nil))
+        (last-completions nil)
+        (default (when arg (tide-get-symbol-at-point))))
     (-when-let (completion
                 (completing-read-default
                  "Search: "
@@ -555,7 +568,7 @@ With a prefix arg, Jump to the type definition."
                           (setq last-completions navto-items)
                           (-map (lambda (navto-item) (plist-get navto-item :name))
                                 navto-items)))))
-                  t) nil t))
+                  t) nil t default))
       (let ((navto-item (-find (lambda (navto-item) (string-equal completion (plist-get navto-item :name))) last-completions)))
         (tide-jump-to-filespan navto-item)))))
 
--- a/tide.el
+++ b/tide.el
@@ -539,12 +539,9 @@ With a prefix arg, Jump to the type definition."
   "Search and navigate to named types."
   (interactive)
   (let ((completion-ignore-case t)
-        (last-completions nil)
-        ;; helm does not work well with dynamically/incrementally
-        ;; generated tables, so disable it for this function!
-        (helm-completing-read-handlers-alist '()))
+        (last-completions nil))
     (-when-let (completion
-                (completing-read
+                (completing-read-default
                  "Search: "
                  (completion-table-dynamic
                   (lambda (prefix)


fbergroth:
Have you tried dropping the dynamic completion tables and using a function as the collection arg? The author of ivy recommands this, see https://github.com/jorgenschaefer/elpy/issues/1084#issuecomment-280634679.

josteink:
Thanks for the suggestions and good advice!

Doing some very quick prototyping I managed to get this working instead:

````elisp
(defun tide-nav ()
  "Search and navigate to named types."
  (interactive)
  (let ((completion-ignore-case t)
        (last-completions nil)
        (buffer-name buffer-file-name))
    (-when-let (completion
                (completing-read-default
                 "Search: "
                 (lambda (prefix str _)
                   (let ((response (tide-command:navto buffer-name prefix)))
                       (when-let (navto-items (plist-get response :body))
                         (setq navto-items
                               (-filter
                                navto-items))
                         (setq last-completions navto-items)
                               navto-items))))) nil nil nil))
      (let ((navto-item (-find (lambda (navto-item) (string-equal completion (plist-get navto-item :name))) last-completions)))
        (tide-jump-to-filespan navto-item)))))

(defun tide-command:navto (buffer-name type)
  (tide-send-command-sync "navto" `(:file ,buffer-name :searchValue ,type :maxResultCount 100)))
````

It's not perfect either, but better matches helm-conventions (although having to press `<TAB>` is a bit unusual for me). I think it would be worthwhile to take a deeper stab at this.

**Comment by fbergroth** [comment]
Does it not work with `completing-read` instead of `completing-read-default`? (I can't test it right now)

**Comment by josteink** [comment]
Nope. Sadly not. That just gets us back to the "old" behaviour of helm doing nothing. (I'm also having issues with ivy-mode then). Not sure where it fails at. Like last time around, I did try various things, but didn't really get it working the way I wanted.

That said, I've pushed these general improvements:

* now it actually *works* with helm
* if invoked with a prefix-argument, it will suggest a "reasonable" default to search for.

I think all in all, this is a decent improvement over what we have.

But so far, you've been really good at coming up with suggestions for further improvements... Any opinion on the latest commit I pushed?

**Comment by fbergroth** [comment]
If dynamic completions are problematic using the standard completion interface, you can introduce a customizable variable `tide-completion-system` (similar to `projectile-completion-system`), and add optional support (no direct dependencies) to helm and ivy with:

```el
(pcase tide-completion-system
  (`ivy
   (if (fboundp 'ivy-read)
       (ivy-read ...)
     (user-error "Please install ivy")))
  (`helm
   (if (fboundp 'helm)
       (helm ...)
     (user-error "Please install helm")))
  (_
   (completing-read-default ...)))
```
Now you can use the direct ivy api with `dynamic-collection` etc. It's a bit of work though, but should provide the best user experience.

Sorry, I haven't tested your branch yet, I'm not using typescript right now, but probably will pretty soon.

josteink:
Appreciate the suggestion, and appreciate you taking time to spar with me. Thanks! :)

`projectile` has the same solution as you propose. Which for a long time meant I thought `projectile` didn't support `helm` at all. This problem gets worse the more packages which chooses it. IMO it just doesn't scale well. For that reason I'd rather not go down that route, if we can avoid it.

As a user I want to be able to just install a package and start using it, without having to read *too much* documentation and creating too many hooks/vars/custom-values.

The solution we have here now seems to work well enough for all current (and future?) completion systems without any need for specific support, and I think that's probably the best solution until we are forced to choose otherwise.

ananthakumaran:
@josteink thank you so much for taking time to check different plugins.

Let's clear about what we want to achieve

* Tide should work well on vanilla emacs.
* If some plugin is popular & used by most of the community, we could add it as a dependency (eg company, flycheck etc) and or expose hooks to make it work well.

It seems like, we are trying to compromise on the first objective to satisfy the second one. The function that is passed to `completing-read` as the second argument should implement a complex interface specified [here](https://www.gnu.org/software/emacs/manual/html_node/elisp/Programmed-Completion.html#Programmed-Completion). `completion-table-dynamic` function is added to bridge the gap. It returns a function which adheres to the interface. The interface expected by itself is quite simple.

We now seem to have a half-baked implementation, just so it would work with helm/ivy et al. It would likely break things for vanilla emacs user.

I propose the following things.

1) if we can avoid breaking helm/ivy etc, we can do it, but not at the expense of vanilla emacs. we can use `completing-read-default` instead of `completing-read` function.
2) we can make the function itself configurable.

tide-completing-read-function => this will be used for static list (like jump to implementation, select code fix etc). we could use `completing-read` as default.
tide-completing-read-dynamic-function => this will be used for dynamic list.. None of the alternate implementations of completing-read seem to implement dynamic function correctly.  so, it's better to use `completing-read-default` as default

@josteink what do think?

josteink:
You might be on to something. If we go "too far" in trying to write especially contrived "voodoo" code in order to make it magically work with various completion-frameworks, that's probably not better, clearer or more maintainable than explicitly writing some code to handle those we feel are big/important enough to support.

And if we in our quest to support these extensions (in my case helm) make it less useful for "vanilla" Emacs users (do they even exist? lol!), I agree that's a bad trade-off.

If we do like you say and create a few select wrapper functions for the actions where we feel use of such extension-frameworks can be useful to handle, we can always accept PRs in the future for completion-frameworks people feel have been left out.

And that means that those parties with a special interest in these matters will ensure the functionality they care about is well tested and well kept in their frameworks of choice, which leaves us with less of a burdon maintenance-wise. That's win win for everyone involved.

I think you may have convinced me. I've changed my mind on this matter :)

Now with that said... I'll go back to this patch and rewrite it to be "vanilla"-friendly first and foremost, and then we can create a separate issue/task/PR to create general completion-framework friendly code. Does that sound OK?

josteink:
And with that, the PR should back in "vanilla"-friendly state.

I'll be happy to consider any review on the current code, without too much side-tracking about completion-framework-wrappers for now :)

tide.el
```diff
@@ -535,16 +535,26 @@ With a prefix arg, Jump to the type definition."
...
 ;;; Navigate to named member
 
-(defun tide-nav ()
+(defun tide-get-symbol-at-point ()
+  "Returns the symbol found at point, if not deemed 'noise'.
+Noise can be anything like braces, reserved keywords, etc."
+
+  (when (not (member (face-at-point) '(font-lock-keyword-face
+                                       font-lock-comment-delimiter-face)))
+    (let* ((symbol (region-str-or-symbol))
```
ananthakumaran: can you check where this symbol is defined? It's not available for me. we could use `thingatpt` module which provides similar functions

josteink: Seems like this is one of my own utility functions from my own `.emacs` file. Oops!

````elisp
(defun region-str-or-symbol ()
  "Return the contents of region or current symbol."
  (if (region-active-p)
      (buffer-substring-no-properties
       (region-beginning)
       (region-end))
    (thing-at-point 'symbol)))
````

I've pushed a new commit where I've replaced this with `(thing-at-point 'symbol)` which is really all that should be needed here.

Using that other utility-function was completely overkill, and in any case an error.

PR #123, opened 2017-02-28
Fix newline issues with code-edits.

### LINKED ISSUE #122: nav-fix does not work reliably on Windows
When using `nav-fix` on Windows to for instance implement abstract members, interface members, or basically anything which involves newlines, the result is *not* how it should be.

Take the contrived example:

````typescript
interface TestInterface {
    boo();
}

class TestClass implements TestInterface {
}
````

When trying to apply a codefix to `TestClass` we expect the following:

````typescript
class TestClass implements TestInterface {
    boo() {
        throw new Error('Method not implemented.');
    }
}
````

Instead we get:

````typescript
class TestClass implements TestInterface {^M
    boo() {^M
throw new Error('Method not implemented.');^M
        }^M
}
````

This happens even though the currently open file has DOS encoding mode!

Trying to load the document may cause the `^M` to stop showing (depending on your find-file hooks and filters)... But it's still there, causing reflowing the document with `(indent-region (point-min) (point-max))` to give bad results:

````typescript
interface TestInterface {
    boo();
}

class TestClass implements TestInterface {^M
                                          boo() {^M
                                                 throw new Error('Method not implemented.');^M
                                                }^M
                                         }
````

I've tried to tweak `tide-format-options` with various `:newLineCharacter`-values, but they all give troublesome results. `\n` and `\r\n` gives the behaviour above. Using plain `\r` gives a visual lack of  `^M` chars, but still causes indentation to go wrong.

What's more interesting is that `tsserver` (in the response JSON) sends us the correct CR-LF format! (JSON pretty-printed below for clarity):

````json
{
  "command": "getCodeFixes",
  "arguments": {
    "file": "d:/git/repo/test.ts",
    "startLine": 55,
    "startOffset": 10,
    "endLine": 55,
    "endOffset": 11,
    "errorCodes": [
      2420
    ]
  }
}
Content-Length: 381

{
  "seq": 0,
  "type": "response",
  "command": "getCodeFixes",
  "request_seq": 0,
  "success": true,
  "body": [
    {
      "description": "Implement interface 'TestFace'.",
      "changes": [
        {
          "fileName": "d:/git/repo/test.ts",
          "textChanges": [
            {
              "start": {
                "line": 55,
                "offset": 39
              },
              "end": {
                "line": 55,
                "offset": 39
              },
              "newText": "boo() {\r\nthrow new Error('Method not implemented.');\r\n}\r\n"
            }
          ]
        }
      ]
    }
  ]
}
````

This means that the error happens in the way we process it, parse it and bring it back to the buffer!

Could this be an error in [`json-read-object`](https://github.com/ananthakumaran/tide/blob/e7d4fff70b5a1aa327ad0cda8c00b3642c6fea93/tide.el#L428-L439)?

**Edit:** This problem is probably not constrained to `tide-fix`, but involves anything which executes code-edits and results in line-shifts.

### PR DESCRIPTION
**Pull request by josteink**
As discussed in issue #122, Emacs actually never expects anything except `\n` to be inserted in a buffer for newlines, no matter how newlines is actually represented in that file.

Us inserting `^M` characters returned from `tsserver` is thus not us doing things properly nor preserving correctness. On the contrary it's our code being a source of (quite terrible!) bugs.

This PR fixes this and closes #122.

### DIFF
--- a/tide.el
+++ b/tide.el
@@ -1115,11 +1115,28 @@ number."
       (tide-format-region (region-beginning) (region-end))
     (tide-format-region (point-min) (point-max))))
 
+(defun tide-normalize-lineshift (str)
+  "Reformat `STR' to only contain line-shift formats expected by Emacs.
+
+When inserting text in an Emacs-buffer Emacs only ever expects \n
+for newlines, no matter what the actual encoding of the file
+is.  Inserting anything else causes issues with formatting and
+code-analysis, so clean it up!"
+
+  (save-match-data
+    ;; convert DOS CR+LF to LF
+    (while (string-match "\r\n" str)
+      (setq str (replace-match "\n" nil nil str)))
+    ;; conveer Mac CR to LF
+    (while (string-match "\r" str)
+      (setq str (replace-match "\n" nil nil str))))
+  str)
+
 (defun tide-apply-edit (edit)
   (goto-char (tide-location-to-point (plist-get edit :start)))
   (delete-region (point) (tide-location-to-point (plist-get edit :end)))
   (let ((start (point-marker)))
-    (insert (plist-get edit :newText))
+    (insert (tide-normalize-lineshift (plist-get edit :newText)))
     (cons start (point-marker))))
 
 (defun tide-apply-edits (edits)
--- a/tide.el
+++ b/tide.el
@@ -1123,13 +1123,10 @@ for newlines, no matter what the actual encoding of the file
 is.  Inserting anything else causes issues with formatting and
 code-analysis, so clean it up!"
 
-  (save-match-data
-    ;; convert DOS CR+LF to LF
-    (while (string-match "\r\n" str)
-      (setq str (replace-match "\n" nil nil str)))
-    ;; conveer Mac CR to LF
-    (while (string-match "\r" str)
-      (setq str (replace-match "\n" nil nil str))))
+  ;; convert DOS CR+LF to LF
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: tide.el [hunk 1]
```diff
@@ -1115,11 +1115,28 @@ number."
...
+  "Reformat `STR' to only contain line-shift formats expected by Emacs.
+
+When inserting text in an Emacs-buffer Emacs only ever expects \n
+for newlines, no matter what the actual encoding of the file
+is.  Inserting anything else causes issues with formatting and
+code-analysis, so clean it up!"
+
+  (save-match-data
+    ;; convert DOS CR+LF to LF
+    (while (string-match "\r\n" str)
```
ananthakumaran: can we use `replace-regexp-in-string`

tide.el
```diff
@@ -1115,11 +1115,28 @@ number."
...
+      (setq str (replace-match "\n" nil nil str)))
+    ;; conveer Mac CR to LF
+    (while (string-match "\r" str)
+      (setq str (replace-match "\n" nil nil str))))
```
ananthakumaran: we could use `subst-char-in-string` here

josteink:
Indeed. I was looking for functions like that, but I always struggle with finding the proper functions for direct string-manipulation.

Thanks for the leads! I've pushed a new commit with the simplified code.

josteink:
Awesome. Thanks.

This one was seriously annoying me here on the Windows-machine my company provides me :)

PR #125, opened 2017-03-04
Improve tide-jump-to-definition.

josteink:
While having `jump-to-implementation` as an available command is nice, having it as an automatic part of the `M-.` workflow makes it much more useful and accessible.

This PR contains changes to detect when navigating to the definition at point is navigating to the same location, and in those cases (and never with prefix args!) instead automatically proceed to implementations of abstract members, if found.

The code changes should be fairly obvious and IMO unintrusive.

### DIFF
--- a/tide.el
+++ b/tide.el
@@ -507,15 +507,28 @@ LINE is one based, OFFSET is one based and column is zero based"
 (defun tide-jump-to-definition (&optional arg)
   "Jump to the definition of the symbol at point.
 
-With a prefix arg, Jump to the type definition."
+If pointed at an abstract member-declaration, will proceed to look for
+implementations.  When invoked with a prefix arg, jump to the type definition."
   (interactive "P")
   (let ((cb (lambda (response)
               (tide-on-response-success response
                 (let ((filespan (car (plist-get response :body))))
-                  (tide-jump-to-filespan filespan tide-jump-to-definition-reuse-window))))))
+                  ;; if we're still at the same location...
+                  ;; maybe we're a abstract member which has impementations?
+                  (if (and (not arg)
+                           (tide-filespan-is-current-location-p filespan))
+                      (tide-jump-to-implementation)
+                    (tide-jump-to-filespan filespan tide-jump-to-definition-reuse-window)))))))
     (if arg
         (tide-command:typeDefinition cb)
       (tide-command:definition cb))))
+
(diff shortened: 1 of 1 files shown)


josteink:
Anyone have any opinions on this patch? Or should I just merge?

CC: @fbergroth @ananthakumaran

ananthakumaran:
I am a bit torn on this issue. Is jumping to implementation expected when you want to jump to the definition? Probably others can provide their opinions.

josteink:
To be clear: It will only jump to implementation(s) if it's *already* at the definition.

Take the following example (where `_` denotes cursor)

````typescript
let stack: IStack;
...
let item = stack._pop();
````

Pressing `M-.` here (at `pop`) will take you to the declaration for the method on the interface itself (again with `_` denoting cursor-position):

````typescript
interface IStack {
    _pop(): any;
    push(item: any);
}
````

Now, in my opinion, when I click `M-.` at any item in Emacs, it's because I want to get closer to the *code* which gets invoked when running whatever is at the current point.

With the current implementation in tide, pressing `M-.` at this point will do nothing (it will navigate back to itself).

If you want to dig further into the implementation (that is, the code which gets executed when the line is run), you will then have to bind `tide-jump-to-implementation` to another key (`M-:`, `C-M.`, etc) and start using that instead of `M-.`.

Basically you suddenly need to invoke a *different flow* for what is essentially the same task (digging deeper into the code).

With this patch, once you're at this position (and it should be obvious that going to the *declaration* is going to the place you're already at), we *then* (and only then!) proceed to look for implementations.

I really think this makes sense and is much more intuitive. Speaking for myself at least, I'm always "surprised" when digging into a interface member gets me nowhere further.

If you're still sceptical about that, or that we have some difficulty reaching a consensus (and that's OK), a different approach could be creating a new defun called `tide-jump-to-declaration-or-implementation` or something, and users can be free to rebind `M-.` to that. How would your feelings on that be?

ananthakumaran:
I will do some testing and merge it soon.

josteink:
From what I can tell, [you've already merged this](https://github.com/ananthakumaran/tide/commits/master), but the PR is still open.

Is this a Github-glitch or did you do something wrong? :)

ananthakumaran:
I don't remember how I merged the branch. Might have done rebase or something like that. Anyways closing this.

Issue #133, opened 2017-04-03
Display function overloads?


Dessix:
When the minibuffer shows something like this:
![image](https://cloud.githubusercontent.com/assets/434942/24593874/593964b8-17dc-11e7-8a61-2d108bc9b53b.png)
Is there a way to access the other overloads? The concept of overloads seem to be unmentioned in the source, but it's generally a piece of information I can use when in a new framework. What would need to be done to introduce this feature?

josteink:
Good question.

Looking into it, it seems the code-section involved for this [is fairly small](https://github.com/ananthakumaran/tide/blob/master/tide.el#L734-L740).

My initial assumption was that [`tide-command:signatureHelp`](https://github.com/ananthakumaran/tide/blob/master/tide.el#L702-L708) was responsible for this. And as far as I can see from the contract, the server is [supposed to return several signatures if found](https://github.com/Microsoft/TypeScript/blob/master/src/server/protocol.ts#L1627-L1653).

So edbugging my way through the code, I found out that the function [`tide-method-call-p`](https://github.com/ananthakumaran/tide/blob/master/tide.el#L710-L711) actually doesn't trigger for our eldoc function, meaning the response we get is from the `tide-command:quickinfo` implementation instead.

To me this certainly looks like a bug. If we can fix this bug, getting your issue fixed too should hopefully not be too hard.

@ananthakumaran : From what I can see, this looks like your code. Do you agree with my analysis?

ananthakumaran:
`signatureHelp` returns the different signatures. It's possible to cycle through the different signatures / show them all at once. `quickinfo` doesn't seem to return all the function signatures though.

PR #134, opened 2017-04-04
use async callbacks for eldoc

### LINKED ISSUE #113: frequent Sync request timed out documentHighlights
I keep getting this error message so frequently. What could be the cause?

Here is a copy of the backtrace that appears when it happens: 
```
Debugger entered--Lisp error: (error "Sync request timed out documentHighlights")
  signal(error ("Sync request timed out documentHighlights"))
  error("Sync request timed out %s" "documentHighlights")
  tide-send-command-sync("documentHighlights" (:file "~/projects/app/src/App.ts" :line 1 :offset 1 :filesToSearch ("~/projects/app/src/App.ts")))
  tide-command:documentHighlights()
  tide--hl-identifier()
  tide-hl-identifier()
  tide--hl-identifiers-function()
  apply(tide--hl-identifiers-function nil)
  timer-event-handler([t 0 0 100000 t tide--hl-identifiers-function nil idle 0])
```

ananthakumaran:
This removes another sync call and will probably fix #113 and could help with #133
\cc @josteink

### DIFF
--- a/tide.el
+++ b/tide.el
@@ -170,20 +170,16 @@ above."
 (defun tide-unsupported-p (command)
   (member command (gethash (tide-project-name) tide-tsserver-unsupported-commands '())))
 
-(defmacro tide-fallback-if-not-supported (new-command new old)
+(defmacro tide-fallback-if-not-supported (new-command new old cb)
   `(if (tide-unsupported-p ,new-command)
-       (let ((response (,old)))
-         (when (tide-response-success-p response)
-           response))
-     (let ((response (,new)))
-       (if (tide-command-unknown-p response)
-           (progn
-             (tide-mark-as-unsupported ,new-command)
-             (let ((response (,old)))
-               (when (tide-response-success-p response)
-                 response)))
-         (when (tide-response-success-p response)
-           response)))))
+       (,old ,cb)
+     (,new
+      (lambda (response)
+        (if (tide-command-unknown-p response)
+            (progn
+              (tide-mark-as-unsupported ,new-command)
+              (,old ,cb))
+          (funcall ,cb response))))))
 
 ;;; Helpers
 
@@ -231,6 +227,12 @@ ones and overrule settings in the other lists."
      (-when-let (msg (plist-get response :message))
        (message "%s" msg))
      nil))
+
+(defmacro tide-on-response-success-callback (response &rest body)
+  (declare (indent 1))
+  `(lambda (,response)
+     (tide-on-response-success ,response
+       ,@body)))
 
 (defun tide-join (list)
   (mapconcat 'identity list ""))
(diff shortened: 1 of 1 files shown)


tide.el
```diff
@@ -358,8 +360,7 @@ LINE is one based, OFFSET is one based and column is zero based"
          (payload (concat encoded-command "\n")))
     (process-send-string (tide-current-server) payload)
     (when callback
-      (puthash request-id (cons (current-buffer) callback) tide-response-callbacks)
-      (accept-process-output nil 0.01))))
+      (puthash request-id (cons (current-buffer) callback) tide-response-callbacks))))
```
josteink: Is there any particular reason for this statement to be removed? Was it redundant and not needed? Does it get in the way for async code-patterns?

josteink:
Sorry for the slow response. I didn't catch the CC and assumed someone else with more time would gloss over this :)

A few questions (apart from the one in the diff): From what I can tell, your basic change here is converting lots of synchronous functions and code-paths into asynchronous ones. There's no other functional changes, right?

For this to impact #113, I'm guessing that async code-paths do not suffer the same time-out issues as synchronous ones does? Is that correct? If so, this looks like a reasonable fix to me.

As for assisting #133, I assume the new helper-function `tide-eldoc-maybe-show` is meant to give us a new mechanism to incrementally cycle through signatures? Is that correct, or are there other changes you've done w.r.t. that particular issue?

Otherwise it all looks good to me, and from what I can tell eldoc still works like it used to, so I don't see anything broken :)

tide.el
ananthakumaran: This adds bit of uncertainity. with accept-process-output, the callback could get executed sync(if the server responds within 0.01) or asyc. This was not a problem in other use cases. For eldoc, the `tide-eldoc-function` always returns nil. If the callback gets executed synchronously, then the value returned by tide-eldoc-function will overwrite the value set by callback. Removing `accept-process-output` will make sure the callback always get executed after the `tide-eldoc-function` returns.

ananthakumaran:
> For this to impact #113, I'm guessing that async code-paths do not suffer the same time-out issues as synchronous ones does? Is that correct? If so, this looks like a reasonable fix to me.

That's my guess also

> As for assisting #133, I assume the new helper-function tide-eldoc-maybe-show is meant to give us a new mechanism to incrementally cycle through signatures? Is that correct, or are there other changes you've done w.r.t. that particular issue?

If we decide to cycle through the signatures, this might be useful, because we have inverted the control and can make changes to the content shown by the eldoc.

josteink:
> That's my guess also

If so, and everything else works, I guess there's no harm in trying and seeing if it does indeed work as expected.

Should we ask @johnbendi to verify before committing these changes? Or do we consider the sync-to-async a general improvement which may improve responsiveness, so we want them regardless?

>  we have inverted the control and can make changes to the content shown by the eldoc

I missed that bit, but that definitely sounds useful. I'm all for it.

Either way, all in all, LGTM. Merge at will.

PR #135, opened 2017-04-06
Don't convert location to point unnecessarily.

lddubeau:
The previous version of this function would always convert the location to point before doing the comparison. When the location was to a point that is greater than the size of the current buffer, it would cause a fatal error. The new version will do the conversion of location to point only if the current buffer's file name and the other file name are a match.

### DIFF
--- a/tide.el
+++ b/tide.el
@@ -529,10 +529,9 @@ implementations.  When invoked with a prefix arg, jump to the type definition."
 
 (defun tide-filespan-is-current-location-p (filespan)
   (let* ((location (plist-get filespan :start))
-         (new-file-name (plist-get filespan :file))
-         (new-point (tide-location-to-point location)))
-    (and (equal new-point (point))
-         (string-equal new-file-name buffer-file-name))))
+         (new-file-name (plist-get filespan :file)))
+    (and (string-equal new-file-name buffer-file-name)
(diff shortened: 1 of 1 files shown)


josteink:
Hey Louis. Thanks for another patch!

I'm always happy when I see you sending along a PR, but on this one I have to admit I had to look over the code a few times trying to figure out what the *real* change was. I just couldn't see what it was supposed to fix.

Re-reading your PR-text, it made it fairly obvious in retrospect, but looking at the code alone, it really becomes an "accident" of implementation which can be easy to mess up later, some time in the future.

To prevent regressions, I would therefore request at least one of the following two things, although both would be stellar :)

* That the code be commented to include a short-hand version of your PR-text (`;; don't calculate location before we need to. may cause overflow!` or something like that)
* Would it be possible to write a test-case for this so that if the comment's get disregarded, or we in some other way fudge this up, we can still verify that we haven't ruined the core functionality? Could you be possible to author some test-document where this can functionality can be verified without *directly* referencing the functions involved?

Essentially just to ensure that fixed bugs stay fixed. I happen to like it that way :)

Cheers

ananthakumaran:
This is a classic mistake, trying to use something before making sure it's safe to do so. I will merge. Please open another PR if you want to add tests.

PR #143, opened 2017-05-06
Filter & sort completion candidates by kind

katspaugh:
This PR solves two issues with the code completion.

### 1. Member completion

When completing properties and methods of an object, Tide would list all possible symbols as candidates.

_Edit_: I've just realized this behavior can be reproduced only for JavaScript files, not if I put the same code into a `.ts` file. 🤷‍♂️ 

Before the patch:

screen shot 2017-05-06 at 17 32 34

With the patch – we display only the properties and methods of the object at point:

screen shot 2017-05-06 at 17 35 12

### 2. Candidate sorting
When completing a variable, Tide provides all possible symbols in the scope, sorted alphabetically.

It would be nicer to sort the candidates by scope. First, show the function parameters and locally defined variables, then local functions and outer variables, and finally library functions (i.e. declarations from typings).

Before:

screen shot 2017-05-06 at 17 32 21

After:

screen shot 2017-05-06 at 17 39 28

---

What do you think? I would be glad to incorporate any feedback or discuss a better solution. Thanks!

### DIFF
--- a/tide-tests.el
+++ b/tide-tests.el
@@ -22,6 +22,7 @@
     (package-install p)))
 
 (require 'tide)
+(require 'dash)
 
 ;; actual tests:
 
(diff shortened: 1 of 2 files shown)


josteink:
Hey there and thanks for the pull!

I think these are some good ideas, and I absolutely welcome them.

I'll try to get some proper testing and reviewing done tomorrow, and provide some general feedback tomorrow :)

josteink:
Having done some basic, manual testing, all I can say is that I find this a positive and productivity-improving change.

Being sucker for automated testing though, I'd also love to see the sorting-rules formalized in the form of a small set of tests.

If we manage to sneak in a few automated tests for new features we add, it's much easier to ship subsequent updates later with good confidence. And that's an important point for me.

And in that regard, the code you've provided seems very test-friendly, and such tests could/should be written directly against `tide-compare-completions`, with no more fuzz about it.

If you could provide that too, I would happy to merge this change.

katspaugh:
Fair enough!

I'll use [company-mode's tests](https://github.com/company-mode/company-mode/tree/master/test) as a guide (they use [ERT](https://www.gnu.org/software/emacs/manual/html_node/ert/)). Since Tide doesn't have a CI set up yet, I think just adding a test command into the Makefile would be OK?

Thanks!

josteink:
> Since Tide doesn't have a CI set up yet

Oh my. @ananthakumaran: Could you have this project set up with Travis for us? :)

> I'll use company-mode's tests as a guide ... I think just adding a test command into the Makefile would be OK?

That would be excellent. You could also consider taking a look at how we do things in [typescript.el](https://github.com/ananthakumaran/typescript.el/blob/master/typescript-mode-tests.el) (which has a CI-build!) to ensure that we're consistent across projects.

ananthakumaran:
Enabled travis for this repo. Just add .travis.yml file, it should work.

katspaugh:
Awesome, thanks, guys!

josteink:
I've added a basic setup which basically runs ERT, but it still doesn't manage tide's dependencies, so the test is pretty much a no-op.

But at least it should be something to build on, without the need to start from scratch.

josteink:
And **there** you should have a working test/CI-environment. Have fun :)

katspaugh:
I've added two tests for filtering and sorting. I hope they are OK!

josteink:
Awesome. Thanks for putting in the effort. Looks good to me! :)

sandersn:
Both these behaviours are intentional on the part of the Typescript language service. I think this PR should be reverted. Failing that, both should be customizable, and default to the language service behaviour.

1. In Javascript files, the language service suggests any token that it has seen for completion. This is similar to what other Javascript editors do.
2. Typescript and Javascript don't distinguish between function and constant declarations, so sorting by kind separates things that are semantically similar. Plus, non-alphabetical sorting just messes up completion. Here's an example from the Typescript codebase itself:

```ts
namespace Debug {
  export function assert(condition: boolean, message?: string) { ... };
  export var assertEachNode = shouldAssert(AssertionLevel.Normal) ?
    (nodes: Node[], test: (node: Node) => boolean): void { ... } :
    () => { };
}
Debug.asser/**/
```

Requesting completion at `/**/`, by typing `(` in my case, completes to `assertEachNode`, when it should complete to `assert`.

Arguably, emacs users who use Javascript can just choose some other javascript-mode to get the behaviour of (1). I'm biased, since I work on Typescript, but I would like tide to be the default mode for Javascript programmers too.

I'll send out a PR when I have time that at least adds custom variables to return the behaviour to the intended defaults.

josteink:
While I think outright reverting this is a bit much, I'm open to changes. Let's discuss the approach we should take in your PR.

PR #147, opened 2017-05-09
Remove duplicate lib.scriptHost.d.ts

aaronjensen:
There were two of these files w/ different file name casings. The remaining one is the one that is still in [Typescript/lib](https://github.com/Microsoft/TypeScript/tree/master/lib). This caused annoyances on my case-insensitive file system.


josteink:
Ooops. That's probably my fault, from the last tsserver-update. Sorry about that!

I just checked what the output from `make update-tsserver` produces, and indeed, the one with the capitalized `h` should go.

Thanks for catching that one and fixing it for us :)

Issue #145, opened 2017-05-09
Duplicate entries in imenu


Eoksni:
After running `(tide-setup)` I get duplicate entries for each function in `imenu`:
![image](https://cloud.githubusercontent.com/assets/2010503/25843825/c98f1542-34d2-11e7-8f96-35d04df26be2.png)

josteink:
To me it sounds like some code is getting run twice here, and the first thing I'd like to eliminate is that it could be due to any local configuration issues.

If you start Emacs without any site files or custom config (`emacs -Q`) can you still reproduce this issue?

Eoksni:
with this:
`emacs -Q check-imenu.js`
```elisp
(package-initialize)
(tide-setup)
(ivy-mode)
```
`M-x imenu`
I get duplicates, but if I turn off `ivy-mode` all is fine and duplicates are gone. `counsel-imenu` (its an extension to ivy) also shows duplicates. 

I guess then it is a bug in ivy? Seems super weird, never had any dup problem with ivy in other places.

josteink:
I've tested with helm, and I get the same results. But looking at our imenu-code, it's really short and to the point:

https://github.com/ananthakumaran/tide/blob/master/tide.el#L1106-L1121

What I do suspect is that something shows up twice, because the this entity may be reported several time as a *different* kind of entity.

I therefore tried to attach a debugger at the first point where we start processing data after receiving them: https://github.com/ananthakumaran/tide/blob/master/tide.el#L1106-L1109

And that's the smoking gun. Basically `hello()` is (by tsserver) reported as a local function (in scope) and also a function nested in the global scope. This results in a duplicate when we flatten the list. And we don't do any work to try to eliminate those duplicates.

In the meantime, I've never had issues with duplicates when working with bigger modules where I've been "proper" about structuring things in classes etc.

I don't have time to create a fix anytime soon, so if you feel like sending in a patch, that's cool, and I'll be more than happy to review it.

ananthakumaran:
@Eoksni can you test it again? it should be fixed in master.

Eoksni:
Yes, it works! Big thanks.

PR #151, opened 2017-05-17
cleanup imenu

### LINKED ISSUE #5: Support for helm-semantic-or-menu ?
It would be nice if tide would support helm-semantic-or-menu to display a list of current functions and classes in the buffer.

ananthakumaran:
* Don't flatten the tree by default
* during flatten step, prefix the parent node's text before the child
  node's text

fixes #5 #145

### DIFF
--- a/tide.el
+++ b/tide.el
@@ -115,6 +115,11 @@ above."
   :type 'boolean
   :group 'tide)
 
+(defcustom tide-imenu-flatten nil
+  "Imenu index will be flattened if set to non-nil."
+  :type 'boolean
+  :group 'tide)
+
 (defmacro tide-def-permanent-buffer-local (name &optional init-value)
   "Declare NAME as buffer local variable."
   `(progn
@@ -1136,23 +1141,35 @@ number."
 
 ;;; Imenu
 
-(defun tide-flatten-navitem (items)
-  (if items
-      (nconc items (apply #'nconc (-map (lambda (item) (tide-flatten-navitem (plist-get item :childItems))) items)))
-    '()))
+(defun tide-flatten-imenu-index (index &optional parent)
+  (when (consp index)
+    (let* ((name (car index))
+           (new-name (if parent (concat parent imenu-level-separator name) name)))
+      (if (and (consp (cdr index)) (listp (cdr index)))
+          (-map (lambda (i) (tide-flatten-imenu-index i new-name)) (cdr index))
+        (cons new-name (cdr index))))))
+
+(defun tide-build-imenu-index (navbar-item)
+  (let ((child-items (plist-get navbar-item :childItems))
+        (text (plist-get navbar-item :text)))
+    (if child-items
+        (cons text
+              (-map #'tide-build-imenu-index child-items))
+      (cons (concat text " " (propertize (plist-get navbar-item :kind) 'face 'tide-imenu-type-face))
+            (tide-span-to-position (plist-get (car (plist-get navbar-item :spans)) :start))))))
 
 (defun tide-command:navbar ()
   (tide-send-command-sync "navbar" `(:file ,buffer-file-name)))
 
 (defun tide-imenu-index ()
   (let ((response (tide-command:navbar)))
-    (when (tide-response-success-p response)
-      (-map
-       (lambda (item)
-         (cons (concat (plist-get item :text) " " (propertize (plist-get item :kind) 'face 'tide-imenu-type-face))
-               (tide-span-to-position (plist-get (car (plist-get item :spans)) :start))))
-       (tide-flatten-navitem (plist-get response :body))))))
-
+    (tide-on-response-success response
+      (let ((index (-map
+                    #'tide-build-imenu-index
+                    (plist-get response :body))))
+        (if tide-imenu-flatten
+            (-flatten (-map #'tide-flatten-imenu-index index))
+          index)))))
 
 ;;; Rename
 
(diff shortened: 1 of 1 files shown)


josteink:
Looks interesting. I'll check it out more in depth tomorrow!

josteink:
Looking at `imenu` now, I seem to get lots more hits than I used to.

* My imports
* My declared classes
* Functions in classes
* declared consts inside those functions

I can't see how this could happen as a result of your changes though. Is this new `tsserver` behaviour? Colour me surprised :)

Anyway: I can definitely see that without setting the `defcustom` to preserve old behaviour we now get lots of things reported, and only inline functions reported as functions (rest reported as `method`).

It seems to work *mostly* as advertised, but from what I can tell doesn't provide information about nested members: It only shows one parent type.

Example: If you have a `const` declared in a `method` declared in a `class`, the imenu index for that const will show it a parent of method. Is this intentional?

ananthakumaran:
> It seems to work mostly as advertised, but from what I can tell doesn't provide information about nested members: It only shows one parent type.

I tried three nested module, tsserver seems to hoist the second one to top level. at most there seem to be two levels.

ananthakumaran:
> I can't see how this could happen as a result of your changes though. Is this new tsserver behaviour? Colour me surprised :)

not sure about this.

josteink:
> I tried three nested module, tsserver seems to hoist the second one to top level. at most there seem to be two levels.

I'm sure that info would be able to reconstruct somehow, but I'm guessing that's going to be both error-prone and cumbersome.

For reference, I've done [something similar for csharp-mode](https://github.com/josteink/csharp-mode/blob/master/csharp-mode.el#L1979-L2146), but as you can see it's quite a lot of code, still has its limitations and if you check [csharp-mode's history](https://github.com/josteink/csharp-mode/search?q=imenu&type=Commits&utf8=%E2%9C%93), you'll see lots of bugs and fixes related to `imenu` before finally landing something "stable".

If you want to go that extra mile, I fully support you (and in that case very much recommend making a good suite of tests!), but if you decide that for now it doesn't seem like a worthwhile effort, I have no problem with that either.

All in all, I think it's fair to say that this represent a genuine improvement, especially for scenarios like having multiple implementations of the same interface in the same file.

I have only two concerns:

* If you don't have a completion-framework and you were used to being able to `M-x imenu` followed by `beggingOfPropName<TAB>`, that would allow you to quickly navigate to that property, no matter what class it was in. Now you will first need to complete the class with a `beggingingOfClassName<TAB><RET>` before you can navigate to the property. This may break people's flow, but I guess it's unavoidable when we use nesting. People who don't like it can re-enable flattening with the new `defcustom`, so it's not *critical*.

* Right now we're reporting so many things, that it's starting to get very noisy without a good filtering completion-framework (like ivy or helm). If we're redoing imenu, would it be wise to consider filtering out some navigation options? For instance... I never once feel like using imenu to navigate to my imports at the top of a file. Same with in-method `cons`-occurances. I use cons everywhere I want data to stay immutable, so I see this everywhere in `imenu` now. We really should clean this up a little.

Unrelated to PR at hand, but I'm also noting that we're actually not creating a imenu *menu*. Do you have any idea if that would be much work to add?

ananthakumaran:
>> I tried three nested module, tsserver seems to hoist the second one to top level. at most there seem to be two levels.

> I'm sure that info would be able to reconstruct somehow, but I'm guessing that's going to be both error-prone and cumbersome.

There are two commands available `navbar` and `navtree`, the later seems to preserve the depth and doesn't hoist things to top level. see https://github.com/Microsoft/TypeScript/issues/11230#issuecomment-250429110

I will update my patch to use `navtree`

> If you don't have a completion-framework and you were used to being able to M-x imenu followed by beggingOfPropName<TAB>, that would allow you to quickly navigate to that property, no matter what class it was in. Now you will first need to complete the class with a beggingingOfClassName<TAB><RET> before you can navigate to the property. This may break people's flow, but I guess it's unavoidable when we use nesting. People who don't like it can re-enable flattening with the new defcustom, so it's not critical.

I am not sure we can do much to improve the vanilla imenu.

> Right now we're reporting so many things, that it's starting to get very noisy without a good filtering completion-framework (like ivy or helm). If we're redoing imenu, would it be wise to consider filtering out some navigation options? For instance... I never once feel like using imenu to navigate to my imports at the top of a file. Same with in-method cons-occurances. I use cons everywhere I want data to stay immutable, so I see this everywhere in imenu now. We really should clean this up a little.

Probably we can try to handle it another PR

> Unrelated to PR at hand, but I'm also noting that we're actually not creating a imenu menu. Do you have any idea if that would be much work to add?

I never use menu in emacs :), So not much motivation to do it now.

josteink:
> I will update my patch to use navtree

Just tested and this does feel much better.

> Probably we can try to handle it another PR

We definitely should, because it still feels pretty cluttered.

At least now, with the hierarchy maintained, now, it's easier to see what all the items returned actually represent, so I still consider it an improvement.

ananthakumaran:
> We definitely should, because it still feels pretty cluttered.
Yes, but the config requires some thought, I am not sure we should go by a whitelist, blacklist using kind or something else. There is also the question of how we should handle the tree hierarchy when filtering, should all the childItems filtered if the parent is not allowed etc.

Please merge this if you think it will improve the current one, we can come back to filtering later.

josteink:
Fair enough. I agree. Merged :)

josteink:
Fair enough. I kinda agree. Merged :)

PR #152, opened 2017-05-18
use magit/hydra like popup for choice

ananthakumaran:
falls back to the old model when there are more than 7 choices.

screenshots
screen shot 2017-05-18 at 9 39 07 am
screen shot 2017-05-18 at 9 38 34 am

### DIFF
--- a/tide-lv.el
+++ b/tide-lv.el
@@ -1,0 +1,89 @@
+;;; tide-lv.el --- Other echo area
+
+;; Copyright (C) 2015  Free Software Foundation, Inc.
+
+;; Author: Oleh Krehel
+
+;; This file is part of GNU Emacs.
+
+;; GNU Emacs is free software: you can redistribute it and/or modify
+;; it under the terms of the GNU General Public License as published by
+;; the Free Software Foundation, either version 3 of the License, or
+;; (at your option) any later version.
+
+;; GNU Emacs is distributed in the hope that it will be useful,
+;; but WITHOUT ANY WARRANTY; without even the implied warranty of
+;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+;; GNU General Public License for more details.
+
+;; You should have received a copy of the GNU General Public License
+;; along with GNU Emacs.  If not, see <http://www.gnu.org/licenses/>.
+
+;;; Commentary:
+;;
+;; This package provides `tide-lv-message' intended to be used in place of
+;; `message' when semi-permanent hints are needed, in order to not
+;; interfere with Echo "Я тихо-тихо пiдглядаю,
+;;     І тiшуся собi, як бачу то,
+;;     Шо страшить i не пiдпускає,
+;;     А iншi п’ють тебе, як воду пiсок." L.V.
+;;
+;;
+;; Modified version of lv.el from github.com/abo-abo/hydra
+
+;;; Code:
+
+(defvar tide-lv-buffer-name " *tide-LV*")
+
+(defvar tide-lv-wnd nil
+  "Holds the current LV window.")
+
+(defun tide-lv-window ()
+  "Ensure that LV window is live and return it."
+  (if (window-live-p tide-lv-wnd)
+      tide-lv-wnd
+    (let ((ori (selected-window))
+          buf)
+      (prog1 (setq tide-lv-wnd
+                   (select-window
+                    (let ((ignore-window-parameters t))
+                      (split-window
+                       (frame-root-window) -1 'below))))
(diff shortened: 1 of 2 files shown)


MarcusRiemer:
I love it!

josteink:
This change breaks integration with helm/ivy and other completion frameworks, if the user has chosen to use one of those.

As such, I would like for this change:

* to be something you could enable/disable using a `defcustom`.
* even better: if there's a way to probe if a completion-framework is active (maybe by checking what `completing-read` is aliased to ?), then only activate hydra if no such framework is active.

There are lots of users of completion frameworks, and for them, in its current form this change is going to cause a degraded/inconsistent experience.

josteink:
I was a bit busy and stressed earlier to day, so my comment may have come off as more negative than I actually intended to.

So let me try all over again, with the more constructive hat on :)

Facts:

* On its own, I think it's a great addition. Compared to plain completing-read this is a night and day difference. Good job!
* I haven't played around with it too much, but I expect to work well, much like magit does.
* But it does not play well with other completion-frameworks, and this [breaks people's existing workflows](https://xkcd.com/1172/). This is something we should avoid at all cost, if possible.
* Right now we risk have several different and inconsistent completion techniques based on the number of results returned. This makes it *impossible* to create a consistent workflow.

Therefore, to help consistency, I propose the following changes:

* Hydra needs to be on or off *globally*. If we find a way to detect completion frameworks, the presence of such, should disable the hydra.
* If detection can't be done and we need to resort to a *defcustom*, we should definitely leave the default *enabled* to enhance discovery. People with completion-frameworks will probably complain, and for those we can direct them to the appropriate settings.
* The hydra should be used *always* when we have more than 1 result. I shouldn't have to guess how many items my query resulted in, to know which finger-memory to apply. The way it works now will super-quickly become annoying.

These changes should be fairly small and easy to implement, but I think they will make a big impact on how this change *feels* when getting down deep in a code-session and you don't want that "flow" interrupted.

Opinions?

ananthakumaran:
My main motivation behind this is to improve the ergonomics of tide-fix. In most common cases, the number of options would be less than 5, so, completing read adds bit of overhead. Jump to implementation was added later because both of these used the same function.

> If detection can't be done and we need to resort to a defcustom, we should definitely leave the default enabled to enhance discovery. People with completion-frameworks will probably complain, and for those we can direct them to the appropriate settings.

I definitely agree with the defcustom, just not sure about whether to enable/disable it by default. Both has it's upsides and downsides.

> The hydra should be used always when we have more than 1 result. I shouldn't have to guess how many items my query resulted in, to know which finger-memory to apply. The way it works now will super-quickly become annoying.

This is not possible I guess. the popup is kind of good for up to 7 maybe 10, after that It will take too much screen space. But my assumption is, this should happen very rarely if ever.

josteink:
> My main motivation behind this is to improve the ergonomics of tide-fix.

That's fair enough. I think particularly for `tide-fix` that without a completion-framework, completing-read may not be the best fit for a function like this.

If we constrain the usage of the hydra to this part of the code, I would have no trouble always having it on, even at the cost of regular completion frameworks. I may not even bother with a `defcustom`.

> Jump to implementation was added later because both of these used the same function.

From my testing, I see that this also affects creating instances of interfaces.  This can get problematic because in typescript interfaces are structural, and often used for message-passing in cases where you would use classes in languages like C# or Java.

Example below:

````typescript
export interface DemoMessage {
    hasLotsOfUsage: boolean;
    message: string;
}

let var1 : DemoMessage = { hasLotsOfUsage: true, message: "Foo" };
let var2 : DemoMessage = { hasLotsOfUsage: false, message: "Bar" };
````

In this case, both of these variables will show up as "implementations" of that interface.

>  In most common cases, the number of options would be less than 5, so, completing read adds bit of overhead.

In the case above, I can guarantee you that with a good re-use of interfaces, that it's not going to be rare to have much more than 5 here. Maybe you'll get 20, maybe 100. Designing for 10 max is definitely going to lead to issues.

While I'm not opposed to this feature, I think we need to think through how it can be best employed and where.

ananthakumaran:
@josteink added an option to disable/enable at a command level. I am not very sure about the default values. If you think we should disable `jump-to-implementation`, I am also OK with it.

josteink:
I honestly think we should disable it for `jump-to-implementation`. As far as I can tell, this also addresses always using it (and not just sometimes) when configured (and not overflowing the alphabet)?

If so, I'm all OK with this :)

ananthakumaran:
removed `jump-to-implementation`. Please merge if are OK with the changes.

josteink:
Looks good to me :)

PR #153, opened 2017-05-19
Allow completion in strings

asmundg:
This lets us complete literal string unions. E.g.

```typescript
type Foo = "FOO" | "FOOZ"
const x: Foo = "F"
````


asmundg:
Oops, I was looking at an old branch.

Issue #159, opened 2017-06-02
`tslint.json` seems to have no effect


yatesco:
***EDIT: Please read the comment thread - I think I might have fixed the issue, but advise welcome***

Hi - first, this is almost certainly a bug in my understanding/expectations. I have a typescript ReactJS app and have configured `tslint.json` sufficiently for Visual Studio Code to light up like a Christmas tree :-). For example, I have a rule stating that strings should be single and not double quotes.

I opened the project up in Emacs, configured `flycheck` and `tide` such that _some_ things are picked up by `flycheck` (e.g. type errors, unused variables) but nowhere near as many as VSCode.

If I run `tslint` from the command line and specify the `project` and `config` properties (e.g. ```tslint --type-check -p ./tsconfig.json -t json --config ./tslint.json```)  then I get the same set of errors I see in emacs. If I just run it without the `config` property (e.g. ```tslint --config ./tslint.json --format json src/page/CentralFragment.tsx```) then I get all of the errors I see in VSCode however, the output is full of warnings about rules needing to be type checked:

``` json
Warning: The 'promise-function-async' rule requires type checking
Warning: The 'await-promise' rule requires type checking
Warning: The 'no-floating-promises' rule requires type checking
Warning: The 'no-for-in-array' rule requires type checking
Warning: The 'no-inferred-empty-object-type' rule requires type checking
Warning: The 'no-unbound-method' rule requires type checking
Warning: The 'no-unsafe-any' rule requires type checking
Warning: The 'no-unused-variable' rule requires type checking
Warning: The 'no-use-before-declare' rule requires type checking
Warning: The 'no-void-expression' rule requires type checking
Warning: The 'restrict-plus-operands' rule requires type checking
Warning: The 'strict-boolean-expressions' rule requires type checking
Warning: The 'strict-type-predicates' rule requires type checking
Warning: The 'use-default-type-parameter' rule requires type checking
Warning: The 'completed-docs' rule requires type checking
Warning: The 'deprecation' rule requires type checking
Warning: The 'match-default-export-name' rule requires type checking
Warning: The 'no-boolean-literal-compare' rule requires type checking
Warning: The 'no-unnecessary-qualifier' rule requires type checking
Warning: The 'no-unnecessary-type-assertion' rule requires type checking
Warning: The 'return-undefined' rule requires type checking
[...]

PR #161, opened 2017-06-02
type and lint errors in web-mode

yatesco:
https://github.com/ananthakumaran/tide/issues/159

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,6 +111,38 @@ true.
 ;; configure jsx-tide checker to run after your default jsx checker
 (flycheck-add-next-checker 'javascript-eslint 'jsx-tide 'append)
 ```
+### Both type errors and `tslint` errors
+`web-mode` doesn't display _all_ the `tslint` errors without a bit of jiggery pokery. Using `typescript-mode` will show all the errors, however to see both type and linting errors in `web-mode` you need to do the following:
+
+``` elisp
+  (use-package tide :diminish t)
+  (defun setup-tide-mode ()
+    (interactive)
+    (tide-setup)
+    (flycheck-add-next-checker 'typescript-tide '(t . typescript-tslint) 'append)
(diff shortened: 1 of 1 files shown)


ananthakumaran:
I guess the changes are too verbose and the real issue is with flycheck/tslint-checker configuration. Probably it would best to add the config under the TSX and JSX section and may be explain or point to how chaining and checkers work in flycheck.

josteink:
I don't mind a verbose section as long as it gives a working config. 

If you think it's too verbose to be on the readme, would it perhaps be better to add functions like tide-configure-tslint and tide-configure-web-mode and reference those in the readme?

ananthakumaran:
This just repeats the same config already mentioned in the readme with some minor additions
```elisp
(flycheck-add-mode 'typescript-tslint 'web-mode)
```
Speaking strictly, the issue has nothing to do with tide. But as tide seems to be the starting point of config for many users, we could add it in the readme with a note. 

> would it perhaps be better to add functions like tide-configure-tslint and tide-configure-web-mode and reference those in the readme?

as I mentioned above, this is not tide issue and as such we shouldn't try to fix it in tide

yatesco:
Hi all - comments appreciated. To be blunt, I am dancing around in the dark here, and I imagine I might not be the only one. For example, spacemacs exhibits exactly the same behaviour.

I get completely that this is not `tide`'s insufficiency, rather it is a lack of knowledge around `flycheck`.

Would it be sufficient to reduce this to a note:

> To get both `tide` and `tslint` errors (as seen in `typescript-mode`) simply add ```(flycheck-add-mode 'typescript-tslint 'web-mode)``` to your config?

What do you think?

josteink:
Sounds good to me :)

ananthakumaran:
Closing this. Updated the readme with the details.

PR #162, opened 2017-06-09
make sure texts are inserted in the end

ananthakumaran:
if the cursor inside the *-errors* buffer is moved while the insert is
going on, text gets inserted in the wrong order.

### DIFF
--- a/tide.el
+++ b/tide.el
@@ -1520,36 +1520,38 @@ code-analysis."
            (last-file-name nil))
       (tide-set-event-listener
        (lambda (response)
-         (let ((inhibit-read-only t)
-               (file-name (tide-plist-get response :body :file))
-               (diagnostics (tide-plist-get response :body :diagnostics)))
-           (pcase (plist-get response :event)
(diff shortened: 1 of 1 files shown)


tide.el
```diff
@@ -1520,36 +1520,38 @@ code-analysis."
...
-                   (insert ": ")
-                   (insert (plist-get diagnostic :text))
-                   (insert "\n")))))
-           (when (and (null syntax-remaining-files) (null semantic-remaining-files))
-             (insert (format "\n%d syntax error(s), %d semantic error(s)\n" syntax-errors semantic-errors))
-             (goto-char (point-min))
-             (tide-clear-event-listener)))))))
+         (save-excursion
+           (goto-char (point-max))
+           (let ((inhibit-read-only t)
```
josteink: So apart from white-space, these two lines here are the only effective change, right?

ananthakumaran: yes, rest are indendation

josteink:
Looks good to me.

Issue #169, opened 2017-06-19
Function documentation does not use correct line terminators for platform


raxod502:
When I request the documentation for `lodash.indexOf`, I get the following:

![image](https://user-images.githubusercontent.com/6559064/27306529-cc7a3a14-54fa-11e7-9108-403bc646604a.png)

Note the `^M` carriage-return characters at the end of each line. It seems to me this is a bug, although I'm not sure where since I'm not familiar with the workings of tsserver.

System information:

```
                -/+:.           rrosborough@[redacted]
               :++++.           -----------------------------
              /+++/.            OS: macOS Sierra 10.12.5 16F73 x86_64
      .:-::- .+/:-``.::-        Model: MacBookPro11,4
   .:/++++++/::::/++++++/:`     Kernel: 16.6.0
 .:///////////////////////:`    Uptime: 20 days, 4 hours, 37 mins
 ////////////////////////`      Packages: 130
-+++++++++++++++++++++++`       Shell: zsh 5.3.1
/++++++++++++++++++++++/        Resolution: 1920x1200, 1440x900
/sssssssssssssssssssssss.       DE: Aqua
:ssssssssssssssssssssssss-      WM: Quartz Compositor
 osssssssssssssssssssssssso/`   Terminal: iTerm2
 `syyyyyyyyyyyyyyyyyyyyyyyy+`   Terminal Font: InputMono-Medium 12
  `ossssssssssssssssssssss/     CPU: Intel i7-4980HQ (8) @ 2.80GHz
    :ooooooooooooooooooo+.      GPU: Intel Iris Pro
     `:+oo+/:-..-:/+o+/-        Memory: 4768MiB / 16384MiB
```

josteink:
Last time I checked into a seemingly [related bug](https://github.com/ananthakumaran/tide/issues/122), I found that no matter what the platform-specific line-ending is, Emacs always expects `\n` and `\n` **only**.

Could it be we need to do some cleanup/normalization of the doc-string returned from tsserver as well?

ananthakumaran:
If we go down that route, we should probably introduce `tide-insert` which normalizes line ending and use it everywhere instead of `insert`

josteink:
We'd only need to use it whenever we are inserting data returned by tsserver where we expect line-shifts though.

Using it everywhere regardless of the data-context might incur a needless performance cost, and I'd like to avoid that.

ananthakumaran:
[...]

PR #170, opened 2017-06-21
Fix problem with incorrect lineshift-format in documentation-buffers

josteink:
This addresses https://github.com/ananthakumaran/tide/issues/169.

cc @raxod502: Could you test and see if this change works for you?

### DIFF
--- a/tide.el
+++ b/tide.el
@@ -316,7 +316,7 @@ LINE is one based, OFFSET is one based and column is zero based"
       (erase-buffer)
       (when string
         (save-excursion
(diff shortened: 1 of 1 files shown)


raxod502:
It works perfectly:

image

Thanks @josteink!

josteink:
Great! And thanks for the quick feedback :)

raxod502:
Indeed. One of the things I value most in an Emacs configuration is being able to pull in upstream changes as quickly as possible. That and getting instant email notifications ;)

PR #172, opened 2017-07-08
add support for refactor command

ananthakumaran:
introduced by https://github.com/Microsoft/TypeScript/pull/15569

### DIFF
--- a/README.md
+++ b/README.md
@@ -145,6 +145,8 @@ at point.
 
 <kbd>M-x tide-fix</kbd> Apply code fix for the error at point.
 
+<kbd>M-x tide-refactor</kbd> Refactor code at point.
+
 ### Features
 
 * ElDoc
@@ -157,6 +159,7 @@ at point.
 * Compile On Save
 * Highlight Identifiers
 * Code Fixes
+* Code Refactor
 
 ### Debugging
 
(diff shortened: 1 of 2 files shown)


josteink:
Looks good upon initial skimming, but I'm on vacation the next 3 weeks and won't be able to properly review this until after that. 

Hope that's ok.

josteink:
I've done some preliminary testing, and as far as I can tell... The `tide-refactor` command does not really work for the supplied sample at all, if in a TypeScript-file.

To make it work, I need to create a **Javascript** file and enable `tide` for my `js-mode` hook too. And to find out that, I needed to dig in various Github issues in the typescript repo.

Is that how it's supposed to work? Or are there refactorings which should be supported in TypeScript too? Right now it's not really very obvious, bordering on undiscoverable.

As such I think we should aim to do better. Would it be possible to add a warning that some refactorings are only available for Javascript-files when tsserver returns that no refactorings are available?

ananthakumaran:
There is only one refactoring available for now and that is only supported for js files.

> As such I think we should aim to do better. Would it be possible to add a warning that some refactorings are only available for Javascript-files when tsserver returns that no refactorings are available?

Well I am ok with adding more info instead of just showing `No refactorings available`, but the caveat is, tsserver could add more refactoring in the future or some plugins could provide extra refactoring options. We can't say for sure.

josteink:
> Well I am ok with adding more info instead of just showing No refactorings available, but the caveat is, tsserver could add more refactoring in the future or some plugins could provide extra refactoring options. We can't say for sure.

How about we display `No refactorings available (A javascript-file may be required)`if we detect that the active major-mode is not `typescript-mode`?

josteink:
So should I just go ahead and make those changes myself?

ananthakumaran:
done

josteink:
LGTM. Merge at will.

PR #177, opened 2017-07-20
ignore empty response incase of eldoc

### LINKED ISSUE #176: ElDoc shows "No content available" in certain places
In certain places, ElDoc shows the message "No content available" instead of showing a function signature. I would expect it to instead show no message at all (because showing this message can obscure other, more useful messages).

I've created [an example repository](https://github.com/raxod502/tide-issue-176) demonstrating the issue. (But you can probably reproduce it just by pasting the code below into your TypeScript project.) Steps to reproduce:

```
$ git clone https://github.com/raxod502/tide-issue-176.git
$ cd tide-issue-176
$ yarn
$ emacs -Q
M-: (require 'tide)
C-x C-f index.ts
M-x tide-setup
M-x eldoc-mode
```

The message shows when point is at the `|`:

```
const foo = [
  function bar() { },|
]
```

The message originally comes from tsserver. Perhaps it should be blacklisted.

ananthakumaran:
fixes #176

### DIFF
--- a/tide.el
+++ b/tide.el
@@ -236,18 +236,19 @@ ones and overrule settings in the other lists."
 (defun tide-tsserver-version-not-supported ()
   (error "Only tsserver 2.0 or greater is supported. Upgrade your tsserver or use older version of tide."))
 
-(defmacro tide-on-response-success (response &rest body)
+(defmacro tide-on-response-success (response ignore-empty-response &rest body)
   (declare (indent 1))
   `(if (tide-response-success-p ,response)
        ,@body
      (-when-let (msg (plist-get response :message))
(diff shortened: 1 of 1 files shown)


ananthakumaran:
@raxod502 could you try this branch

raxod502:
Unfortunately, this still seems to show the message. If it works on your end then I can try to figure out why.

ananthakumaran:
sorry, I was only testing `quickinfo` command, but your example triggers `signatureHelp`. Now added for both of them. Can you test again?

raxod502:
Works great, thanks for the fix.

PR #185, opened 2017-08-15
add support for extract method

ananthakumaran:
https://github.com/Microsoft/TypeScript/pull/16960

### DIFF
--- a/README.md
+++ b/README.md
@@ -145,7 +145,7 @@ at point.
 
 <kbd>M-x tide-fix</kbd> Apply code fix for the error at point.
 
-<kbd>M-x tide-refactor</kbd> Refactor code at point.
+<kbd>M-x tide-refactor</kbd> Refactor code at point or current region.
 
 ### Features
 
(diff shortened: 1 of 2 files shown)


josteink:
Looking at the code, it looks nice and clean. I have no objections at that aspect.

But again, I'm having troubles actually getting the refactoring activated.

Does this too depend on being in `js-mode`? Are there quirks in the implementation which you as a user have to be aware of, and adjust to, in order to make this work? If so, is this something we should document better?

ananthakumaran:
sorry, should have mentioned. It's only available in TypeScript/master. Not yet released. It's supported for both js & ts files. Just select a chunk of code, it should be available
screen shot 2017-08-15 at 5 42 50 pm

As for this patch, the only difference is the proper handling of selected region

josteink:
I just tried with latest tsserver from npm, which I guess exlains why I didn't see any results.

But if this is just staying ahead of the curve and preparing for upcoming functionality, I'm totally OK with that.

LGTM

PR #188, opened 2017-08-25
add post code edit hook

### LINKED ISSUE #168: Allow tide-fix to supress semicolon
I use `tide-fix` all the time is great! But I usually work on project that uses `"semicolon": [true, "never"]`on tslint.json. So I always get the to do the extra step of removing the semicolon. Any workarounds?

ananthakumaran:
```elisp
;; eslint-fix-file function will be called for each buffer that has changed
(add-hook 'tide-post-code-edit-hook #'eslint-fix-file)
```

fixes #168

### DIFF
--- a/tide.el
+++ b/tide.el
@@ -83,6 +83,11 @@ above."
   :type '(choice (const nil) string)
   :group 'tide)
 
+(defcustom tide-post-code-edit-hook nil
+  "Hook run after code edits are applied in a buffer."
+  :type 'hook
+  :group 'tide)
+
 (defvar tide-format-options '()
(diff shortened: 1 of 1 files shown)


josteink:
The code itself looks good to me.

But should we also consider mentioning this in the readme/documentation?

ananthakumaran:
I am not sure at the moment, we don't document all the custom variables and not sure where this fits. But if many users ask about this, probably we could add something in the FAQ?

josteink:
Fair enough. We can document it where it feels natural, later, *if* we find people still come asking.

PR #191, opened 2017-09-01
WIP: ignore huge response from tsserver

ananthakumaran:
Don't merge

@aaronjensen can you try this branch?


josteink:
I'm NOT merging, but just looking at the code... Aren't all your modifications inside `tide-decode-response`, which gets called after we've received and inserted the huge `tsserver`-output into the `tsserver`-buffer?

If so... We'll still need some sort of filter in `tide-net-filter` .

Or is that what you're trying to fix with `(toggle-truncate-lines 1)`? If so, at what point will that start cutting off lines?

ananthakumaran:
I did some crude profiling and looks like `json-read-object` is the one which seems to cause the major slowdown.

```
+ command-execute                                                 173  44%
- ...                                                             135  34%
 - prog1                                                           77  19%
  - progn                                                          77  19%
   - tide-decode-response                                          77  19%
    - save-current-buffer                                          77  19%
     - let                                                         77  19%
      - if                                                         77  19%
       - progn                                                     77  19%
        - tide-dispatch                                            77  19%
         - prog2                                                   77  19%
          - json-read-object                                       77  19%
           - json-read                                             77  19%
            - apply                                                77  19%
             - json-read-object                                    77  19%
              - json-read                                          77  19%
               - apply                                             77  19%
                - json-read-array                                  77  19%
                 - json-read                                       75  19%
                  - apply                                          74  19%
                   - json-read-object                              65  16%
                    + json-read                                    22   5%
                    + json-read-string                             10   2%
                      json-add-to-object                            3   0%
   Automatic GC                                                    58  14%
+ timer-event-handler                                              70  18%
+ redisplay_internal (C function)                                   9   2%
```
I am hoping if we remove the `json-read-object` we don't have to solve the insert issue in `tide-net-filter`. So far from my testing, it seems to have brought down the processing time to acceptable level.

> Or is that what you're trying to fix with (toggle-truncate-lines 1)? If so, at what point will that start cutting off lines?

I wrapped `tide-net-filter` with `with-timer` and can't find any noticeable difference between toggle-truncate-lines either enabled/disabled. may be the performance comes into picture only when the buffer is visible to the user. I just left it because I believe disabled is better than enabled.

```
(defmacro with-timer (title &rest forms)
  "Run the given FORMS, counting the elapsed time.
A message including the given TITLE and the corresponding elapsed
time is displayed."
  (declare (indent 1))
  (let ((nowvar (make-symbol "now"))
        (body   `(progn ,@forms)))
    `(let ((,nowvar (current-time)))
       (prog1 ,body
         (let ((elapsed
                (float-time (time-subtract (current-time) ,nowvar))))
           (when (> elapsed 0.001)
            (message "%s... done (%.5fs)" ,title elapsed)))))))
```

josteink:
Not a very scientific data-point, but I opened a buffer in fundamental-mode and did some interactive testing there.

I created a 100-char text-sequence and copied that. Than I kept yanking by holding in `CTRL-Y`.

Performance is pretty consistent up to 15,000 chars. At around 20,000 chars you can tell it's slowing down. At 30,000 chars performance have started degrading significantly. After this it just gets worse. At 50-60,000 chars insert-performance is IMO unacceptable.

Based on those numbers... Is 100KB still a reasonable cut-off point?

I also still think we may risk locking up on a big enough input (like 900KB), but I guess that's a problem we can try to solve separately later, if we decide we need to do that as well.

ananthakumaran:
> Than I kept yanking by holding in CTRL-Y.

The performance differs substantially depends on whether the buffer is visible to the user or not. Probably a good way to profile is to use elisp to create & insert text.

josteink:
Good point. That way capturing actual times:

````elisp
(defun test-insert-100 (times)
  (with-temp-buffer
    (buffer-disable-undo)
    (dotimes (time times)
      (insert "12345678901234567890123456789012345678901234567890")
      (insert "12345678901234567890123456789012345678901234567890"))))

(dotimes (i 200)
  (with-timer (concat "Insert " (number-to-string i) "KB.")
    (test-insert-100 (* 10 i))))

````

gives the following output:

````
Insert 19KB.... done (0.00100s)
...
Insert 199KB.... done (0.00350s)
````

I'll have to agree this is probably a non-issue.

aaronjensen:
This is great, thanks! On my faster machine this definitely fixed the issue. No stutter now. I can try it on my slower machine tomorrow.

aaronjensen:
fwiw, i'm using this on my machines now.. .this project isn't really workable w/o it. Is there a path to getting this merged in? Thanks again.

josteink:
I guess @ananthakumaran needs to clean up whatever he considers unfinished about this code (since it's tagged WIP). 

But all in all, I guess we now know what fundamental part of the code is slowing us down and that it's easy to fix.

ananthakumaran:
Will get merged soon, I intend to clean up the code and add a custom-variable to control the threshold

aaronjensen:
Thanks @ananthakumaran!

Issue #190, opened 2017-08-31
Look for a typescript package and use tsserver from that package; keep the stock tsserver as a fallback


lddubeau:
It would be nice if `tide` could look for a locally installed `typescript` package and use the `tsserver` from that package instead of using the `tsserver` installed with `tide`. Perhaps the latter could be used as a fallback, without being the default.

The current behavior of `tide` with regards to selecting a `tsserver` has been a significant problem for me every since I started using `tide`. AFAIK, it's never been the case that the `tsserver` shipped with `tide` and the `typescript` version I used to compile any of my projects were at the same version number. I got a lolwut moment early on when I saw that the errors reported in Emacs and outside Emacs were different until I realized the problem was the `tsserver` used by `tide`. (And I suspect this is going to continue to bite users of `tide` time and again.)

So I quickly decided to override the default behavior with the following code. I'm not putting this as a PR because it is probably too idiosyncratic, but it can serve as a start for discussion at least. This is part of my `setup-tide-mode` function:

```
      ;; If there is a directory with "package.json" in it, we take that
      ;; as the location from which to search for a tsserver and set
      ;; tide-tsserver-executable to it, if it exists.
      (let* ((package-root (locate-dominating-file default-directory
                                                   "package.json"))
             (path
              (and package-root
                   (expand-file-name "node_modules/typescript/bin/tsserver"
                                     (expand-file-name package-root)))))
        (when (and path
                   (file-exists-p path))
          (make-local-variable 'tide-tsserver-executable)
          (setq tide-tsserver-executable path)
        ))
```

I've used this code for months now. Never had a problem.

leo-shopify:
This looks much more sustainable.
Checking for a global installation after `node-modules` would make sense. `which tsserver` (and the windows equivalent)?

Would it still make sense to embed a static version?
It's a good out of the box experience but at the same time anyone using typescript is likely capable of installing it independently :)

A function to log version and configuration could be useful for reporting issues, but this should provably a separate discussion.

ananthakumaran:
Tide used to ship modified version tsserver.js in the start, which is probably the reason for prefering it. Another good property of this setup is, the chance of encountering problems because of protocol mismatch between tide and tsserver is very less. But nowadays we try to keep tide backward compatibility with any 2.x version.

That said, we should probably try to give more preference to node_modules/ in the project root.

1) if tide-tsserver-executable is set, just use it.
[...]

PR #193, opened 2017-09-04
Define a search criteria for the tsserver executable.

leo-shopify:
This is a follow up to #190. It specifies a search order and calls `npm bin` to find `tsserver`.

The search order is:
1. User defined through `tide-tsserver-executable`.
2. Local npm package.
3. Global npm.
4. Bundled version.

⚠️ The code is not tested.

- [X] implement search for `(projectile-project-root)`
- [ ] tests

### DIFF
--- a/tide.el
+++ b/tide.el
@@ -464,7 +464,7 @@ Return a string representing the path or nil."
 (defun tide--project-package-bin ()
   "Return the package's node_module bin directory using projectile's project root or nil."
   (when (and (functionp 'projectile-project-p) (projectile-project-p))
-    (concat (projectile-project-p) "node_modules/typescript/lib")))
+    (concat (projectile-project-root) "node_modules/typescript/lib")))
 
 (defun tide--tsserver-file-name ()
   "Return a string representing the full path to the typescript server, including the system's extension, or nil."
--- a/tide.el
+++ b/tide.el
@@ -446,7 +446,7 @@ LINE is one based, OFFSET is one based and column is zero based"
       (kill-buffer (process-buffer process)))
     (tide-cleanup-project project-name)))
 
-(defun tide--npm-bin (&optional globalp)
+(defun tide--typescript-npm-package (&optional globalp)
   "Return the path to the npm bin folder.
 Use the global package path if GLOBALP is not nil, or the local path otherwise.
 Return a string representing the path or nil."
@@ -457,13 +457,14 @@ Return a string representing the path or nil."
     (with-temp-buffer
       (when (= no-errors
                (call-process "npm" no-input (current-buffer) no-redisplay
-                             "bin" global-switch))
-        (string-trim-right (buffer-string))))))
+                             "list" "typescript"
+                             "--depth 0" "--parseable" global-switch))
+        (concat (string-trim-right (buffer-string)) "/lib")))))
 
 (defun tide--project-package-bin ()
   "Return the package's node_module bin directory using projectile's project root or nil."
   (when (functionp 'projectile-project-root)
-    (concat (projectile-project-root) "node_modules/.bin/")))
+    (concat (projectile-project-root) "node_modules/typescript/lib")))
 
 (defun tide--tsserver-file-name ()
   "Return a string representing the full path to the typescript server, including the system's extension, or nil."
@@ -475,15 +476,18 @@ Return a string representing the path or nil."
   (let ((exe (expand-file-name file-name path)))
     (when (file-exists-p exe) exe)))
 
+(defconst tide--tsserver "tsserver.js"
+  "File-name of the file that node executes to start the typescript server.")
+
 (defun tide-locate-tsserver-executable ()
   "Locate the typescript server executable.
 If TIDE-TSSERVER-EXECUTABLE is set by the user use it.  Otherwise check in the npm local package directory, in the project root as defined by projectile, and in the npm global installation.  If nothing is found use the bundled version."
   (or
    (and tide-tsserver-executable (expand-file-name tide-tsserver-executable))
-   (tide--locate (tide--tsserver-file-name) (tide--npm-bin))
-   (tide--locate (tide--tsserver-file-name) (tide--project-package-bin))
-   (tide--locate (tide--tsserver-file-name) (tide--npm-bin 'global))
-   (expand-file-name "tsserver.js" tide-tsserver-directory)))
+   (tide--locate tide--tsserver (tide--typescript-npm-package))
+   (tide--locate tide--tsserver (tide--project-package-bin))
+   (tide--locate tide--tsserver (tide--typescript-npm-package 'global))
+   (expand-file-name tide--tsserver tide-tsserver-directory)))
 
 (defun tide-start-server ()
   (when (tide-current-server)
--- a/tide.el
+++ b/tide.el
@@ -460,6 +460,11 @@ Return a string representing the path or nil."
                              "bin" global-switch))
         (string-trim-right (buffer-string))))))
 
+(defun tide--project-package-bin ()
+  "Return the package's node_module bin directory using projectile's project root or nil."
+  (when (functionp 'projectile-project-root)
+    (concat (projectile-project-root) "node_modules/.bin/")))
+
 (defun tide--tsserver-file-name ()
   "Return a string representing the full path to the typescript server, including the system's extension, or nil."
   (let ((extension (if (eq system-type 'windows-nt) ".cmd" "")))
@@ -472,10 +477,11 @@ Return a string representing the path or nil."
 
 (defun tide-locate-tsserver-executable ()
   "Locate the typescript server executable.
-If TIDE-TSSERVER-EXECUTABLE is set by the user use it.  Otherwise check npm package local first and npm global installation after.  If nothing is found use the bundled version."
+If TIDE-TSSERVER-EXECUTABLE is set by the user use it.  Otherwise check in the npm local package directory, in the project root as defined by projectile, and in the npm global installation.  If nothing is found use the bundled version."
   (or
    (and tide-tsserver-executable (expand-file-name tide-tsserver-executable))
    (tide--locate (tide--tsserver-file-name) (tide--npm-bin))
+   (tide--locate (tide--tsserver-file-name) (tide--project-package-bin))
    (tide--locate (tide--tsserver-file-name) (tide--npm-bin 'global))
    (expand-file-name "tsserver.js" tide-tsserver-directory)))
 
--- a/tide.el
+++ b/tide.el
@@ -446,13 +446,28 @@ LINE is one based, OFFSET is one based and column is zero based"
       (kill-buffer (process-buffer process)))
     (tide-cleanup-project project-name)))
 
-(defun tide--npm-bin (&optional global-p)
-  "Return the npm bin path.  Relative to the default directory or the global path if GLOBAL-P is not nil."
-  (let ((exe (expand-file-name
-              "tsserver"
-              (with-temp-buffer
-                (when (= 0 (call-process "npm" nil (current-buffer) nil "bin" (if global-p "--global" "")))
-                  (string-trim-right (buffer-string)))))))
+(defun tide--npm-bin (&optional globalp)
+  "Return the path to the npm bin folder.
+Use the global package path if GLOBALP is not nil, or the local path otherwise.
+Return a string representing the path or nil."
+  (let ((no-errors 0)
+        (no-input nil)
+        (no-redisplay nil)
+        (global-switch (if globalp "--global" "")))
+    (with-temp-buffer
+      (when (= no-errors
+               (call-process "npm" no-input (current-buffer) no-redisplay
+                             "bin" global-switch))
+        (string-trim-right (buffer-string))))))
+
+(defun tide--tsserver-file-name ()
+  "Return a string representing the full path to the typescript server, including the system's extension, or nil."
+  (let ((extension (if (eq system-type 'windows-nt) ".cmd" "")))
+    (concat "tsserver" extension)))
+
+(defun tide--locate (file-name path)
+  "Locate the FILE-NAME in PATH.  Return a string representing the full path or nil."
+  (let ((exe (expand-file-name file-name path)))
     (when (file-exists-p exe) exe)))
 
 (defun tide-locate-tsserver-executable ()
@@ -460,8 +475,8 @@ LINE is one based, OFFSET is one based and column is zero based"
 If TIDE-TSSERVER-EXECUTABLE is set by the user use it.  Otherwise check npm package local first and npm global installation after.  If nothing is found use the bundled version."
   (or
    (and tide-tsserver-executable (expand-file-name tide-tsserver-executable))
-   (tide--npm-bin)
-   (tide--npm-bin 'global)
+   (tide--locate (tide--tsserver-file-name) (tide--npm-bin))
+   (tide--locate (tide--tsserver-file-name) (tide--npm-bin 'global))
    (expand-file-name "tsserver.js" tide-tsserver-directory)))
 
 (defun tide-start-server ()
--- a/tide.el
+++ b/tide.el
@@ -449,7 +449,8 @@ LINE is one based, OFFSET is one based and column is zero based"
 (defun tide--typescript-npm-package (&optional globalp)
   "Return the path to the npm bin folder.
 Use the global package path if GLOBALP is not nil, or the local path otherwise.
-Return a string representing the path or nil."
+Return a string representing the path or nil.
+If there are multiple results throw an error."
   (let ((no-errors 0)
         (no-input nil)
         (no-redisplay nil)
@@ -459,7 +460,17 @@ Return a string representing the path or nil."
                (call-process "npm" no-input (current-buffer) no-redisplay
                              "list" "typescript"
                              "--depth 0" "--parseable" global-switch))
-        (concat (string-trim-right (buffer-string)) "/lib")))))
+        ;; when the process returns without errors,
+        ;; split the buffer into LINES
+        (let ((lines (split-string (buffer-string) "\r?\n" 'omit-empty-lines)))
+          ;; if there are 0 lines return nil
+          ;; if there is only 1 line return it with /lib appended
+          (if (= 1 (length lines))
+              (concat (string-trim (car lines)) "/lib")
+            ;; if there are multiple results signal an error
+            (when (> 1 (length lines))
+              (error "Enexpected multiple tsserver paths:\n%s"
+                     (string-join lines "\n")))))))))
 
 (defun tide--project-package-bin ()
   "Return the package's node_module bin directory using projectile's project root or nil."
--- a/tide.el
+++ b/tide.el
@@ -463,8 +463,8 @@ Return a string representing the path or nil."
 
 (defun tide--project-package-bin ()
   "Return the package's node_module bin directory using projectile's project root or nil."
-  (when (functionp 'projectile-project-root)
-    (concat (projectile-project-root) "node_modules/typescript/lib")))
+  (when (and (functionp 'projectile-project-p) (projectile-project-p))
+    (concat (projectile-project-p) "node_modules/typescript/lib")))
 
 (defun tide--tsserver-file-name ()
   "Return a string representing the full path to the typescript server, including the system's extension, or nil."
--- a/tide.el
+++ b/tide.el
@@ -446,6 +446,24 @@ LINE is one based, OFFSET is one based and column is zero based"
       (kill-buffer (process-buffer process)))
     (tide-cleanup-project project-name)))
 
+(defun tide--npm-bin (&optional global-p)
+  "Return the npm bin path.  Relative to the default directory or the global path if GLOBAL-P is not nil."
+  (let ((exe (expand-file-name
+              "tsserver"
+              (with-temp-buffer
+                (when (= 0 (call-process "npm" nil (current-buffer) nil "bin" (if global-p "--global" "")))
+                  (string-trim-right (buffer-string)))))))
+    (when (file-exists-p exe) exe)))
+
+(defun tide-locate-tsserver-executable ()
+  "Locate the typescript server executable.
+If TIDE-TSSERVER-EXECUTABLE is set by the user use it.  Otherwise check npm package local first and npm global installation after.  If nothing is found use the bundled version."
+  (or
+   (and tide-tsserver-executable (expand-file-name tide-tsserver-executable))
+   (tide--npm-bin)
+   (tide--npm-bin 'global)
+   (expand-file-name "tsserver.js" tide-tsserver-directory)))
+
 (defun tide-start-server ()
   (when (tide-current-server)
     (error "Server already exist"))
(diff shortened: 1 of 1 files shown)


tide.el
```diff
@@ -446,6 +446,24 @@ LINE is one based, OFFSET is one based and column is zero based"
...
     (tide-cleanup-project project-name)))
 
+(defun tide--npm-bin (&optional global-p)
+  "Return the npm bin path.  Relative to the default directory or the global path if GLOBAL-P is not nil."
+  (let ((exe (expand-file-name
+              "tsserver"
+              (with-temp-buffer
+                (when (= 0 (call-process "npm" nil (current-buffer) nil "bin" (if global-p "--global" "")))
+                  (string-trim-right (buffer-string)))))))
+    (when (file-exists-p exe) exe)))
```
josteink: I know some people prefer compact one-liners, but I really like splitting things into smaller, clearly named functions. In my opinion that better communicates semantics and intent.

In that regard, this function could benefit from some de-clevering/de-compacting of the expression in lines 453 to 455 here.

I literally had to run it through edebug to ensure I understood what it does and how it works, and that's not really the best of signs :)

josteink:
Hey leo!

Thanks for the PR. Apart from the one comment on one instance of code clarity, I think this all in all looks good.

Once cleared up, I'll be happy to get this merged.

leo-shopify:
@josteink Thanks for the feedback. To be honest I'm not sure it's a good idea to merge this or not.
The two options I see are:
1. Choose a version of typescript automatically and let the user override it. This PR.
2. The reverse: Give the user the tools to configure it, making no assumptions on their behalf.

I can, of course make the adjustments (although I don't have a windows box to test) if you thinks this is the way to go.

Thoughts?

asmundg:
A use case that I'm constantly running into, is finding tsserver in a monorepo (see https://github.com/lerna/lerna) architecture, where there is an outer node_modules at the git root and several inner node_modules in various package directories. Typescript tends to get installed into the outer node_modules.

tide-project-root will happily use the directory containing the inner node_modules as default-directory, failing to find tsserver.

So a search step that tries to resolve tide-tsserver-executable relative to the git root, as well as the tsconfig root, could be valuable.

leo-shopify:
Allowing the user to set `tide-tsserver-executable` to any arbitrary path is the way to handle any use-case outside of the predefined one.
I'm not sure though that *local npm*, *global npm*, *bundled* is common enough.
I suggested leaving things as they are now, assuming the user will configure it to their liking, adding more documentation instead of code.
I'm happy to help with either choice, but would be interested in people's opinion, especially the maintainer.

ananthakumaran:
as @leo-shopify mentioned, we can't make all the cases work out of the box. For monorepo, perhaps a `.dir-locals.el` in the root folder might work out.

I am not much happy about falling back to global tsserver, because it's going to break my local setup. But I leave it up to others to decide.

josteink:
I'm happy with the current approach, as long as we can get it to work similarly on Windows too (and I'll be glad to help out with that).

> I am not much happy about falling back to global tsserver, because it's going to break my local setup.

Some people install typescript as a global toolchain (especially on Windows where it doesn't require admin/sudo), so I think including that in the fallback strategy makes sense.

Either way, it will only be used if the project/package has no local toolchain, and that's probably not very common.

@asmundg: If I understand you correctly, you are saying that in a lerna-style repo, you have modules and inner-modules, and when editing the inner-modules npm will happily report the path to the package-local tsserver, but you'd rather use the outer-module's tsserver?

May I ask why? The code will be parsed and compiled by the inner-modules tsc and tsserver in the end, right?

As @ananthakumaran says, I guess a `.dir-locals.el` could work here...

> So a search step that tries to resolve tide-tsserver-executable relative to the git root, as well as the tsconfig root, could be valuable.

This could work for some setups, but other mixed-mode mono-repos has a git-root where the node-package is a subfolder somewhere down the tree with no direct `packages.json` at git-root-level. How should we support that? Dig *up* from git-root?

Also: How would you suggest it should be prioritized then?

I'm not against the idea per se, but I think we need to think a little bit about how to make it work best for most people, and hopefully without any needless complication and sources of confusion.

asmundg:
@josteink what generally happens, is that shared dependencies, typically including typescript, get hoisted to the outer node_modules (https://github.com/lerna/lerna/blob/master/doc/hoist.md).

In this case, there is no tsserver in the inner node_modules, and tide fails to locate and start the outer one. Specifying ../../node_modules/typescript etc. as the tsserver executable in dir locals would probably work fine, though.

josteink:
Ok. Gotcha.

I think in that case I would like to propose 2 changes:

There's no `git-root`. Not everyone uses git, and this needlessly ties tide's implementation to a specific VCS. Let's instead piggyback on `(projectile-project-root)`, if present, which has this problem solved in a more general sense.

With that said, a more appropriate search order ... is in order:

* user-configured value (should always be used if present)
* local NPM packages
* projectile-root packages
* global NPM packages
* use tide-supplied tsserver

I think with this we should be able to support quote a lot of setups:

* "standard"
* LERNA-based monorepo setups
* even my somewhat "obscure" mixed-mode mono-repo, since tsserver should be resolved by local NPM search

Anything we're missing? Does this approach make everyone happy, or do we still have cases not properly handled or considered?

MarcusRiemer:
This seems fine to me, I especially like relying on projectile! I only have a question about the implementation: Would it be possible to make the order configurable like `Projectile Project Root Files Functions` (in the `projectile`-group of `customize`). I agree that this seems to be the most sane default but would love to have a possibility to trivially override the behaviour as a nice extra.

josteink:
> I only have a question about the implementation: Would it be possible to make the order configurable 

For me this is really not very important, but if enough people wants it and feel like putting in manpower to making it happen, I won't be opposed to it either.

For now, I would like to land this PR as a general enhancement, without further complicating matters.

Once landed, feel free to issue (or root for!) another PR which can then improve upon what we end up with here. Does that sound acceptable?

MarcusRiemer:
> Does that sound acceptable?

Fully! Sorry if that wasn't clear from my answer. Relying on `projectile-root packages` makes this change work in all of my projects.

josteink:
@leo-shopify... That leaves us 2 simple things to do:

* implement search for `(projectile-project-root)`
* and Windows-testing

Once you've amended your PR to include the first, I'll be around to assist with the second. Sounds good?

leo-shopify:
@josteink Sounds good. I'll be looking at this this weekend.

josteink:
@leo-shopify Any news on this one?

Not meaning to nag on you or anything, but this genuinely sounds like an improvement, and I'm interested in knowing if someone is still working on this or not :)

leo-shopify:
@josteink Sorry, I haven't had the time. If for some reason I can't make progress this weekend I'll let you know, but my intention is to work on it.

josteink:
No rush. Just checking that it's not an abandoned effort :)

leo-shopify:
@josteink I pushed an update with code cleanups and adding the `.cmd` extension on windows.
`(projectile-project-root)` is not done yet.

josteink:
Sounds good. Will do.

josteink:
My first test here fails, but I cannot immediately blame your code either because this is a new Windows-setup, where I haven't really used Emacs before.

Basically it fails to start tsserver properly, because it exits immediately with error-code 1. But it also does so when I disable probing for npm-locations at all... So not sure what the error is.

I'll try to debug this properly when I have more time.

leo-shopify:
Maybe `tide--npm-bin` is failing to find `npm` in the path?
Is the bundled `tsserver.js` being found?

josteink:
That’s not it. Resolving tsserver through npm works fine. 

Something bad happens *after* that. I’ll look into it.

josteink:
I think I've found the error.

Basically you try to resolve the `tssserver` executable. (That is either the *nix-script which with a shebang, or the windows `tsserver.cmd`-script. Both of these scripts then launch `tsserver.js` through `node` when invoked.

You successfully find the location of this script according to the spec given.... And store it in a variable called `tsserverjs` which the rest of the code assumes is a node-loadable JS-file. Which it ofcourse isn't.

You either need to ensure that:

1. `tide-locate-tsserver-executable` always returns a executable and never a JS-file (now it returns JS when it falls back to `tide`'s bundled `tsserver`). This executable needs to be launched directly and not through `node`.
2. `tide-locate-tsserver-executable` always returns a JS-file and never a executable (because now it is executable in almost all outcomes). This JS-file must always be launched through `node`.

Personally I would prefer the latter version, since that leaves less platform-specific issues in our code (like having to know about `.cmd` on Windows).

Sounds like something you can handle? :)

leo-shopify:
Good catch! Scripts executable by node vs. the shell.

I also prefer option 2: *always returns a JS-file*, on the basis that the distinction was introduced by my code. It is a localized change rather than having to review the whole code-base.

I believe we should still be able to delegate the `node_modules` search to NPM instead of re-implementing it in elisp by running:

- `npm list typescript --depth 0 --parseable` for local. 
    On my machine:
    > /Users/leo-shopify/src/some-project/node_modules/typescript
- `npm list typescript --depth 0 --parseable --global` for global. 
    On my machine:
    > /usr/local/lib/node_modules/typescript

In both cases I 'd have to append `/lib/tsserver.js`, regardless of the OS.

This approach does have a caveat:
If the user is using [nvm](https://github.com/creationix/nvm) or similar, running `npm list typescript --depth 0 --parseable --global` does **not** search the global package directory but whatever nvm defines as global. e.g.:
> /Users/leo-shopify/.nvm/versions/node/v8.1.4/lib

Any suggestions?

josteink:
I don’t use nvm and don’t know how it works, nor how common a system it is. 

In the cases where nvm screws things up, won’t the projectile search still return package-local results though?

If so, I’d be OK with that. 

Anyone else have an opinion?

lddubeau:
@leo-shopify I don't understand this part.

> If the user is using [nvm](https://github.com/creationix/nvm) or similar, running `npm list typescript --depth 0 --parseable --global` does **not** search the global package directory but whatever nvm defines as global. e.g.:

All Node installations can be configured to put global packages wherever the user wants. There's a default built-in, but the default does not have to be used. Case in point, on my development machine I've set Node to put global packages under a subdirectory under my home. This way I dodge the whole permission issue. (I don't have to do `sudo npm install -g foo`.) I've pretty much never used the default that Node provides.

I'm not familiar with `nvm` but AFAIK `nvm` just create new Node installations as needed, and sets the global package directory for the new installation to a location that won't interfere with other installations. This, however, is not special, as all Node installations can have custom locations for their global packages. When the user creates a new Node installation `N` with nvm, then as far as all code executed by `N` is concerned, the global package directory is exactly "what nvm defines as global". So, yes, if you execute `npm list typescript --depth 0 --parseable --global` with `N` and the path returned is inside the directory that nvm defines as global for `N`, then the result is completely correct. If you are looking for another result, you're asking Node to work differently than it does.

Again I don't use `nvm`. If I'm off please correct me.

----

nvm does create complications. A user could be editing files that are part of projects that should be compiled under different Node installations created by nvm (or the equivalent). I have no idea whether it is possible to detect in Elisp what Node installation should be used by which project. I suspect the answer is that there's no general solution.

If there is no general solution, then tide should only do a local and global search and let the chips fall where they may. If someone has multiple Node installations they need to handle, they can write custom code to deal with such cases.

leo-shopify:
@lddubeau Yes, you are right, it is not an issue. The behaviour of npm under nvm is correct.
@josteink `npm list` works correctly for locally installed packages (and global, as @lddubeau pointed out). I got confused since my system version of node and the one under nvm were the same and I expected the packages to visible to both. Which of course would defeat the purpose of having isolated environments.

tide.el
```diff
@@ -446,6 +446,49 @@ LINE is one based, OFFSET is one based and column is zero based"
...
+               (call-process "npm" no-input (current-buffer) no-redisplay
+                             "list" "typescript"
+                             "--depth 0" "--parseable" global-switch))
+        (concat (string-trim-right (buffer-string)) "/lib")))))
+
+(defun tide--project-package-bin ()
+  "Return the package's node_module bin directory using projectile's project root or nil."
+  (when (functionp 'projectile-project-root)
+    (concat (projectile-project-root) "node_modules/typescript/lib")))
+
```
josteink: This one seems to error out when not in a project. You might want to wrap this in a `(ignore-errors)` block to that error, or even better avoid the error entirely by checking `(projectile-project-p)`.

tide.el
```diff
@@ -446,6 +446,49 @@ LINE is one based, OFFSET is one based and column is zero based"
...
+(defun tide-locate-tsserver-executable ()
+  "Locate the typescript server executable.
+If TIDE-TSSERVER-EXECUTABLE is set by the user use it.  Otherwise check in the npm local package directory, in the project root as defined by projectile, and in the npm global installation.  If nothing is found use the bundled version."
+  (or
+   (and tide-tsserver-executable (expand-file-name tide-tsserver-executable))
+   (tide--locate tide--tsserver (tide--typescript-npm-package))
+   (tide--locate tide--tsserver (tide--project-package-bin))
+   (tide--locate tide--tsserver (tide--typescript-npm-package 'global))
+   (expand-file-name tide--tsserver tide-tsserver-directory)))
```
josteink: Running this one on my system produces the following odd output:

````elisp
"/usr/lib/node_modules/dts-gen/node_modules/typescript
/usr/lib/node_modules/tsun/node_modules/typescript
/usr/lib/node_modules/typescript
/usr/lib/node_modules/typings/node_modules/typescript/lib"
````

The correct folder for my global typescript-installation should be `/usr/lib/node_modules/typescript`, so I'm not sure what's going on here, and how to best handle this without risk breaking custom global paths, like those @lddubeau has.

Perhaps we can check and verify it's not located inside a nested `node_modules` folder?

**Comment by josteink** [comment]
IMO the code itself looks good to. I have no qualms about the style, or choice of solution.

That said, from my testing so far (and note that for the last commit, this testing was Linux-only), we're not entirely ready to land this yet, but we're too far away either ;)

Good job so far!

File: tide.el (hunk 3 above)
Comment (leo-shopify) [reply tide.el]: What do you get when you run `npm list --depth 0 --parseable --global` in a shell prompt?

**Comment by leo-shopify** [comment]
@josteink No worries, thanks for testing. As I mentioned in the PR comment there are some potential issues with `npm list`, like de-duplication, links and multiple versions of a package which make the pr experimental until we have a clear understanding of the implications and how to solve them.

File: tide.el (hunk 3 above)
Comment (josteink) [reply tide.el]:
````sh
➜  ~ npm list typescript --depth 0 --parseable --global  


/usr/lib/node_modules/typescript

➜  ~ 
````

Worth noting: The multiple instances of new-lines are not copy-paste errors.

Comment (leo-shopify) [reply tide.el]: Thanks! That's what I expected and what I get locally.
It seems that your tests might not be evaluating the latest definitions, `tsserver.js` is missing from the end.

Would you mind:
1. Restarting emacs
2. Evaluate region (lines 449 to 490):
https://github.com/leo-shopify/tide/blob/ff032e5b12aa1989e0c69deeef3ecd7458c979d5/tide.el#L449-L490
3. In the scratch buffer, running each line separately:
```elisp
   (tide--locate tide--tsserver (tide--typescript-npm-package))
   (tide--locate tide--tsserver (tide--project-package-bin))
   (tide--locate tide--tsserver (tide--typescript-npm-package 'global))
```

Right now running `tide--typescript-npm-package` can return multiple lines. I'm trimming white-space, so it works fine if there is a single result, but if there are multiple, like in your original example, we have to choose one. I haven't figured out yet which are all the possible cases where `npm list` would return multiple results and how to handle them.
Understanding the possibilities will tell us if this approach is viable. If it is then each case should become a test. If it is not viable we can go back to `npm bin` or use a different approach.
What do you think?

leo-shopify:
I cleaned up the code so that more people can test it.
Now if an `npm list` search returns multiple paths (both global and local) a warning is logged listing them.
The first path is always picked.

If anyone has time I'd be interested to know:
-  Cases where multiple paths are found.
- What criteria should be used to pick one path from the list.

Thanks.

josteink:
The latest code-updates looks really nice, and I'll try to get it tested as soon as possible.

While we *did* get some crucial aspects of how the implementation should be ironed out before anyone got hurt...

This PR has definitely been hanging around longer than I prefer them too. For the sake of everyone involved, I'd like to get it closed sooner rather than later.

Thanks for being patient!

josteink:
Impressions so far (testing on Windows)...

Most things seems to work as expected. We're clearly almost there.

While we don't have much windows-specific code we apply `/` as a directory-separator on top of whatever npm/projectile returns. In my testing this seems to be fine though:

````elisp
(file-exists-p "d:\\Git\\TrunkGmail\\BrowserExtension\\node_modules\\typescript/lib/tsserver.js")
t
````

So cross-platform really should be handled, and it's done through Emacs core, which is good because it means less work for us.

What's more of an issue is npm... At least on Windows. I'm seeing 2 big problems:

**Performance:** npm is terribly, terribly slow. It can churn away up to 10 seconds on a tripple-SSD setup with 32GBs of RAM and Xeon CPU for one single search. Repeat this for local and global and opening as TS-file takes almost 30 seconds with Emacs completely locked in the mean time.

(This may be less of an issue on Linux, but I still think it's worth highlighting)

This isn't our faults per sé, but it makes it harder to depend on this as something we need to run when opening each and every new buffer.

I was aware that performance might become an issue we needed to tackle somehow, but I was hoping it could be tackled independently of these changes.

Right now though, we really can't merge this to *master* with these performance-issues unaddressed.

Please note: This does not mean I'm *blocking* the PR and putting the burden of addressing this solely on you.

Instead I propose that once everything else is settled, we merge this into an intermediate branch, and that collaborate on a performance fix on top of your patch there. I already have some pretty clear ideas of how we can get this done, but I'll save that discussion for later.

**Consistency:** the output from npm on the command-line is inconsistent with what I experience in Emacs. Which makes no sense.

The output I receive seems to indicate that the global search does *not* use the `--depth  0` parameter but the code clearly says it should...

If I issue the following command from the command line, I get this:

````
D:\Git\tide\example> npm list typescript --depth 0 --parseable --global

C:\Users\josteink\AppData\Roaming\npm\node_modules\typescript
````

When opening `test.ts` (from examples folder above) from the tide-project and eval `(tide--npm-package 'global)`, I get the following output:

````
Warning (tide): Multiple global typescript paths:
C:\Users\josteink\AppData\Roaming\npm\node_modules\tsun\node_modules\typescript
C:\Users\josteink\AppData\Roaming\npm\node_modules\typescript
C:\Users\josteink\AppData\Roaming\npm\node_modules\typings\node_modules\typescript
Using: C:\Users\josteink\AppData\Roaming\npm\node_modules\tsun\node_modules\typescript
````

How can that be? That's clearly not depth-limited.... Anyone idea what's going on here?

If we can get that sorted out, I'll be happy to move this PR into the next stage (that is, performance fixing).

@asmundg , @MarcusRiemer : Could you please check out the latest changes and see if you're happy about the state of things right now? Are all your use-cases well handled?

Now's the time to let us know :)

lddubeau:
> **Performance:** npm is terribly, terribly slow. It can churn away up to 10 seconds on a tripple-SSD setup with 32GBs of RAM and Xeon CPU for one single search. Repeat this for local and global and opening as TS-file takes almost 30 seconds with Emacs completely locked in the mean time.

Oh wow. That's not viable. I tried the code a bit when `npm bin` was still the method used to find `tsserver`. On my development machine, running `npm bin` took in the neighborhood of 1 second. I use `desktop` and have hundreds of buffers open in Emacs. At 1 second per TypeScript buffer, `npm bin` was too slow for my taste.

I've just checked with `npm list`. That's 5 seconds on my machine. :(

The figures above are on the latest Debian stable, by the way.

I almost made a comment when I did my testing with `npm bin`. I guess I'm going to make it now. Two things:

1. If we are going to search in Node's global package directory, then we *have* to query Node, because as I mentioned in an earlier comment, the global package directory is configurable, so we cannot assume anything. However, it would be possible to do this search *once* per Emacs session and cache the result instead of repeating the search every time we need to find a global `tsserver`. (If we are to end up with support for multiple Node environments created through `nvm`, then the search could be done once per environment, which would still be generally better than once per buffer.)

   In theory the location of the global package directory could change while Emacs is running. It would be legitimate IMHO to require the user to flush the cache manually as needed. (This is just like sometimes `tsserver` needs to be manually restarted due to changes outside Emacs, like upgrading TypeScript to a new version.)

2. As far as local packages go, I expect that once a suitable project root is located, the number of possibilities as to the location of the local `tsserver` is limited. I'd expect one possibility for *nix systems, and one for Windows, and that's that. For instance, to my knowledge, Node does not have a configuration option to name `node_modules` differently. So it would be possible, once the project root is located, to compute the location of the local `tsserver` in Elisp code instead of calling upon Node. (Of course, if there *is* a way to configure Node that would make it impossible to know where a local `tsserver` is installed, then this suggestion goes out the window.)

I do realize it may mean that some code layouts won't be supported out of the box. However, if the choice is between:

A. Supporting common simple layouts at decent speed, and

B. Supporting less common layouts but a the cost of significantly slowing down initialization for *everyone*, then

I'd take A over B every time.

josteink:
> Oh wow. That's not viable.

Two things:

* I've now done some testing on my Linux laptop and the performance is noticeably better than on Windows. It's still not as instant as it used to be though.
* I've also misrepresented how tide does things slightly in a manner which may turn people's people's judgement about these changes in a unfair direction.

**The big error in my my previous comment was that I claimed this delay would be incurred on each file opened.** The truth is that `tide` only starts one `tsserver` per project.

Even with that in place, like @lddubeau says, it takes a second or two to open the first, initial typescript-buffer.

> However, if the choice is between:
>
> A. Supporting common simple layouts at decent speed, and
>
> B. Supporting less common layouts but a the cost of significantly slowing down initialization for everyone, then
>
> I'd take A over B every time.

Just for clarity, I absolutely agree about this one. My hope was that we could support B, without the significant slowdown.

What I had in mind, was trying to create a minimally complex caching layer.

My idea was to wrap all calls to `tide--npm-package` in a intermediate caching function, taking `(file-name-directory buffer-file-name)` as a cache-key for local searches and `'global` as cache-key for global searches.

But we're still going to incur that performance cost at least for one buffer... If that's unacceptable... We'll have to actually consider alternative approaches.

In which case, sorry @leo-shopify. Sometimes you have to be the one who lands all the hard work, just to prove that it's not the right solution :( 

So with that said: if we are to assume a different approach, I do have some clear ideas and concepts, but I'd rather start a new issue for discussing that separate from this PR. IMO a PR should ideally be discussion about code in a mostly agreed upon specification/change, without too much noise. And we've clearly crossed that threshold in this case.

So those in favour of starting a new issue to discuss alternative strategies before we commit/discard this... Let me know.

leo-shopify:
@josteink No worries, I saw this PR as experimental, I'm glad we were able to gain useful knowledge from it. 
Referring back to my [initial comment](https://github.com/ananthakumaran/tide/pull/193#issuecomment-328968931), I think ananthakumaran's original design is the right approach.
Finding the executable is a separate concern from tide's core function. All tide needs is to be told where the tsserver is. 
How tsserver is found, given the trade-offs and variety of use cases, I think would be best served by a separate module or api. A "one size fits all" approach will bloat tide with incidental concerns and very likely won't be able to satisfy everyone.
I think we should consider these boundaries in the new issue.

josteink:
In lieu of @lddubeau comment:

> However, if the choice is between:
>
> A. Supporting common simple layouts at decent speed, and
>
> B. Supporting less common layouts but a the cost of significantly slowing down initialization for everyone, then
>
> I'd take A over B every time.

I decided to investigate how work this would really be... And the fastest way to estimate coding time is obviously by doing :)

So I created this: https://github.com/ananthakumaran/tide/tree/lightweith-locate-tsserver-executable

And specifically, on top of everything @leo-shopify has done so far, I wrote this really simple stuff: https://github.com/ananthakumaran/tide/commit/503b17c6404214c5f53879eaf4984046f5c3f8ef (ironically I haven't tested the Windows-code yet :) )

Would anyone like to fool around with that, and see if that works better for them? I'm especially interested if you have anything against it, @lddubeau, seeing as you're one of the more serious contributors we have around here, and your opinion does carry weight. 

If we can all agree that this is "good enough" and works, I guess we don't need to spend too much time discussing other alternate approaches.

josteink:
Closed via https://github.com/ananthakumaran/tide/pull/201.

PR #195, opened 2017-09-16
ensure `buffer-file-name` is a string

leo-shopify:
`tide-command:openfile` assumes `buffer-file-name` is a string when evaluating its body.
This is not the case when exporting javascript org-mode code blocks to html, for example.
Checking that `buffer-file-name` is a string prevents evaluation errors and returning `nil` is enough for the org-mode export case to work.

This was the top of the stack trace before the change:
```elisp
Debugger entered--Lisp error: (wrong-type-argument stringp nil)
  file-name-nondirectory(nil)
  file-name-extension(nil)
  tide-command:openfile()
  tide-configure-buffer()
  tide-setup()
  setup-tide-mode()
  js2-mode()
  funcall(js2-mode)
  org-html-fontify-code("" "js")
```

### DIFF
--- a/tide.el
+++ b/tide.el
@@ -1458,15 +1458,21 @@ code-analysis."
 (defun tide-setup ()
   "Setup `tide-mode' in current buffer."
   (interactive)
-  (tide-start-server-if-required)
-  (tide-mode 1)
-  (set (make-local-variable 'eldoc-documentation-function)
-       'tide-eldoc-function)
-  (set (make-local-variable 'imenu-auto-rescan) t)
-  (set (make-local-variable 'imenu-create-index-function)
-       'tide-imenu-index)
 
-  (tide-configure-buffer))
+  ;; skip buffers where buffer-file-name is not defined, such as org-mode code
+  ;; blocks as they are fontified for html export
+  (if (stringp buffer-file-name)
+      (progn
+        (tide-start-server-if-required)
+        (tide-mode 1)
+        (set (make-local-variable 'eldoc-documentation-function)
+             'tide-eldoc-function)
+        (set (make-local-variable 'imenu-auto-rescan) t)
+        (set (make-local-variable 'imenu-create-index-function)
+             'tide-imenu-index)
+
+        (tide-configure-buffer))
+    (display-warning 'tide "A file backed buffer is required to start the tsserver.")))
 
 ;;;###autoload
 (define-minor-mode tide-mode


ananthakumaran:
Tide mode only works if the buffer is associated with a real file. The entire codebase makes an implicit assumption that buffer-file-name is non-nil and points to a file.

There are two ways we can go about buffers that are not backed by a file
1) handle it properly in tide
2) don't enable tide mode on these buffers.

No 2 is easy, you can just add another check in your setup-tide-mode config. If you want to fix tide, then the right approach is to throw an error with good message. The test should probably be done in`tide-setup`

josteink:
Throwing an error might cancel out other mode-hooks though, so I think the best approach would be to fail gracefully and don't enable tide, but without signaling an error.

ananthakumaran:
@josteink If throwing an error is a bit aggressive, then perhaps a `(warn *)` without enabling tide would do?

josteink:
I'd be fine with that.

leo-shopify:
I reverted the change and added the guard to the setup example in the readme instead. Realizing the org-mode export to html failure was related to tide was not obvious and it might help others.

leo-shopify:
How would you go about wrapping all expression that might fail with a warning?

josteink:
CI was just flaky. I re-ran the build now, its green and there's no issues with this PR at least.

josteink:
Superficially this looks good to me, although I haven't had the chance to test this yet.

Assuming there's nothing wrong with the code, IMO if we squash this, I think merging is OK.

leo-shopify:
If you have the latest stable version of *org-mode* (9.1 or newer) you can test the PR by:
1. Creating an org buffer with this content:
```org
#+begin_src js
  // test
#+end_src
```
2. `M-x` `org-html-export-as-html`
3. Verify that a `*Warnings*` buffer was created with:
```
Warning (tide): A file backed buffer is required to start the tsserver.
```
4. Verify that an `html` buffer was created. The content will vary depending on your configuration, but it should have somewhere in the body:
```html
<pre><code class="src src-js">// test
</code></pre>
```

leo-shopify:
Personally I find the warning a little noisy as in the org export case I wouldn't expect tide to have any involvement and there is nothing actionable to make it go away. 
Would an entry in the `*Messages*` buffer be reasonable?
Can you think of any cases where the warning would be a better option?

ananthakumaran:
ok, If you find warnings annoying, I am ok with `(message ..)` also. Can you squash the commits so that we can merge it?

josteink:
> Can you squash the commits so that we can merge it?

FYI: Github lets you do that too now when merging:

image

No need to put an extra burdon on this poor chap over that technicality ;)

ananthakumaran:
:), I know about that option, but not sure who will become the author of that commit, so bit reluctant to use it.

josteink:
If you squash it, it will credit the original contributor as the author, but will list you as *committing* that code.

leo-shopify:
I squashed the commits as they were. I didn't change the warning to a message in case you want to merge it now.

josteink:
Ok. That definitely looks strange. Ideally I'd like for all CI-builds to always run, even though I can tell the failure seems related to Emacs git-master.

I'll try to dig a little into it, and we'll get this PR going again.

josteink:
I've seen the same CI build-failure on other repos I manage, and it all seems constrained to a package which for some reason isn't available for Emacs git master.

I haven't quite decided what to do with this, short term or long term, and I apologise for the inconvenience that is causing.

@ananthakumaran: If we for now disable testing against Emacs git master we are losing compatibility test against future emerging releases. I think that's a bad thing, and would like to re-enable this as soon as possible. But that means we will have to remember to do so... Should we create a separate PR for that? That would get us a CI-build we can monitor for when things are back to normal.

ananthakumaran:
@josteink, instead of disabling, could we use allow_failures feature for master? https://docs.travis-ci.com/user/customizing-the-build#Rows-that-are-Allowed-to-Fail

josteink:
I'm not sure what that buys us, since that still leaves us failing PRs when there's technically nothing wrong with the PR itself. And as such, the CI build doesn't provide a useful metric for the correctness of the PR.

ananthakumaran:
> I'm not sure what that buys us, since that still leaves us failing PRs
allowed_failures mean build won't fail if that env failed. Also, we can go look at Travis for any build and see if it supports emacs master

josteink:
If you're happy about that, feel free to set it up. It sounds like you've read the tech-docs :)

PR #197, opened 2017-09-24
trigger rename command after refactor

ananthakumaran:
see https://github.com/Microsoft/TypeScript/pull/18351 for details. This is backward compatible if renameLocation is not present it won't do anything. To test it locally, you need to use `release-2.5` branch of `TypeScript`

### DIFF
--- a/tide.el
+++ b/tide.el
@@ -260,7 +260,8 @@ ones and overrule settings in the other lists."
 (defmacro tide-on-response-success (response ignore-empty-response &rest body)
   (declare (indent 2))
   `(if (tide-response-success-p ,response)
-       ,@body
+       (progn
+         ,@body)
      (-when-let (msg (plist-get response :message))
        (unless (and ,ignore-empty-response (string-equal msg "No content available."))
          (message "%s" msg)))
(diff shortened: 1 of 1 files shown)


josteink:
I haven't tested it yet, but looking at the code, and the TypeScript-issue that all seems to make sense to me. Looks like a good improvement.

I'll check it in more detail tomorrow.

ananthakumaran:
no need to hurry, take your time.

josteink:
Tested and works. Commit away!

One comment only, and that's really orthogonal to this PR...

The more I use `tide-refactor`, the more I feel it's a stepping-stone I'd rather not use.

In Visual Studio and other editors, I have direct keys for specific refactorings. Some examples: `Ctrl-R Ctrl-R` for symbol-rename, `Ctrl-R Ctrl-M` for method-extraction, etc.

I very often know what I want, and I'd like to just get *right to it* without asking my editor if it's actually an option, because I can tell up front it should be.

Should we add some extra `tide-` commands to apply known refactorings directly without having to use `tide-refactor` as a stepping-stone?

ananthakumaran:
> Should we add some extra tide- commands to apply known refactorings directly without having to use tide-refactor as a stepping-stone?

Yes, we could. It still going to ask about which scope you want to extract the method etc, but in many cases, the operation could be carried away without confirmation.

PR #201, opened 2017-10-07
Improved algorithm for finding a good tsserver

josteink:
This PR contains the changes proposed and implemented by @leo-shopify in #193, but includes additional patches, and as such supersedes it.

It further includes changes proposed in the discussion in #190.

Maybe I'm overly optimistic, but I'd like to think that the *specification* for how this should work is mostly settled after 2 looong threads in separate issues.

Therefore at this point, I hope review and comments should be able to focus on the code and implementation.

That said... Testing, review and comments is very much welcome :)

CC: @ananthakumaran @lddubeau @leo-shopify @asmundg @MarcusRiemer

### DIFF
--- a/tide.el
+++ b/tide.el
@@ -453,6 +453,75 @@ LINE is one based, OFFSET is one based and column is zero based"
       (kill-buffer (process-buffer process)))
     (tide-cleanup-project project-name)))
 
+(defun tide--typescript-npm-list (&optional globalp)
+  "Return a list of paths to the root of the typescript package as per npm list.
+Use the global package path if GLOBALP is not nil, or the local path otherwise.
+Return a list of strings representing each path or nil."
+  (with-temp-buffer
+    (let ((no-errors 0)
+          (no-input nil)
+          (no-redisplay nil)
+          (global-switch (if globalp "--global" "")))
+      (when (= no-errors
+               (call-process "npm" no-input (current-buffer) no-redisplay
+                             "list" "typescript"
+                             "--depth 0" "--parseable" global-switch))
+        ;; when the process returns without errors split the buffer into LINES
+        (split-string (buffer-string) "\r?\n" 'omit-empty-lines)))))
+
+(defun tide--multiple-paths-msg (context path-list used-path)
+  "Return a string with the typescript paths found either locally or globally.
+Use for logging purposes.
+CONTEXT is a string representing the context of the paths.
+PATH-LIST is the list of paths to the typescript package as strings.
+USED-PATH is a string representing the chosen path."
+  (concat "Multiple " context " typescript paths:\n"
+          (string-join path-list "\n")
+          "\nUsing: " used-path))
+
+(defun tide--pick-typescript-path (path-list &optional globalp)
+  "Return a single path to the typescript package directory or nil.
+PATH-LIST is the list of paths to the typescript package as strings.
+GLOBALP is an optional boolean representing whether PATH-LIST is the result of a global or local search."
+  (let ((path-list-length (length path-list)))
+    (when (> path-list-length 0)
+      (let ((chosen-path (string-trim (car path-list))))
+        (when (> path-list-length 1)
+          (display-warning
+           'tide (tide--multiple-paths-msg
+                  (if globalp "global" "local") path-list chosen-path)))
+        chosen-path))))
+
+(defun tide--npm-package (&optional globalp)
+  "Return a single path to the typescript package directory or nil.
+GLOBALP is an optional boolean indicating if the search is global or local."
+  (concat (tide--pick-typescript-path (tide--typescript-npm-list globalp) globalp)
+          "/lib"))
+
+(defun tide--project-package ()
+  "Return the package's node_module bin directory using projectile's project root or nil."
+  (when (and (functionp 'projectile-project-p) (projectile-project-p))
+    (concat (projectile-project-root) "node_modules/typescript/lib")))
+
+(defconst tide--tsserver "tsserver.js"
+  "File-name of the file that node executes to start the typescript server.")
+
+(defun tide--locate-tsserver (path)
+  "Locate the typescript server in PATH.
+Return a string representing the existing full path or nil."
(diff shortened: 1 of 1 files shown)


josteink:
I re-pushed (forced) the branch to re-introduce some commits I lost: Namely ensuring good Windows support, which accidentally got lost in a rebase/clean-up.

I hereby declare the PR testable on all supported platforms :)

ananthakumaran:
just came back from vacation. This approach looks lot better than calling out npm executable, will review soon

tide.el
```diff
@@ -453,6 +453,53 @@ LINE is one based, OFFSET is one based and column is zero based"
...
+  "Locate tsserver through project-local or global system settings."
+  (or
+   (tide--locate-tsserver (tide--npm-local))
+   (tide--locate-tsserver (tide--project-package))
+   (tide--locate-tsserver (tide--npm-global))))
+
+(defcustom tide-tsserver-locator-function #'tide-tsserver-locater-npmlocal-projectile-npmglobal
+  "Function used by tide to locate tsserver."
+  :type 'function
+  :group 'tide)
```
ananthakumaran: can we move this to the top alongside other defcustoms

tide.el
```diff
@@ -453,6 +453,53 @@ LINE is one based, OFFSET is one based and column is zero based"
       (kill-buffer (process-buffer process)))
     (tide-cleanup-project project-name)))
 
+(defun tide--npm-local ()
+  "Return a single path to the package-local typescript package directory or nil."
+
+  (when-let* ((packages-folder (locate-dominating-file default-directory "package.json")))
```
ananthakumaran: when-let would be sufficient I guess

ananthakumaran:
I have tested it locally (Mac)

* I had to change when-let* to -when-let to get it working. `tide--npm-local: Symbol’s function definition is void: when-let*`

* tide--npm-global looks at different path, mine is located at `/usr/local/lib/node_modules/typescript/lib/tsserver.js`. But I don't think we could get guess the global path correctly across all OS/configs etc. I am ok with this failing.

Other fallbacks seem to work as intended

tide.el
josteink: My reason for putting it down here, was that it refers to a function, a symbol which won't be defined at the time the `defcustom` is declared, unless we also move all of the supporting code.

I figured it was better to move the `defcustom` down, than to intermix 2-3 `defun`s into the otherwise clean `defcustom` section.

I guess if I made an empty `defalias` before the `defcustom` that would work too. Opinions?

tide.el
josteink: Agreed. Also we use `-when-let` and `-when-let*` everywhere else, so we should also stick to that here for consistency.

josteink:
> mine is located at /usr/local/lib/node_modules/typescript/lib/tsserver.js.

If that is the default on MacOS... It might apply on FreeBSD too? FreeBSD is pretty heavy on use of `/usr/local` after all.

> I am ok with this failing.

I'm not :)

I'm perfectly OK with probing this pretty standardized path too if `/usr/lib/node_modules` doesn't fan out.

Let's add a check like that to a new function `tide--npm-global-usrlocal` and add that to the overall strategy? It's not much code, and it's not like it will interfere with anything else we do.

Working seamlessly out of the box is a really nice thing, so I think it's worth putting some extra effort into those aspects.

tide.el
ananthakumaran: we could just use `(defcustom tide-tsserver-locator-function 'tide-tsserver-locater-npmlocal-projectile-npmglobal` without the `#'`, funcall works will plain symbols also

josteink: I don't think `#'` is actually ever *needed*, but it provides a way to be explicit about what points to functions and what doesn't. Using this consistently helps the compiler inform you about possible errors in your code.

Therefore (AFAIK) it's considered good Elisp-convention these days to use `#'` when pointing to functions.

Do you see any other reason for removing it, besides it technically not being needed?

ananthakumaran: `#'abc` is a shorthand(reader-macro) for `(function abc)`. I thought that was causing the problem. If it works, I have no issue with using it.

josteink: It works fine as far as I can see.

josteink: I just double-checked and test on a clean Emacs-session without any settings persisted.

It works fine as far as I can see.

tide.el
```diff
@@ -453,6 +458,59 @@ LINE is one based, OFFSET is one based and column is zero based"
       (kill-buffer (process-buffer process)))
     (tide-cleanup-project project-name)))
 
+(defun tide--npm-local ()
+  "Return a single path to the package-local typescript package directory or nil."
+
+  (-when-let ((packages-folder (locate-dominating-file default-directory "package.json")))
```
ananthakumaran: it should be 
```lisp
(-when-let (packages-folder (locate-dominating-file default-directory "package.json"))
    (concat packages-folder "node_modules/typescript/lib/"))
```
otherwise it always returns nil.

josteink: Indeed. Good catch. Will fix!

ananthakumaran:
👍 from my side. Perhaps others could pitch in to verify other OS

MarcusRiemer:
I gave this a quick shot on my Arch Linux machine (`GNU Emacs 25.3.1`) and everything seemed to work just fine OOTB. The only nitpick I have is that it might be valuable to extend the `(client-8e9d076b45) Starting tsserver...`  message to contain the path of the server that is started.

josteink:
> The only nitpick I have is that it might be valuable to extend the (client-8e9d076b45) Starting tsserver... message to contain the path of the server that is started.

I think we've had this discussion before, and the consensus was that it was hard to make this message informative without making it too long and noisy.

I suggest we re-visit it in a new issue once this is merged.

josteink:
I've also re-tested on Windows now and it works here too.

Unless anyone has any objections, I'll go ahead and merge and close this one, after only a month or so :)

lddubeau:
No objection here. :cake:

josteink:
That’s it then. Let’s close this one proper :)

PR #203, opened 2017-10-15
Add support for indirect buffers

### LINKED ISSUE #198: Warning (tide): A file backed buffer is required to start the tsserver.
when i run emacs in cygwin, "tide-setup" doesn't startup the tsserver

robbert-vdh:
This allows tide to be used in e.g. template files. The only caveat is that the file has to be manually added to tsconfig.json, as the globs in the include array ignore non-typescript files (see https://github.com/Microsoft/TypeScript/issues/8328). Otherwise the file will be added to an inferred project with default compiler settings. This fixes #198.

### DIFF
--- a/tide.el
+++ b/tide.el
@@ -212,15 +212,14 @@ above."
     (concat (file-name-nondirectory full-path) "-" (substring (md5 full-path) 0 10))))
 
 (defun tide-buffer-file-name ()
-  "Returns either the path to the currently open file. This is
-  needed to support indirect buffers, as they don't set
-  `buffer-file-name' correctly."
-  (or (and (bound-and-true-p edit-indirect--overlay)
-           (buffer-file-name (overlay-buffer edit-indirect--overlay)))
-      (and (bound-and-true-p org-src--overlay)
-           (buffer-file-name (overlay-buffer org-src--overlay)))
-      buffer-file-name
-      (buffer-file-name (buffer-base-buffer))))
+  "Returns the path to either the currently open file or the
+  current buffer's parent. This is needed to support indirect
+  buffers, as they don't set `buffer-file-name' correctly."
+  (buffer-file-name (or (and (bound-and-true-p edit-indirect--overlay)
+                             (overlay-buffer edit-indirect--overlay))
+                        (and (bound-and-true-p org-src--overlay)
+                             (overlay-buffer org-src--overlay))
+                        (buffer-base-buffer))))
 
 ;;; Compatibility
 
@@ -1626,7 +1625,7 @@ code-analysis."
   ;; tsconfig.json, otherwise the file will be loaded as part of an 'inferred
   ;; project'. This won't be necessary anymore after TypeScript allows defining
   ;; custom file extensions. https://github.com/Microsoft/TypeScript/issues/8328
-  (when tide-require-manual-sync
+  (when (and tide-require-manual-sync (tide-buffer-file-name))
     (tide-command:projectInfo
      (lambda (response)
        (tide-on-response-success response nil
--- a/tide.el
+++ b/tide.el
@@ -225,6 +225,9 @@ above."
                              (overlay-buffer edit-indirect--overlay))
                         (and (bound-and-true-p org-src--overlay)
                              (overlay-buffer org-src--overlay))
+                        ;; Needed for org-mode 8.x compatibility
+                        (and (bound-and-true-p org-edit-src-overlay)
+                             (overlay-buffer org-edit-src-overlay))
                         (buffer-base-buffer))))
 
 ;;; Compatibility
--- a/tide.el
+++ b/tide.el
@@ -192,7 +192,7 @@ above."
 (tide-def-permanent-buffer-local tide-buffer-dirty nil)
 (tide-def-permanent-buffer-local tide-buffer-tmp-file nil)
 (tide-def-permanent-buffer-local tide-active-buffer-file-name nil)
-(tide-def-permanent-buffer-local tide-require-manual-sync nil)
+(tide-def-permanent-buffer-local tide-require-manual-setup nil)
 
 (defvar tide-servers (make-hash-table :test 'equal))
 (defvar tide-response-callbacks (make-hash-table :test 'equal))
@@ -661,7 +661,7 @@ If TIDE-TSSERVER-EXECUTABLE is set by the user use it.  Otherwise check in the n
 
 (defun tide-command:openfile ()
   (tide-send-command "open"
-                     (if tide-require-manual-sync
+                     (if tide-require-manual-setup
                          `(:file
                            ,(tide-buffer-file-name)
                            :scriptKindName ,tide-default-mode
@@ -1614,7 +1614,7 @@ code-analysis."
   ;; case we should either add an ignore list or don't do anything at all when
   ;; there are more than a certain amount of snippets.
   (unless (stringp buffer-file-name)
-    (setq tide-require-manual-sync t))
+    (setq tide-require-manual-setup t))
 
   (tide-start-server-if-required)
   (tide-mode 1)
@@ -1630,7 +1630,7 @@ code-analysis."
   ;; tsconfig.json, otherwise the file will be loaded as part of an 'inferred
   ;; project'. This won't be necessary anymore after TypeScript allows defining
   ;; custom file extensions. https://github.com/Microsoft/TypeScript/issues/8328
-  (when (and tide-require-manual-sync (tide-buffer-file-name))
+  (when (and tide-require-manual-setup (tide-buffer-file-name))
     (tide-command:projectInfo
      (lambda (response)
        (tide-on-response-success response nil
--- a/tide.el
+++ b/tide.el
@@ -168,6 +168,12 @@ above."
   :type 'function
   :group 'tide)
 
+(defcustom tide-default-mode "TS"
+  "The default mode to open buffers not backed by files (e.g. Org
+  source blocks) in."
+  :type '(choice (const "TS") (const "TSX") (const "JS")(const  "JSX"))
+  :group 'tide)
+
 (defmacro tide-def-permanent-buffer-local (name &optional init-value)
   "Declare NAME as buffer local variable."
   `(progn
@@ -658,8 +664,7 @@ If TIDE-TSSERVER-EXECUTABLE is set by the user use it.  Otherwise check in the n
                      (if tide-require-manual-sync
                          `(:file
                            ,(tide-buffer-file-name)
-                           ;; TODO: Set this according to a variable or an alist
-                           :scriptKindName "TS"
+                           :scriptKindName ,tide-default-mode
                            :fileContent ,(buffer-string))
                        (append `(:file ,(tide-buffer-file-name))
                                (let ((extension (upcase (file-name-extension (tide-buffer-file-name)))))
--- a/tide.el
+++ b/tide.el
@@ -191,7 +191,7 @@ above."
 (defvar tide-servers (make-hash-table :test 'equal))
 (defvar tide-response-callbacks (make-hash-table :test 'equal))
 
-(defvar tide-source-root-directory (file-name-directory (or load-file-name (tide-buffer-file-name))))
+(defvar tide-source-root-directory (file-name-directory (or load-file-name buffer-file-name)))
 (defvar tide-tsserver-directory (expand-file-name "tsserver" tide-source-root-directory))
 
 (defun tide-project-root ()
@@ -277,6 +277,14 @@ ones and overrule settings in the other lists."
         (setq p (pop ls) v (pop ls))
         (setq rtn (plist-put rtn p v))))
     rtn))
+
+(defun tide-get-file-buffer (file)
+  "Returns a buffer associated with a file. This will return the
+  current buffer if it matches `file'. This way we can support
+  temporary and indirect buffers."
+  (if (equal file (tide-buffer-file-name))
+      (current-buffer)
+    (find-file-noselect file)))
 
 (defun tide-response-success-p (response)
   (and response (equal (plist-get response :success) t)))
@@ -720,8 +728,8 @@ implementations.  When invoked with a prefix arg, jump to the type definition."
     (unless no-marker
       (ring-insert find-tag-marker-ring (point-marker)))
     (if reuse-window
-        (pop-to-buffer (find-file-noselect file) '((display-buffer-reuse-window display-buffer-same-window)))
-      (pop-to-buffer (find-file-noselect file)))
+        (pop-to-buffer (tide-get-file-buffer file) '((display-buffer-reuse-window display-buffer-same-window)))
+      (pop-to-buffer (tide-get-file-buffer file)))
     (tide-move-to-location (plist-get filespan :start))))
 
 (defalias 'tide-jump-back 'pop-tag-mark)
@@ -734,7 +742,7 @@ implementations.  When invoked with a prefix arg, jump to the type definition."
 (defun tide-jump-to-implementation-format-item (item)
   (let* ((file-name (plist-get item :file))
          (line (save-excursion
-                 (with-current-buffer (find-file-noselect file-name)
+                 (with-current-buffer (tide-get-file-buffer file-name)
                    (tide-move-to-location (plist-get item :start))
                    (replace-regexp-in-string "\n" "" (thing-at-point 'line)))))
          (file-pos (concat
@@ -988,10 +996,14 @@ Noise can be anything like braces, reserved keywords, etc."
 (defun tide-apply-code-edits (file-code-edits)
   (save-excursion
     (dolist (file-code-edit file-code-edits)
-      (with-current-buffer (find-file-noselect (plist-get file-code-edit :fileName))
-        (tide-format-regions (tide-apply-edits (plist-get file-code-edit :textChanges)))
-        (basic-save-buffer)
-        (run-hooks 'tide-post-code-edit-hook)))))
+      (let ((file (plist-get file-code-edit :fileName)))
+        (with-current-buffer (tide-get-file-buffer file)
+          (tide-format-regions (tide-apply-edits (plist-get file-code-edit :textChanges)))
+          ;; Saving won't work if the current buffer is temporary or an indirect
+          ;; buffer
+          (when (equal buffer-file-name file)
+            (basic-save-buffer))
+          (run-hooks 'tide-post-code-edit-hook))))))
 
 (defun tide-get-flycheck-errors-ids-at-point ()
   (-map #'flycheck-error-id (flycheck-overlay-errors-at (point))))
@@ -1066,7 +1078,7 @@ Noise can be anything like braces, reserved keywords, etc."
       (deactivate-mark)
       (tide-apply-code-edits (tide-plist-get response :body :edits))
       (-when-let (rename-location (tide-plist-get response :body :renameLocation))
-        (with-current-buffer (find-file-noselect (tide-plist-get response :body :renameFilename))
+        (with-current-buffer (tide-get-file-buffer (tide-plist-get response :body :renameFilename))
           (tide-move-to-location rename-location)
           (when (tide-can-rename-symbol-p)
             (tide-rename-symbol)))))))
@@ -1431,7 +1443,7 @@ number."
 (defun tide-rename-symbol-at-location (location new-symbol)
   (let ((file (plist-get location :file)))
     (save-excursion
-      (with-current-buffer (find-file-noselect file)
+      (with-current-buffer (tide-get-file-buffer file)
         (-each
             (-map (lambda (filespan)
                     (tide-move-to-location (plist-get filespan :start))
@@ -1443,7 +1455,10 @@ number."
               (goto-char marker)
               (delete-char (- (tide-plist-get filespan :end :offset) (tide-plist-get filespan :start :offset)))
               (insert new-symbol))))
-        (basic-save-buffer)
+        ;; Saving won't work if the current buffer is temporary or an indirect
+        ;; buffer
+        (when (equal buffer-file-name file)
+          (basic-save-buffer))
         (length (plist-get location :locs))))))
 
 (defun tide-read-new-symbol (old-symbol)
--- a/tide.el
+++ b/tide.el
@@ -186,11 +186,12 @@ above."
 (tide-def-permanent-buffer-local tide-buffer-dirty nil)
 (tide-def-permanent-buffer-local tide-buffer-tmp-file nil)
 (tide-def-permanent-buffer-local tide-active-buffer-file-name nil)
+(tide-def-permanent-buffer-local tide-require-manual-sync nil)
 
 (defvar tide-servers (make-hash-table :test 'equal))
 (defvar tide-response-callbacks (make-hash-table :test 'equal))
 
-(defvar tide-source-root-directory (file-name-directory (or load-file-name buffer-file-name)))
+(defvar tide-source-root-directory (file-name-directory (or load-file-name (tide-buffer-file-name))))
 (defvar tide-tsserver-directory (expand-file-name "tsserver" tide-source-root-directory))
 
 (defun tide-project-root ()
@@ -209,6 +210,17 @@ above."
 (defun tide-project-name ()
   (let ((full-path (directory-file-name (tide-project-root))))
     (concat (file-name-nondirectory full-path) "-" (substring (md5 full-path) 0 10))))
+
+(defun tide-buffer-file-name ()
+  "Returns either the path to the currently open file. This is
+  needed to support indirect buffers, as they don't set
+  `buffer-file-name' correctly."
+  (or (and (bound-and-true-p edit-indirect--overlay)
+           (buffer-file-name (overlay-buffer edit-indirect--overlay)))
+      (and (bound-and-true-p org-src--overlay)
+           (buffer-file-name (overlay-buffer org-src--overlay)))
+      buffer-file-name
+      (buffer-file-name (buffer-base-buffer))))
 
 ;;; Compatibility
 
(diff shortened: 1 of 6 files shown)


josteink:
This looks like an amazing (and a little big) pull.

I’ll try to get it tested, reviewed and provide comments over the next few days. 

Anyway, I have to say already now: Awesome. Glad to see you pulled through on this one!

robbert-vdh:
I was surprised that this actually worked as well as it does. I expected that we'd have to manually send changes in the buffer to the server process, but somehow everything works out of the box. The only problem is that TypeScript doesn't include non-typescript files in the project unless you manually tell it to.

josteink:
Is that something we can fix in the configure-command?

robbert-vdh:
I don't think so. I tried manually specifying the location of tsconfig.json through the `projectRootName` parameter, but then tsserver starts throwing exceptions around because it doesn't see the file as part of that project. Until TypeScript supports custom file extensions/globs, I think the only options are either automatically adding the file to the files array (though that might not always be what you want) or telling the user to do it themselves.

ananthakumaran:
Thanks for the PR. I haven't gone through entire patch, I just want to point out some issue that might occur

* replacing `buffer-file-name` with `tide-buffer-file-name` will not work in all the cases. The first one is a value that is kept up to date by emacs. It could be changed by file rename operation. `tide-buffer-file-name` on the hand is a snapshot of buffer-file-name. Currently, tide automatically re-configure the buffer if the file is renamed(grep for tide-active-buffer-file-name), This would likely break that functionality

* does it work if there are more than one indirect buffers in the same file?

* how do we handle cases like refactor/rename-symbol etc where tsserver will give us the filename/pos and ask us to change. Does it work correctly now? because I guess we would be trying to apply it in the parent file and it might not work

robbert-vdh:
> replacing buffer-file-name with tide-buffer-file-name will not work in all the cases. The first one is a value that is kept up to date by emacs. It could be changed by file rename operation. tide-buffer-file-name on the other hand is a snapshot of buffer-file-name. Currently, tide automatically re-configure the buffer if the file is renamed(grep for tide-active-buffer-file-name), This would likely break that functionality. Probably a function would do the trick

Good catch! I'll move it to a function.

> does it work if there are more than one indirect buffers in the same file?

I was worried about that, but it shouldn't be a problem as the actual major mode hooks are only executed when the indirect buffer gets opened through `edit-indirect-region`. Org-mode is an exception, but that also shouldn't be a problem as multiple source blocks would just overwrite eachother (`ob-typescript` doesn't let you share state between multiple blocks anyway).

> how do we handle cases like refactor/rename-symbol etc where tsserver will give us the filename/pos and ask us to change. Does it work correctly now? because I guess we would be trying to apply it in the parent file and it might not work

I just tried it. Right now `tide-apply-refactor` does indeed try to edit the parent buffer instead of the opened indirect buffer, so that might need some work to get working.

I also noticed that this approach doesn't work for org-mode documents yet as they don't use `edit-indirect.el` and `tide-setup` doesn't execute, but that should be an easy fix.

ananthakumaran:
could you also add a minimal vuejs/org file inside the example folder, so it would be easier to test? I have never used vuejs myself.

robbert-vdh:
I haven't had the time to properly test everything yet, but refactoring should now work in temporary and indirect buffers, and there should be no more errors in org-mode.

ananthakumaran:
How well this works for `.vue` files? It seems like we take care of rename that originates from the indirect buffer. What happens when we rename symbols in different ts file, that is also referenced by `.vue` file. I tried it in different scenarios and I get mixed result

1) rename Person class in test.ts, it works surprising. I guess tsserver is still able to parse the file as typescript and it considers all the tags as syntax errors..

2) edit the ts code in `.vue` file and commit the result. Now tsserver thinks that `.vue` file contains only ts code. If you go to test.ts and do the rename again, the refactor is applied at the wrong place in `.vue` file.

All this issue might go away if tsserver is aware of the `.vue` file. Have you tried https://github.com/sandersn/vue-ts-plugin

robbert-vdh:
I hadn't heard of vue-ts-plugin before. It looks like it extracts the script part of the component. That would allow for cross-file refactoring and imports without a shim (you'd normally use a [shim](https://github.com/robbert-vdh/4-6-calculator/blob/82069dfb03b50eb74252bce22507fb273bc4875d/src/vue-shims.d.ts) for importing .vue components and let webpack do all the compilation), but using it will probably break editing as the plugin assumes tsserver is looking at the entire file, not just the script part. The [official VsCode plugin](https://github.com/vuejs/vetur) seems to solve it by bundling an actual language server, but that's way seems way out of scope. It would be much better if we could somehow associate a file with a TypeScript project without having to add it to tsconfig.json. That would still not fix cross-file renaming, but it would at least not break anything. You'd probably want to keep your components as simple as possible anyway, and some refactors could potentially break the template part of the component.

tide.el
```diff
@@ -170,11 +170,12 @@ above."
 (tide-def-permanent-buffer-local tide-buffer-dirty nil)
 (tide-def-permanent-buffer-local tide-buffer-tmp-file nil)
 (tide-def-permanent-buffer-local tide-active-buffer-file-name nil)
+(tide-def-permanent-buffer-local tide-require-manual-sync nil)
 
 (defvar tide-servers (make-hash-table :test 'equal))
 (defvar tide-response-callbacks (make-hash-table :test 'equal))
 
-(defvar tide-source-root-directory (file-name-directory (or load-file-name buffer-file-name)))
+(defvar tide-source-root-directory (file-name-directory (or load-file-name (tide-buffer-file-name))))
```
josteink: In this particular case, aren't we actually interested in the `buffer-file-name` (since the buffer is going to be `tide.el`, not something the user is editing)?

*If* we're actually interested in `(tide-buffer-file-name)`, this `defvar` has one issue: it's introduced in the source before `tide-buffer-file-name` is, meaning evaluating this statement interactively from scratch will fail.

For a `tide`-developer, that's pretty annoying, not to mention something I just fixed, so I'd like for it not to get broken again immediately :)

tide.el
```diff
@@ -194,6 +195,16 @@ above."
   (let ((full-path (directory-file-name (tide-project-root))))
     (concat (file-name-nondirectory full-path) "-" (substring (md5 full-path) 0 10))))
 
+(defun tide-buffer-file-name ()
+  "Returns either the path to the currently open file. This is
+  needed to support indirect buffers, as they don't set
```
josteink: This function seems to work as intended, but the docstring has some issues.

Returns either {something} ... or what?

tide.el
```diff
@@ -250,6 +261,14 @@ ones and overrule settings in the other lists."
...
     rtn))
 
+(defun tide-get-file-buffer (file)
+  "Returns a buffer associated with a file. This will return the
+  current buffer if it matches `file'. This way we can support
+  temporary and indirect buffers."
+  (if (equal file (tide-buffer-file-name))
+      (current-buffer)
+    (find-file-noselect file)))
+
```
josteink: This is going to leave buffers lying around after we're done editing them.

I'm not saying that is *wrong* per se, but I'm just wondering if it's intentional.

tide.el
```diff
@@ -613,34 +632,39 @@ If TIDE-TSSERVER-EXECUTABLE is set by the user use it.  Otherwise check in the n
...
   (tide-send-command "open"
-                     (append `(:file ,buffer-file-name)
-                             (let ((extension (upcase (file-name-extension buffer-file-name))))
-                               (if (member extension '("TS" "JS" "TSX" "JSX"))
-                                   `(:scriptKindName ,extension)
-                                 nil)))))
+                     (if tide-require-manual-sync
+                         `(:file
+                           ,(tide-buffer-file-name)
+                           ;; TODO: Set this according to a variable or an alist
```
josteink: Is this still something which needs to be done, or is this just a leftover comment from when you worked on it?

tide.el
```diff
@@ -170,11 +170,12 @@ above."
 (tide-def-permanent-buffer-local tide-buffer-dirty nil)
 (tide-def-permanent-buffer-local tide-buffer-tmp-file nil)
 (tide-def-permanent-buffer-local tide-active-buffer-file-name nil)
+(tide-def-permanent-buffer-local tide-require-manual-sync nil)
```
josteink: I can tell we're not exactly leading by example in the existing code...

But not yet being fully familiar with this code, I'd really appreciate a docstring for when I do a `apropos` lookup. (Like the very descriptive code-comments in `tide-setup`)

Feel free to extend the `tide-def-permanent-buffer-local` to add an option for such docstrings.

tide.el
```diff
@@ -941,30 +965,35 @@ Noise can be anything like braces, reserved keywords, etc."
...
-        (basic-save-buffer)
-        (run-hooks 'tide-post-code-edit-hook)))))
+      (let ((file (plist-get file-code-edit :fileName)))
+        (with-current-buffer (tide-get-file-buffer file)
+          (tide-format-regions (tide-apply-edits (plist-get file-code-edit :textChanges)))
+          ;; Saving won't work if the current buffer is temporary or an indirect
+          ;; buffer
+          (when (equal buffer-file-name file)
+            (basic-save-buffer))
+          (basic-save-buffer)
```
josteink: I really appreciate comments in code-sections like this. It helps guide the intent behind the code in ways the code alone doesn't fully convey.

That said: Here the comment says saving may not work... And then immediately after the `(when)`-guard... we save anyway (and potentially ending up saving twice).

This looks fishy or it needs even *better* documentation to explain why this isn't wrong.

tide.el
```diff
@@ -1397,7 +1426,10 @@ number."
               (goto-char marker)
               (delete-char (- (tide-plist-get filespan :end :offset) (tide-plist-get filespan :start :offset)))
               (insert new-symbol))))
-        (basic-save-buffer)
+        ;; Saving won't work if the current buffer is temporary or an indirect
+        ;; buffer
+        (when (equal buffer-file-name file)
+          (basic-save-buffer))
         (length (plist-get location :locs))))))
```
josteink: And here you *don't* save twice. Basically this is code-repetition where we introduce the possibility of inconsistencies and bugs.

Maybe this could be extracted into a helper-function to avoid such errors in the future?

tide.el
```diff
@@ -1528,20 +1561,36 @@ code-analysis."
...
+  ;; project'. This won't be necessary anymore after TypeScript allows defining
+  ;; custom file extensions. https://github.com/Microsoft/TypeScript/issues/8328
+  (when (and tide-require-manual-sync (tide-buffer-file-name))
+    (tide-command:projectInfo
+     (lambda (response)
+       (tide-on-response-success response nil
+         (when (string-prefix-p "/dev/null/inferredProject"
+                                (tide-plist-get response :body :configFileName))
+           (message (format "'%s' is not part of a project, add it to the files array in tsconfig.json"
+                            (tide-buffer-file-name)))))))))
```
josteink: I really appreciate code like this which guides the user when there's nothing we can do about to situation, not to mention the code-comments helping us decide when things like this can possibly be obsoleted it in the future.

Very nice.

josteink:
Hey Robbert!

I haven't had time to do very much *interactive testing* yet, but at least now I've looked over the code to see what kind of changes was done.

All in all I have to say I really like it, and for several reasons:

1. Lot's of comments. Lot's of trying to do the right thing.
2. You've been very good at isolating your changes from the remaining parts of the code.

That's also quite a diff, and despite what may look like lots of comment, I wasn't able to find much in terms of things I thought were *wrong* or needed severe changes.

This is clearly 10 out of 10 for an initial contribution!

Let me know what you think about my comments so far, and I can try to do some actual end-user testing too :)

tide.el
robbert-vdh: Good catch, that's what you get for carelessly find-replacing.

tide.el
robbert-vdh: How would you propose closing unneeded buffers?

tide.el
robbert-vdh: It would only be needed to support TSX in a an indirect buffer. Maybe a simple variable that could be overridden in the dir-local variables would suffice?

tide.el
robbert-vdh: I originally thought that we'd have to manually synchronize the buffer's contents to tsserver (through [ChangeRequest](https://github.com/Microsoft/TypeScript/blob/ceba50750b7bd6c8cfc21a5886e6eb9de7f82a36/lib/protocol.d.ts#L1710)), but it turns out that tsserver is smart enough to use diffs after you pass it the [initial file contents](https://github.com/robbert-vdh/tide/blob/2b1f4816881c73519da5c19c740a72b581503c2c/tide.el#L647), so right now that variable is only needed for setting up the buffer. I'll add docstrings so we can document that!

tide.el
robbert-vdh: Good catch, that wasn't intended.

robbert-vdh:
Thank you very much for the help! 'm quite new to elisp and Emacs' inner workings in general, so a little guidance comes very welcome.

josteink:
Now I've done some functional testing, and it's clearly much better than it used to be :)

That said, I seem to be having some issues compared to what I expected. When diving into the typescript babel-section in `literate.org` I get no warning about file not being "supported". As such, I expect most things to work.

And most things did work like wonders (like eldoc, highlighting, navigation, etc). Amazing stuff. Amazing improvement. Let's just say that before diving into the issues :)

I'm not sure whether you are at fault here or not, but I did find some things which did not work entirely. 

Some of these issues are not really so bad in themselves, but they get compounded by one single issue which has potential for causing a lot of confusion: seemingly delayed `flycheck` reports/errors.

Basically errors in my code are not reported as errors until I finish my editing (as in close the indirect buffer) and re-open it.

As an example, adding a new variable named x at the end of `literate.org`:

````typescript
let x = new Snake("boo");
x.|
````

With the cursor at  `x.` I expect company-completions for the type `Snake`, but I get nothing. How so?

If I save the buffer, close it and re-open it, I *now* get the `flycheck` error telling me that `x` is already defined and I can't redefine it now...

Empowered by this knowledge, I rename the variable to `x1` (manually, no refactor) and voila, everything works! But `flycheck` didn't tell me until I fully closed the buffer and re-opened it.

Do you think this is something you can fix? Is this related to your concept of manually synched buffers?

Also: where did `tsserver` pick up `x` as a variable? If I remove the `let` so we're just left with basic assignment... And I try to navigate... I get to this in `another.js`:

````javascript
var x = {
    another: 'name'
};
````

Well that's certainly interesting! Maybe `tsserver` compounds all literate and non-module files into one giant global namespace?

To test out this theory I created another literate section above your's and added the following to it:

````org-mode
#+BEGIN_SRC typescript
  abstract class AnimalBase {
      canMove: boolean;
  }
#+END_SRC
````

And after that I changed your `Animal`-class (in the other section) to derive from it. But then `tsserver` complained that `AnimalBase` was not defined... So clearly that is *not* in a global namespace.

(Not sure if this is your fault or not though. But just mentioning it as a finding.)

Anyone have any idea what we should call "expected behaviour" in this case? With some things automatically imported and others not... This can quickly get confusing for the average org-user.

That said... All in all, I'm very much for these changes I you have my full support for getting this PR merged once we get all the small details settled :)

File: tide.el (hunk 3 above)
Comment (josteink) [reply tide.el]: Looking around the code other places, we are pretty careless about usage of `find-file-noselect` ourselves already.

I don't have any issues with this per se, and from inspecting my buffer list after working with this, I can't say I see Emacs literred with new, additional leftover buffers.

I guess you can disregard this comment. This is fine as it is.

File: tide.el (hunk 4 above)
Comment (josteink) [reply tide.el]: As far as I understand (not developing React/JSX/TSX myself), our JSX/TSX-support is already a little spotty.

Not having 100% support out of the box for *literate* TSX is IMO not going to be a big deal, and especially if we don't have an easy way to reliably establish what kind of scripting-kind this is.

Unless anyone else objects or has a way to properly determine this, I say `TS` as a default is fine for now.

File: tide.el (hunk 5 above)
Comment (josteink) [reply tide.el]: > it turns out that tsserver is smart enough to use diffs after you pass it the initial file contents

Should this in theory also apply to `flycheck`? Because `flycheck` is not heaving properly in interactive editing mode, only on initial buffer-load.

**Comment by robbert-vdh** [comment]
That's weird, I'll look into it this weekend. I haven't experienced any such issues ([gfy](https://gfycat.com/PlayfulDearestGalapagosmockingbird)), but I might be using the version of tsserver that comes bundled with tsc 2.5.3 instead of the one included in the tide repo.

**Comment by josteink** [comment]
Sounds good.

No rush. Take whatever time you need. :)

File: tide.el (hunk 5 above)
Comment (ananthakumaran) [reply tide.el]: > but it turns out that tsserver is smart enough to use diffs after you pass it the initial file contents,

No tsserver doesn't do anything. Once we open a file using `open` command, tide is supposed to keep the file in sync. Tide currently reloads the entire file see `tide-sync-buffer-contents`.

> Should this in theory also apply to flycheck? Because flycheck is not heaving properly in interactive editing mode, only on initial buffer-load.

if you have the config in readme `(setq flycheck-check-syntax-automatically '(save mode-enabled))`, flycheck is triggered only on save and mode-enabled. You probably need to add newline to the list I guess.

File: tide.el (hunk 6 above)
Comment (ananthakumaran) [reply tide.el]: it would be even better to create a function with a proper name, like

```elisp
(unless (tide-indirect-buffer-p file)
   (basic-save-buffer))
```
The intent is not clear when you read the code

josteink:
Hey there. Any news on this one? :)

robbert-vdh:
I kind of forgot about it! I'll look into it this evening.

robbert-vdh:
I've fixed some minor things, but I could not reproduce your issues @josteink. Are they still happening? I did notice that it took a newline for flycheck to update when switching between blocks in the .org file as tsserver doesn't know the file's 'contents' have been changed.

I think the only big remaining issue is cross file refactoring (https://github.com/ananthakumaran/tide/pull/203#issuecomment-337964019). Not sure how you'd tackle that one though.

tide.el
josteink: You forgot to fix this one :)

josteink:
I'm now at work where I use windows... And while I'm not sure that matters, I can't get any literate examples to work at all now.

I'm just getting lots of errors inside `tsserver`:

````
Error processing request. Cannot read property 'replace' of null
TypeError: Cannot read property 'replace' of null
    at normalizeSlashes (d:\Throwaway\test-tide\node_modules\typescript\lib\tsserver.js:2452:20)
    at normalizePathAndParts (d:\Throwaway\test-tide\node_modules\typescript\lib\tsserver.js:2507:16)
    at Object.normalizePath (d:\Throwaway\test-tide\node_modules\typescript\lib\tsserver.js:2503:16)
    at Object.toNormalizedPath (d:\Throwaway\test-tide\node_modules\typescript\lib\tsserver.js:81176:23)
    at IOSession.Session.getFileAndProjectWorker (d:\Throwaway\test-tide\node_modules\typescript\lib\tsserver.js:86054:35)
    at IOSession.Session.getFileAndProject (d:\Throwaway\test-tide\node_modules\typescript\lib\tsserver.js:86040:29)
    at IOSession.Session.getQuickInfoWorker (d:\Throwaway\test-tide\node_modules\typescript\lib\tsserver.js:86100:31)
    at Session.handlers.ts.createMapFromTemplate._a.(anonymous function) (d:\Throwaway\test-tide\node_modules\typescript\lib\tsserver.js:85269:61)
....
````

I guess I can double-check if this works better on Linux when I get home, I guess.

In the meantime, we did leave some comments behind on your former commits, and I can't see them having been addressed. (Simple-to-fix stuff like double-saving, etc)

Could you just take a look over those comments again? :)

robbert-vdh:
I've gotten that error before, that seems to happen when tsserver does not receive a valid filename (i.e. when `(tide-buffer-file-name)` returns `nil`).

Oh and sorry for that! I was sure I had already fixed those things, but I might have forgotten to push the changes.

josteink:
Just reporting in that I\ve now tested on Linux and that seems to work fine,

I'm still having issues with flycheck though, but even if we can't figure that one out/or get it fully resolved, I'd still consider this a massive improvement.

I'll see if I can figure out why this doesn't seem to work on my Windows-machine at work though. Not having proper cross-platform support would be quite a bummer.

josteink:
As far as I can diagnose this right now, it seems like this fails on Windows because the value we send in for file is null:

````
    {"seq":0,"type":"response","command":"semanticDiagnosticsSync","request_seq":"28","success":false,"message":"Error processing request. Cannot read property 'replace' of null\nTypeError: Cannot read property 'replace' of null\n    at normalizeSlashes (d:\\Git\\tstestcase\\node_modules\\typescript\\lib\\tsserver.js:2461:20)\n    at normalizePathAndParts (d:\\Git\\tstestcase\\node_modules\\typescript\\lib\\tsserver.js:2516:16)\n    at Object.normalizePath (d:\\Git\\tstestcase\\node_modules\\typescript\\lib\\tsserver.js:2512:16)\n    at Object.toNormalizedPath (d:\\Git\\tstestcase\\node_modules\\typescript\\lib\\tsserver.js:81455:23)\n    at IOSession.Session.getConfigFileAndProject (d:\\Git\\tstestcase\\node_modules\\typescript\\lib\\tsserver.js:86018:35)\n    at IOSession.Session.getSemanticDiagnosticsSync (d:\\Git\\tstestcase\\node_modules\\typescript\\lib\\tsserver.js:86163:31)\n    at Session.handlers.ts.createMapFromTemplate._a.(anonymous function) (d:\\Git\\tstestcase\\node_modules\\typescript\\lib\\tsserver.js:85695:61)\n    at d:\\Git\\tstestcase\\node_modules\\typescript\\lib\\tsserver.js:87022:88\n    at IOSession.Session.executeWithRequestId (d:\\Git\\tstestcase\\node_modules\\typescript\\lib\\tsserver.js:87013:28)\n    at IOSession.Session.executeCommand (d:\\Git\\tstestcase\\node_modules\\typescript\\lib\\tsserver.js:87022:33)\n    at IOSession.Session.onMessage (d:\\Git\\tstestcase\\node_modules\\typescript\\lib\\tsserver.js:87042:35)\n    at Interface.<anonymous> (d:\\Git\\tstestcase\\node_modules\\typescript\\lib\\tsserver.js:88234:27)\n    at emitOne (events.js:96:13)\n    at Interface.emit (events.js:188:7)\n    at Interface._onLine (readline.js:239:10)\n    at Interface.<anonymous> (readline.js:376:12)\n    at Array.forEach (native)\n    at Interface._normalWrite (readline.js:375:11)\n    at Socket.ondata (readline.js:107:10)\n    at emitOne (events.js:96:13)\n    at Socket.emit (events.js:188:7)\n    at readableAddChunk (_stream_readable.js:176:18)\n    at Socket.Readable.push (_stream_readable.js:134:10)\n    at Pipe.onread (net.js:548:20)"}
Info 16   [7:51:30.349] request:
    {"command":"navtree","seq":"29","arguments":{"file":null}}
Err 17    [7:51:30.350] Exception on executing command {"command":"navtree","seq":"29","arguments":{"file":null}}:

    Cannot read property 'replace' of null

    TypeError: Cannot read property 'replace' of null
    at normalizeSlashes (d:\Git\tstestcase\node_modules\typescript\lib\tsserver.js:2461:20)
    at normalizePathAndParts (d:\Git\tstestcase\node_modules\typescript\lib\tsserver.js:2516:16)
    at Object.normalizePath (d:\Git\tstestcase\node_modules\typescript\lib\tsserver.js:2512:16)
    at Object.toNormalizedPath (d:\Git\tstestcase\node_modules\typescript\lib\tsserver.js:81455:23)
    at IOSession.Session.getFileAndLanguageServiceForSyntacticOperation (d:\Git\tstestcase\node_modules\typescript\lib\tsserver.js:86396:35)
    at IOSession.Session.getNavigationTree (d:\Git\tstestcase\node_modules\typescript\lib\tsserver.js:86724:31)
    at Session.handlers.ts.createMapFromTemplate._a.(anonymous function) (d:\Git\tstestcase\node_modules\typescript\lib\tsserver.js:85750:61)
    at d:\Git\tstestcase\node_modules\typescript\lib\tsserver.js:87022:88
````

I guess I'll have to verify what the behaviour is on Linux, and see if there's something in our path-handling logic which acts differently there.

I'd love to get this fixed before landing it proper.

josteink:
@robbert-vdh : I feel like I'm the slowest reviewer and tester on earth, but I'd like to ensure things works properly before I merge. Sorry about the delay, and I hope you're still with me.

Running (an old) Emacs 24.5.1 on Ubuntu 16.04 I can reproduce the same exact same issues as I did on Windows (where I'm running 25.3.1). This leads me to believe that this is somehow dependant on other external factors (org-version?) to function 100% correctly, and that it's not tied to Windows being Windows per se.

From what I can tell, it all boils down to `tide-buffer-file-name` always returning `nil` in babel-buffers. Should it return something else? What *should* it return in a org-babel source-code buffer?

Inspecting its implementation and trying to edebug it, I keep finding all the (org) "overlay" parts undefined. Should they be defined? I can't find them in the org source-code.

Care to give me some leads? I'd hate to see this effort go wasted :)

robbert-vdh:
No problem at all, I've been very busy as well. I think you might be onto something. `(overlay-buffer org-src--overlay)` is supposed to return the buffer the src block was opened from (i.e. the parent buffer) and it works as expected with the latest melpa version of `org-plus-contrib` (tested in Spacemacs). I quickly looked at org-mode's git history and the variable seems to be renamed from `org-edit-src-overlay` back in 2014, meaning that Org 8.x uses the old `org-edit-src-overlay`. Adding backwards compatibility should be trivial unless other parts are also broken.

Then there's still the issue of cross-file refactoring writing the wrong parts of files, though I'm not sure if there's anything we can do about that without modifying tsserver.

josteink:
> Adding backwards compatibility should be trivial unless other parts are also broken.

Good good. Make it so.

> Then there's still the issue of cross-file refactoring writing the wrong parts of files, though I'm not sure if there's anything we can do about that without modifying `tsserver`.

That's not a show-stopper by any means. I'm fairly certain most people will understand that this not working 100% correctly is not a shortcoming of `tide`, and that working with complex refactorings outside "regular" typescript-projects has limitations.

If you can fix that org-compatibility-issue, and I can verify it working on my "boring" Windows install, I think we can consider this good to go and ready for merge!

robbert-vdh:
I've rebased and added the fix, hopefully it works now!

josteink:
That's a confirmed fix on my end. Consider it merged!

Thanks for taking time to develop and contribute this awesome feature! Feel free to stay around and send us patches whenever you see something which can be improved. 

Contributions are always welcome, and I promise next time around I won't spend this long reviewing :)

PR #205, opened 2017-11-09
add support for auto-import on code completion

ananthakumaran:
**Testing notes**
this feature works only with tsserver >= 2.6.1, but should not break anything in old tsserver

### DIFF
--- a/tide.el
+++ b/tide.el
@@ -973,6 +973,12 @@ Noise can be anything like braces, reserved keywords, etc."
   "Apply a single `FIX', which may apply to several files."
   (tide-apply-code-edits (plist-get fix :changes)))
 
+(defun tide-apply-codefixes (fixes)
+  (cond ((= 0 (length fixes)) (message "No code-fixes available."))
+        ((= 1 (length fixes)) (tide-apply-codefix (car fixes)))
+        (t (tide-apply-codefix
+            (tide-select-item-from-list "Select fix: " fixes #'tide-get-fix-description (tide-can-use-popup-p 'code-fix))))))
+
 
 (defun tide-fix ()
   "Apply code fix for the error at point."
@@ -982,10 +988,7 @@ Noise can be anything like braces, reserved keywords, etc."
   (let ((response (tide-command:getCodeFixes)))
     (tide-on-response-success response nil
       (let ((fixes (plist-get response :body)))
-        (cond ((= 0 (length fixes)) (message "No code-fixes available."))
-              ((= 1 (length fixes)) (tide-apply-codefix (car fixes)))
-              (t (tide-apply-codefix
-                  (tide-select-item-from-list "Select fix: " fixes #'tide-get-fix-description (tide-can-use-popup-p 'code-fix)))))))))
+        (tide-apply-codefixes fixes)))))
 
 ;;; Refactor
 
@@ -1145,8 +1148,11 @@ Noise can be anything like braces, reserved keywords, etc."
           (tide-annotate-completions (plist-get response :body) prefix file-location)))))))
 
 (defun tide-command:completionEntryDetails (name)
-  (let ((arguments (-concat (get-text-property 0 'file-location name)
-                            `(:entryNames (,name)))))
+  (let* ((source (plist-get (get-text-property 0 'completion name) :source))
+         (arguments (-concat (get-text-property 0 'file-location name)
+                            (if source
+                                `(:entryNames [(:name ,name :source ,source)])
+                              `(:entryNames (,name))))))
     (-when-let (response (tide-send-command-sync "completionEntryDetails" arguments))
       (when (tide-response-success-p response)
         response))))
@@ -1168,6 +1174,13 @@ Noise can be anything like braces, reserved keywords, etc."
                (detail (car (plist-get response :body))))
     (tide-construct-documentation detail)))
 
+(defun tide-post-completion (name)
+  (when (plist-get (get-text-property 0 'completion name) :hasAction)
+    (-when-let* ((response (tide-completion-entry-details name))
+                 (detail (car (plist-get response :body)))
+                 (fixes (plist-get detail :codeActions)))
+      (tide-apply-codefixes fixes))))
+
 ;;;###autoload
 (defun company-tide (command &optional arg &rest ignored)
   (interactive (list 'interactive))
@@ -1186,7 +1199,8 @@ Noise can be anything like braces, reserved keywords, etc."
     (ignore-case tide-completion-ignore-case)
     (meta (tide-completion-meta arg))
     (annotation (tide-completion-annotation arg))
-    (doc-buffer (tide-completion-doc-buffer arg))))
+    (doc-buffer (tide-completion-doc-buffer arg))
+    (post-completion (tide-post-completion arg))))
 
 (eval-after-load 'company
   '(progn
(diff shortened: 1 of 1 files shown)


josteink:
Before diving too deeply into the code, I've just done some preliminary testing.

I realize some of this may be up to `tsserver`, but I'd just like to make sure I've understood how the functionality is supposed to work before making any deeper judgement.

I've created a new TS-project from scratch with TypeScript 2.6.1. In this project I've created 2 TS-files, one exporting a function and a class.

````typescript
export function eventFactory(name: string) {
    return new EventMaster(name);
}

export class EventMaster {
    constructor(public name: string) {
    }
}
````

In the other file I've been testing the completions and imports.

Now if I try to get completions from the other file without any kinds of imports, nothing happens. I don't get the right completion-candidates, and thus there's nothing to auto-import either.

If I however manually import one member from the other file, like this:

````typescript
import { eventFactory } from "./eventMaster";

let x = eventFactory("moo");
````

And then proceed to try to type `Event`, I get `EventMaster` as a completion, and my import-statement is extended to include this exported member too. (Basically it does the right thing. Very nice!)

But this seems inconsistent from your screencast. Is this how it's supposed to work?

File: tide.el [hunk 1]
```diff
@@ -1145,8 +1148,11 @@ Noise can be anything like braces, reserved keywords, etc."
...
 
 (defun tide-command:completionEntryDetails (name)
-  (let ((arguments (-concat (get-text-property 0 'file-location name)
-                            `(:entryNames (,name)))))
+  (let* ((source (plist-get (get-text-property 0 'completion name) :source))
+         (arguments (-concat (get-text-property 0 'file-location name)
+                            (if source
+                                `(:entryNames [(:name ,name :source ,source)])
+                              `(:entryNames (,name))))))
     (-when-let (response (tide-send-command-sync "completionEntryDetails" arguments))
```
josteink: Why are we making these changes? Are these changes to `arguments`required to get more code-fixes back with the completion-results?

Again: I feel the intent and reasoning behind the code here isn't communicated properly, and I'm afraid that can lead to problems if we're going to make changes here in the future.

tide.el
```diff
@@ -1145,8 +1148,11 @@ Noise can be anything like braces, reserved keywords, etc."
           (tide-annotate-completions (plist-get response :body) prefix file-location)))))))
 
 (defun tide-command:completionEntryDetails (name)
-  (let ((arguments (-concat (get-text-property 0 'file-location name)
-                            `(:entryNames (,name)))))
+  (let* ((source (plist-get (get-text-property 0 'completion name) :source))
```
josteink: On overall I don't think this code is very self-explanatory. Is `:source` a auto-complete thing here? or does it belong to `tsserver`. It's hard to tell by looking at this code alone.

tide.el
```diff
@@ -1168,6 +1174,13 @@ Noise can be anything like braces, reserved keywords, etc."
                (detail (car (plist-get response :body))))
     (tide-construct-documentation detail)))
 
+(defun tide-post-completion (name)
+  (when (plist-get (get-text-property 0 'completion name) :hasAction)
+    (-when-let* ((response (tide-completion-entry-details name))
+                 (detail (car (plist-get response :body)))
+                 (fixes (plist-get detail :codeActions)))
+      (tide-apply-codefixes fixes))))
+
```
josteink: This code is very nice and easy to follow. If I were to nitpick anything it might be form...

Would it be better to put this all into one big `(-when-let*)` instead of having that `(when)`-form proceeding it? I can't decide :)

tide.el
ananthakumaran:
```typescript
    export interface CompletionEntry {
        /**
         * The symbol's name.
         */
        name: string;
        /**
         * The symbol's kind (such as 'className' or 'parameterName').
         */
        kind: ScriptElementKind;
        /**
         * Optional modifiers for the kind (such as 'public').
         */
        kindModifiers: string;
        /**
         * A string that is used for comparing completion items so that they can be ordered.  This
         * is often the same as the name but may be different in certain circumstances.
         */
        sortText: string;
        /**
         * An optional span that indicates the text to be replaced by this completion item. If present,
         * this span should be used instead of the default one.
         */
        replacementSpan?: TextSpan;
        /**
         * Indicates whether commiting this completion entry will require additional code actions to be
         * made to avoid errors. The CompletionEntryDetails will have these actions.
         */
        hasAction?: true;
        /**
         * Identifier (not necessarily human-readable) identifying where this completion came from.
         */
        source?: string;
    }

    /**
     * Arguments for completion details request.
     */
    export interface CompletionDetailsRequestArgs extends FileLocationRequestArgs {
        /**
         * Names of one or more entries for which to obtain details.
         */
        entryNames: (string | CompletionEntryIdentifier)[];
    }

    export interface CompletionEntryIdentifier {
        name: string;
        source: string;
    }
```

initially there was no source field, but they later added it because name alone was not enough to distinguish a entry. When you ask for completion entry details, you have to supply the source also if present, otherwise you will not get any response. We can't just use the new format for everything because 1) we have to support the old tsservers, 2) the source field itself is an optional field and might not present in all CompletionEntry.

I agree it might be bit confusing/hard to understand where the data is coming from. Basically we only get the string back from the company on callbacks. I used the text-property hack to avoid maintaining another hash to store the details of the completion.

```
(let ((name (plist-get completion :name)))
       (put-text-property 0 1 'file-location file-location name)
       (put-text-property 0 1 'completion completion name)
       name)
```

tide.el
ananthakumaran: see the previous comment

ananthakumaran: updated as you suggested

ananthakumaran:
Specification wise, we now get an extra field called `hasAction` in CompletionEntry which signals whether the editor should apply code-fixes on selection of that entry. Implementation wise, I think there are lot of open bugs that need to be sorted out in tsserver. The problem may in tide side or tsserver side, but as we don't get any completion entries at all, I would guess it's on the tsserver side.

tide.el
```diff
@@ -1145,8 +1148,11 @@ Noise can be anything like braces, reserved keywords, etc."
           (tide-annotate-completions (plist-get response :body) prefix file-location)))))))
 
 (defun tide-command:completionEntryDetails (name)
-  (let ((arguments (-concat (get-text-property 0 'file-location name)
-                            `(:entryNames (,name)))))
+  (let* ((source (plist-get (get-text-property 0 'completion name) :source))
+         (arguments (-concat (get-text-property 0 'file-location name)
+                            (if source
+                                `(:entryNames [(:name ,name :source ,source)])
+                              `(:entryNames (,name))))))
```
josteink: So this quirky bit here comes from a quirky `tsserver` contract. Got it.

josteink: So this quirky bit here comes from a quirky `tsserver` contract. Got it.

I still feel this portion could have been cleaned up a little bit more to make it less messy. If we move the `(if source...)` bit into another let-binding, this IMO makes the `arguments` cleaner:

````elisp
(defun tide-command:completionEntryDetails (name)
         (entryNames (if source
                        `(:entryNames [(:name ,name :source ,source)])
                      `(:entryNames (,name))))
         (arguments (-concat (get-text-property 0 'file-location name) entryNames)))
    (-when-let (response (tide-send-command-sync "completionEntryDetails" arguments))
      (when (tide-response-success-p response)
        response))))
````

ananthakumaran: done

josteink:
LGTM. I'm pretty sure I've seen this asked for a few times around.

Unless you feel the need to do any additional work on this, I say it's good to go :)

PR #206, opened 2017-11-11
Prototype for new direct refactorings.

josteink:
This PR is **not** meant to be merged at this point, but just meant to have something concrete to discuss around.

As discussed earlier, `tide-refactor` quickly becomes a annoying stepping-stone which one may want to avoid when you want to activate a known refactoring which you know should be applicable at the current point in the document. 

As such, it would be nice to have some direct commands/functions which you can apply key-bindings to, if you want to trigger this refactorings directly.

This PR contains a very minimal proof of concept for how this can be done. These have been created by inspecting the response from `tsserver` after triggering the `tide-refactor` command.

I realize this code probably needs some additional cleanup, verification and possibly also user-feedback if a refactoring is found not to be applicable at the current location..

That said, does anyone have any comment to the approach chosen here? Is the naming-convention OK? I for instance observe already now that it differs from `tide-rename-symbol` and `tide-fix`.  Does anyone have a better suggestion?

cc: @ananthakumaran @lddubeau

### DIFF
--- a/tide.el
+++ b/tide.el
@@ -1024,6 +1024,19 @@ Noise can be anything like braces, reserved keywords, etc."
           applicable-refactor-infos)))
     (tide-select-item-from-list "Select refactor: " available-refactors #'tide-get-refactor-description (tide-can-use-popup-p 'refactor))))
 
+(defun tide-refactor-class ()
+  (interactive)
+  (tide-apply-refactor '(:action "convert" :refactor "Convert to ES2015 class")))
+
+(defun tide-refactor-extract-function ()
+  (interactive)
+  (tide-apply-refactor '(:action "function_scope_1" :refactor "Extract Symbol")))
+
+(defun tide-refactor-extract-internal-function ()
+  (interactive)
+  (tide-apply-refactor '(:action "function_scope_0" :refactor "Extract Symbol")))
+
+my-test
 (defun tide-apply-refactor (selected)
   (let ((response (tide-command:getEditsForRefactor (plist-get selected :refactor) (plist-get selected :action))))
     (tide-on-response-success response nil


ananthakumaran:
My biggest concern is the keywords we are using for action are not constants, there are just opaque strings according to the protocol. So there are no assurance that it will be the same across different versions and they are free to change it tomorrow.

josteink:
> My biggest concern is the keywords we are using for action are not constants, there are just opaque strings according to the protocol.

That may indeed become an issue. But on the flip-side, Microsoft has historically been world-class on backwards compatibility.

> So there are no assurance that it will be the same across different versions and they are free to change it tomorrow.

Indeed, there are no guarantees, but I do find it hard to believe Microsoft would change these strings for no good reason, seeing as they are just about the only integration-point `tsserver` has for this functionality.

If we were to make functionality like this, what other options would we realistically have? By running the generic `refactor` command, and analysing the user-facing strings in the response instead? I suspect such an approach would be even more risky, as I believe these strings are much more likely to change over time.

If neither option is good.... The most exhaustive approach may be to see what VSCode is doing, and how it's doing it (although VSCode is shipped with a built in `tsserver` so it can make *actual* guarantees about the "constants" used), but I'd rather not take this deep dive, if it can be avoided.

ananthakumaran:
Does vscode provide seperate commands for refactoring? From the release note it seems like a generic one we have now https://code.visualstudio.com/updates/v1_16#_javascript-and-typescript-refactorings https://code.visualstudio.com/updates/v1_18#_extract-local-refactoring-for-javascript-and-typescript

josteink:
> Does vscode provide seperate commands for refactoring? From the release note it seems like the generic one we have now

Looking at it again, I would have to agree.

But still: Is using fixed strings really a big problem here? Is that a show-stopper?

Like I said, Microsoft doesn't really have a history to change stuff like this, and *if* this becomes an issue some way down the road, it should IMO be easy to address at that point.

josteink:
Old issue, not getting worked on. Closing.

PR #207, opened 2017-11-11
add support for jsdoc template command

ananthakumaran:

### DIFF
--- a/tide.el
+++ b/tide.el
@@ -1519,6 +1519,20 @@ code-analysis."
                     (-sort '<))))
     (tide-format-region (-min positions) (-max positions))))
 
+;;; JSDoc Comment Template
+
+(defun tide-command:docCommentTemplate ()
+  (tide-send-command-sync "docCommentTemplate" `(:file ,buffer-file-name :line ,(tide-line-number-at-pos) :offset ,(tide-current-offset))))
+
+(defun tide-jsdoc-template ()
+  "Insert JSDoc comment template at point"
+  (interactive)
+  (let ((response (tide-command:docCommentTemplate)))
+    (tide-on-response-success response nil
+      (save-excursion
+        (tide-insert (tide-plist-get response :body :newText)))
+      (forward-char (tide-plist-get response :body :caretOffset)))))
+
 ;;; Mode
 
 (defvar tide-mode-map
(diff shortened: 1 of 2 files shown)


tide.el
```diff
@@ -1519,6 +1519,20 @@ code-analysis."
                     (-sort '<))))
     (tide-format-region (-min positions) (-max positions))))
 
+;;; JSDoc Comment Template
+
+(defun tide-command:docCommentTemplate ()
+  (tide-send-command-sync "docCommentTemplate" `(:file ,buffer-file-name :line ,(tide-line-number-at-pos) :offset ,(tide-current-offset))))
```
josteink: This bit with ‘:file’ and friends are getting repeated in the *exact same* form quite a bit throughout the code-base now.  Should we extract it into its own function?

tide.el
```diff
@@ -1519,6 +1519,20 @@ code-analysis."
...
+
+(defun tide-jsdoc-template ()
+  "Insert JSDoc comment template at point"
+  (interactive)
+  (let ((response (tide-command:docCommentTemplate)))
+    (tide-on-response-success response nil
+      (save-excursion
+        (tide-insert (tide-plist-get response :body :newText)))
+      (forward-char (tide-plist-get response :body :caretOffset)))))
+
```
josteink: Does this require a minimum tsserver version? If so, should we safeguard against that and inform the user, like we do elsewhere?

tide.el
ananthakumaran: yes some parts are repetitive, but not sure we could abstract it any further. One or two keys changes in most of the request

tide.el
ananthakumaran: It's an old command, seems to be there from v2.0.5 https://github.com/Microsoft/TypeScript/commit/cd1efd5109fb37b6f345fa6a675941182d40c410

tide.el
josteink: But these 3 exact keys are used quite a lot though. I guess we can clean this up later too. 

Was just an observation.

tide.el
josteink: If so, I say let’s just keep it as is.

tide.el
ananthakumaran: I will think about it, if needed probably do it in a different PR as it will touch too many places

PR #213, opened 2017-12-01
document defcustom in readme

ananthakumaran:


josteink:
This is great stuff. I fully support initiatives like this. One thing I suspect now though is that will be possible to create a `defcustom` which won't create a clean `README`-entry if you omit the doc-string.

And if we're *sloppy* during review or coding (you know, it happens)... Then the next `make readme` might make a bad looking `README`... And if we're double sloppy we won't notice until later. In best case, if we do notice, we'll still have a job cleaning up whatever caused it.

It would be great if we could avoid those kind of situations entirely by having `make readme` error out on incompletely declared `defcustom`-declaration, and add `make readme` to the CI build-definition.

That way we will can be sure that we always will be able to cleanly update the `README`-file.

Or is that over-engineering? What do you think?

ananthakumaran:
I added a check for missing documentation case. As we are going to run it by hand, I guess we would catch any missing docs just by looking at the diff. Adding it to the CI is probably overkill, we could revisit the decision if we did commit anything bad in the master.

PR #215, opened 2017-12-01
defcustom to turn off sorting completions by kind

### LINKED ISSUE #214: Tide shouldn't sort completions by kind
Typescript and Javascript don't distinguish between function and constant declarations, so sorting by kind separates things that are semantically similar. Plus, non-alphabetical sorting just messes up completion. Here's an example from the Typescript codebase itself:

```ts
namespace Debug {
  export function assert(condition: boolean, message?: string) { ... };
  export var assertEachNode = shouldAssert(AssertionLevel.Normal) ?
    (nodes: Node[], test: (node: Node) => boolean): void { ... } :
    () => { };
}
Debug.asser/**/
```

Requesting completion at /**/, by typing ( in my case, completes to assertEachNode, when it should complete to assert.

I'll send out a PR when I have time that at least adds custom variables to return the behaviour to the intended defaults. However, I'd prefer just to remove the sorting code entirely.

sandersn:
Fixes #214 by adding a custom variable. Ideally the sorting would just be removed, but some people may prefer to have it on.

Note that the variable defaults to `nil`, however.

### DIFF
--- a/tide-tests.el
+++ b/tide-tests.el
@@ -32,7 +32,7 @@
   (should (equal "this\nis\nfour\nlines"
                  (tide-normalize-lineshift "this\nis\r\nfour\nlines"))))
 
-(ert-deftest completions-get-sorted-by-kind ()
+(ert-deftest completions-get-sorted ()
   "Tests that completion candidates get sorted by kind."
   (let ((mock-completions
          '(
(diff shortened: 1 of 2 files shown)


josteink:
Hey there, and thanks for another patch! :)

I can see how this is something people might have different opinions about, so I welcome the initiative.

That said... While we used to have *one* behaviour (which at least was consistent), right now you're making that into two:

* sort by kind
* don't sort at all (which may or may not provide consistent results)

As a preference... I'm not sure I like that. Wouldn't it be better to have *different* sorting options instead?

* no sort
* sort by kind
* sort by name
* sort by Case
* etc.

As such, I think a `defcustom` which would make more sense IMO would involve choosing a ranking-method, not whether we should do sorting at all.

If we did this, and also implemented some basic ranking-mechanisms, people would also be able to compose their own custom sorting quite easily in a way which works nicely with the `defcustom` system... (Like sort by Case, then name, thenk kind).

As I can see it, adapting your patch to fit such a scheme shouldn't really involve much work. You just need to provide another option (like no sort or name-sort).

What do you think?

sandersn:
I'm biased since I work on Typescript, but I like the default order from Typescript, which is basically just a case-insensitive sort of names. So I change to have three choices for `tide-sort-completions`: `'default`, `'case` and `'kind`.

I do think that ranked sorting is not likely to be useful to many people, especially with only two non-mutually-exclusive sorts available.

ananthakumaran:
My opinion is, tide should act as thin emacs UI layer over tsserver. It shouldn't try to do much.

Most of the time when we try to make it better, like introducting filter/sort for completions, we inadvertently break something else. This happened with filtering and we had to remove it later.

At this point, I guess we should make tsserver results as default and for backward compatability, we could make by-kind available as an option that could be enabled by user. I am -1 on adding any new kind of sorting unless explicitly asked by users.

josteink:
Thanks for the comment. And I guess that’s fair enough. 

It was a suggestion/thought on my part, but if people don’t agree with it I don’t really have super-strong feelings either way.

sandersn:
Sounds good to me. I reverted the additional options.

josteink:
I guess in that case your effective change:

* keeps the sorting code in place, mostly as is
* but guards it behind a new defcustom
* a defcustom which by default is off, effectively reverting 

Unless anyone else has any issues with that, I'm happy to merge.

josteink:
LGTM. Consider it merged.

Thanks for the feedback and thanks for patching it up! :)

PR #218, opened 2017-12-04
Skip first layer when building a hierarchical imenu index

### LINKED ISSUE #217: Skip one-level of imenu
Is there a reason why `tide-imenu-index` does not skip the first level of the navtree when `tide-imenu-flatten` is nil? I like having the hierarchical imenu, but the first layer is always just `*Rescan*` and the filename.

felipeochoa:
Fixes #217

I'm not familiar with the `tsserver` methods, but it seems like the navtree hierarchy always has a single element as its root.

### DIFF
--- a/tide.el
+++ b/tide.el
@@ -1396,10 +1396,10 @@ number."
 (defun tide-imenu-index ()
   (let ((response (tide-command:navbar)))
     (tide-on-response-success response nil
-      (let ((navtree (plist-get response :body)))
+      (let ((children (tide-plist-get response :body :childItems)))
         (if tide-imenu-flatten
-            (-flatten (-map #'tide-build-flat-imenu-index (plist-get navtree :childItems)))
-          (list (tide-build-imenu-index navtree)))))))
+            (-flatten (-map #'tide-build-flat-imenu-index children))
+          (mapcar #'tide-build-imenu-index children))))))
 
 ;;; Rename
 


josteink:
If I were to play devils advocate... I guess the first level menu is useful in a ‘imenu-anywhere’ type scenario. 

But on the flip side, with tsserver-based code analysis and navigation available, I find it really hard to believe anyone actually relies on that in their flow. 

So I’m mostly ok with this move. Anyone else have any objections?

ananthakumaran:
@felipeochoa The structure of tsserver response can be found in here https://github.com/Microsoft/TypeScript/blob/master/src/server/protocol.ts#L2456

The top level has only one element
```
{"text":"\"circle\"","kind":"module","kindModifiers":"","spans":[{"start":{"line":1,"offset":1},"end":{"line":82,"offset":39}}],"childItems":[{"text":"Circle","kind":"class","kindModifiers"
```
So, the question is whether the text `circle` is useful or not? We already throw it away for flat index, I guess for imenu we could do the same because it just seem to be either namespace or `<global>` incase of js files..

felipeochoa:
Ok, thanks for clarifying. I think in that case, then yes, the top-level text is pretty useless. I haven't encountered such a top-level item in any other imenu function, so I imagine even something like `imenu-anywhere` (which I've never used) can handle that situation just fine.

josteink:
> so I imagine even something like imenu-anywhere (which I've never used) can handle that situation just fine.

`imenu-anywhere` will *handle* anything. I'm talking about how it may be used by users :)

Basically `imenu-anywhere` lets you jump to any imenu item in any open buffer. Combined with `helm` that gives you a fuzzy search for anything you currently have open, and *if* we keep the filename in the `imenu`-index, that gives the user the possibility to also use filename as a quick and easy criterion for result-filtering and navigation.

That said... That's the only possible use *I* can see for it, and I really think that with a tsserver backend, this is too niche a use-case for us to hold back other changes.

I just wanted to hear if anyone else could see any possible issues with this change. But if people in general are happy, I'm happy :)

tide.el
```diff
@@ -1396,7 +1396,8 @@ number."
       (let ((navtree (plist-get response :body)))
         (if tide-imenu-flatten
             (-flatten (-map #'tide-build-flat-imenu-index (plist-get navtree :childItems)))
-          (list (tide-build-imenu-index navtree)))))))
+          (mapcar #'tide-build-imenu-index
+                  (plist-get navtree :childItems)))))))
```
josteink: Right now we have a `let` block on the top there where `navtree` is extracted through `plist-get`.

I also see that we have *two* identical instances of `(plist-get navtree :childItems)` contained within that tiny `let`-block.

I think we should instead use a `(let*)`-block and extract both values there , or just extract `childItems` directly for less redundant code.

felipeochoa:
Just pushed the changes (I squashed the commmits)

tide.el
```diff
@@ -1396,10 +1396,11 @@ number."
 (defun tide-imenu-index ()
   (let ((response (tide-command:navbar)))
     (tide-on-response-success response nil
-      (let ((navtree (plist-get response :body)))
+      (let* ((navtree (plist-get response :body))
+             (children (plist-get navtree :childItems)))
```
ananthakumaran: as navtree is not referred, you could just use tide-plist-get which supports nested get

felipeochoa:
Ok updated again :)

josteink:
Looks good to me.

PR #219, opened 2017-12-04
Derive project-errors and references modes from special-mode

### LINKED ISSUE #216: Derive tide-project-errors and tide-references from special-mode
From the [info manual](https://www.gnu.org/software/emacs/manual/html_node/elisp/Derived-Modes.html):

> The recommended way to define a new major mode is to derive it from an existing one 
> using define-derived-mode. If there is no closely related mode, you should inherit from
> either `text-mode`, `special-mode`, or `prog-mode`. See [Basic Major Modes][1]. If none of
> these are suitable, you can inherit from `fundamental-mode` (see [Major Modes][2]). 

I think `tide-references-mode` and `tide-project-errors-mode` should inherit from `special-mode`.

This lets any tools/customizations which are activated only special modes to do this automatically without having to special-case these modes.

[1]: https://www.gnu.org/software/emacs/manual/html_node/elisp/Basic-Major-Modes.html#Basic-Major-Modes
[2]: https://www.gnu.org/software/emacs/manual/html_node/elisp/Major-Modes.html#Major-Modes

felipeochoa:
Closes #216 

I took out the read-only declaration, the `quit-window` binding, and the `use-local-map` call. The first two are definitely taken care of by special-mode; the last one I think all minor modes do. I tested the changes on 25.2 and everything worked the same

### DIFF
--- a/tide.el
+++ b/tide.el
@@ -1261,15 +1261,12 @@ Noise can be anything like braces, reserved keywords, etc."
     (define-key map (kbd "p") #'tide-find-previous-reference)
     (define-key map (kbd "C-m") #'tide-goto-reference)
     (define-key map [mouse-1] #'tide-goto-reference)
-    (define-key map (kbd "q") #'quit-window)
     map))
 
-(define-derived-mode tide-references-mode nil "tide-references"
+(define-derived-mode tide-references-mode special-mode "tide-references"
   "Major mode for tide references list.
 
 \\{tide-references-mode-map}"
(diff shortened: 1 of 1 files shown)


josteink:
Awesome. Thanks for another patch!

I’ll try to look over these PRs and get them reviewed over the next few days.

felipeochoa:
Great, thanks! Let me know if you have more questions/need any changes

josteink:
LGTM. Thanks for the patch!

PR #221, opened 2017-12-09
enable auto import suggestion by default

### LINKED ISSUE #220: Autocomplete can't find names from other files
And it affects also auto-impor feature.
Probably it's because of argument "includeExternalModuleExports", recently added to "completions" command in tsserver.

```
    /**
     * Arguments for completions messages.
     */
    interface CompletionsRequestArgs extends FileLocationRequestArgs {
        /**
         * Optional prefix to apply to possible completions.
         */
        prefix?: string;
        /**
         * If enabled, TypeScript will search through all external modules' exports and add them to the completions list.
         * This affects lone identifier completions but not completions on the right hand side of `obj.`.
         */
        includeExternalModuleExports: boolean;
}
```

ananthakumaran:
fixes #220

### DIFF
--- a/README.md
+++ b/README.md
@@ -224,6 +224,10 @@ Name or path of the node executable binary file.
 
 Hook run after code edits are applied in a buffer.
 
+**tide-sort-completions-by-kind** `nil`
+
+Whether completions should be sorted by kind.
+
 **tide-format-options** `'nil`
 
 Format options plist.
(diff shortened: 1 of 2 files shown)


josteink:
LGTM. I like the new auto documentation generator :)

domachine:
Is there anything I can do to help to get this out as a release to MELPA?

josteink:
Do you mean MELPA *stable*? This should have been in "regular MELPA" the day after it was merged.

domachine:
Ah, okay my fault, I'm already on the latest version, sorry. Unfortunately I still have no auto-import suggestions :-( but I think then it's a configuration problem on my side.

josteink:
It's probably tied to the version of TypeScript and tsserver your project uses. Try upgrading your projects Typescript-package to latest and see if that helps.

domachine:
That is so true! Thank you very, very much!

PR #223, opened 2017-12-14
Fix tide-command:getCodeFixes failing with getting file name

normano64:
I wasn't able to use `tide-fix` with latest master, this PR fixes it but I have no knowledge in eslip/lisp so I have no idea what I'm doing 🤔


josteink:

That looks like a bug introduced with our latest future, support for indirect buffers. 

Your fix looks nice and clean.  Thanks for finding, fixing and submitting!

PR #228, opened 2018-01-11
Added information about integration with use-package

dunmaksim:
Configuration for use-package from [here](https://github.com/frgomes/.emacs.d)

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,6 +42,47 @@
 (add-hook 'typescript-mode-hook #'setup-tide-mode)
 ```
 
+#### Integration with use-package
+```elisp
+(use-package tide
+  :commands tide-setup tide-mode setup-tide-mode
+  :config
+  (defun setup-tide-mode ()
+    (interactive)
+    (tide-setup)
+    (tide-mode +1)
+    ;; Flycheck, eldoc and company are optional
+    (flycheck-mode +1)
+    (setq-default tab-width 2)
+    (setq flycheck-check-syntax-automatically '(save mode-enabled))
+    (eldoc-mode +1)
+    (tide-hl-identifier-mode +1)
+    (company-mode +1))
+  (require 'typescript-mode)
+  (add-hook 'typescript-mode-hook #'setup-tide-mode)
+
+  ;; aligns annotation to the right hand side
+  (setq company-tooltip-align-annotations t)
+  ;; format options
+  (setq tide-format-options '(:insertSpaceAfterFunctionKeywordForAnonymousFunctions t :placeOpenBraceOnNewLineForFunctions nil))
+  ;; support for JS files
+  (add-hook 'js2-mode-hook #'setup-tide-mode)
+  ;; from now on web-mode is required
+  (require 'web-mode)
+  ;; support for TSX files
+  (add-to-list 'auto-mode-alist '("\\.tsx\\'" . web-mode))
+  (add-hook 'web-mode-hook
+            (lambda ()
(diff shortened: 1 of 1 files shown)


ananthakumaran:
is it possible to avoid the repetition of all the config described earlier? Probably like a template which says where to place all the config discussed earlier for use-package users?

aappddeevv:
I noticed that this does not set tide-mode for .ts files or .js files just tsx and jsx. Is that intentional?

dunmaksim:
No. I copy this code from Emacs config from GitHub. Also I have problem 
with TypeScript server initialization on Windows. Do I think, you must 
decline my pull request. I don't know simple solution of these 2 problems.


aappddeevv <notifications@github.com> 13 января 2018 г. 6:29:41 ПП написал:

jojojames:
Here's what I have for tide (semi-generalized) that should work a little better than what's in the link. (e.g. eliminating excessive requires, supporting javascript/webmode, handling tsfmt/typescript definition files, etc)

```
(use-package tide
  :ensure t
  :commands (tide-setup)
  :init
  (defun +tide-setup ()
    (interactive)
    (when (locate-dominating-file default-directory "tsfmt.json")
      (add-hook 'before-save-hook #'tide-format-before-save nil t))
    ;; Disable `flycheck' linting for Typescript Definition files.
    (when (and (buffer-file-name)
               (string-match-p ".d.ts$" (buffer-file-name)))
      (when (fboundp 'flycheck-mode)
        (flycheck-mode -1)))
    (tide-setup)
    (tide-hl-identifier-mode +1))
  (add-hook 'typescript-mode-hook #'+tide-setup)

  (add-hook 'js2-mode-hook
            (lambda ()
              (when (or
                     (locate-dominating-file default-directory "tsconfig.json")
                     (locate-dominating-file default-directory "jsconfig.json"))
                (+tide-setup))))

  (add-hook 'web-mode-hook
            (lambda ()
              ;; Set up Tide mode if Typescript.
              (when (string-equal (file-name-extension buffer-file-name) "tsx")
                (setq-local web-mode-enable-auto-quoting nil)
                (when (fboundp 'yas-activate-extra-mode)
                  (yas-activate-extra-mode 'typescript-mode))
                (+tide-setup))))
  :config
  ;; Set up Typescript linting with `web-mode'.
  (with-eval-after-load 'flycheck
    (flycheck-add-mode 'typescript-tslint 'web-mode)))
```

ananthakumaran:
I would like to avoid duplicate the config code in the readme. Either we need to come up with a way to avoid duplication or add this as a wiki page.

josteink:
It seems to me like setting up `tide` completely  with `use-package` involves a pretty elaborate setup. Same with JSX/TSX (and `web-mode`).

How about we move these kind of configurations out of the `README` and into a new file (`CONFIGURATION.md`?) where we can be more verbose about the different setups without causing the readme itself to be overly repetitive?

dunmaksim:
Do I think wiki is better.


Jostein Kjønigsen <notifications@github.com> 29 января 2018 г. 11:30:20 ДП 
написал:

josteink:
Yeah a wiki is good too. We already [have one](https://github.com/ananthakumaran/tide/wiki), so the barrier to entry should be fairly low.

@ananthakumaran: Are you OK with that idea?

ananthakumaran:
I am OK with wiki, suggested the same thing earlier.

josteink:
Right. In that case I suggest we just close this PR and @dunmaksim moves on to create some pages in the wiki :)

PR #230, opened 2018-01-15
respect extends in tsconfig.json file

### LINKED ISSUE #225: compileOnSave isn't inherited via extends
I believe `compileOnSave` should be inherited and possibly overridden with `extends`.

**src/tsconfig.base.json:**
```json
{
  "compileOnSave": true
}
```

**src/project1/tsconfig.json:**
```json
{
  "extends": "../tsconfig.base"
}
```

**src/project2/tsconfig.json:**
```json
{
  "extends": "../tsconfig.base",
  "compileOnSave": false
}
```

**src/project3/tsconfig.json:**
```json
{
  "extends": "../tsconfig.base",
  "compileOnSave": true
}
```

Given the above configuration, the *project1* and *project3* sources should compile on save, and the *project2* sources should not. Currently only *project3* respects `compileOnSave` because it is explicitly configured and `extends` is not respected.

ananthakumaran:
fixes #225 

@sbuzonas please test this branch

### DIFF
--- a/tide.el
+++ b/tide.el
@@ -2036,6 +2036,20 @@ timeout."
      (when (and config (eq (plist-get config :compileOnSave) t))
        (tide-command:compileOnSaveEmitFile)))))
 
+(defun tide-load-tsconfig (path loaded-paths)
+  (when (member path loaded-paths)
+    (error "tsconfig file has cyclic dependency, config file at %S is already loaded." path))
+  (when (not (file-exists-p path))
+    (error "tsconfig file not found at %S." path))
+  (let ((config (tide-safe-json-read-file path)))
+    (if-let (extends (plist-get config :extends))
+        (let ((base (tide-load-tsconfig (expand-file-name extends (file-name-directory path)) (cons path loaded-paths))))
+          (tide-combine-plists
+           base
+           config
+           `(:compilerOptions ,(tide-combine-plists (plist-get base :compilerOptions) (plist-get config :compilerOptions)))))
+      config)))
+
 (defun tide-project-config (cb)
   (let ((config (gethash (tide-project-name) tide-project-configs :not-loaded)))
     (if (eq config :not-loaded)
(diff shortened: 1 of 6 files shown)


tide.el
```diff
@@ -2036,14 +2036,28 @@ timeout."
      (when (and config (eq (plist-get config :compileOnSave) t))
        (tide-command:compileOnSaveEmitFile)))))
 
+(defun tide-load-tsconfig (path loaded-paths)
```
josteink: IMO this function looks like something which would be easier to test with unit-tests than manual testing.

tide.el
```diff
@@ -2036,14 +2036,28 @@ timeout."
...
+  (when (member path loaded-paths)
+    (error "tsconfig file has cyclic dependency, config file at %S is already loaded." path))
```
josteink: This error-condition should also be fairly easy to verify in a unit-test.

josteink:
Apart from my regular comments about automated tests, I think the rest looks like a good improvement.

Thumbs up.

josteink:
And that makes it even better :)

josteink:
I did some rudimentary testing based on the build-output. One reason the test fails seems to be that the *order* of the elements in the alist (and contained alists) matters:

````elisp
;; failing check in build output
'(equal
 '(:compileOnSave t :compilerOptions
                 (:target "ES7" :sourceMap t)
                 :extends "./base.json")
 '(:compilerOptions
  (:sourceMap t :target "ES7")
  :compileOnSave t :extends "./base.json"))
;; nil

;; same case-reformatted for easier comparison
(equal
 '(:compileOnSave t :compilerOptions (:target "ES7" :sourceMap t) :extends "./base.json")
 '(:compilerOptions (:sourceMap t :target "ES7") :compileOnSave t :extends "./base.json"))
;; nil

;; outer alist-keys re-ordered
(equal
 '(:compileOnSave t :compilerOptions (:target "ES7" :sourceMap t) :extends "./base.json")
 '(:compileOnSave t :compilerOptions (:sourceMap t :target "ES7") :extends "./base.json"))
;; nil

;; all alist-keys reordered to match, also in contained alists
(equal
 '(:compileOnSave t :compilerOptions (:target "ES7" :sourceMap t) :extends "./base.json")
 '(:compileOnSave t :compilerOptions (:target "ES7" :sourceMap t) :extends "./base.json"))
;; t
````

I'm guessing the `equal` implementation across Emacs-versions has different behaviours in this regard.

ananthakumaran:
Yes, looks like we need to implement `plist-equal` or something similar

josteink:
We can be lazy and disable the tests for older Emacs versions if we don't feel this slight piece of indirection is worth the effort right now.

But if we're not, at least that would be a function which would be useful for other, future test-cases where we need to somehow assert on JSON-based values.

josteink:
When you put it like that... Who am I to object? 💃 

Totally approved!

PR #234, opened 2018-01-28
add support for fix all in file & project-info

ananthakumaran:
Related to https://github.com/Microsoft/TypeScript/pull/20338. Need typescript@next for testing.

### DIFF
--- a/tide.el
+++ b/tide.el
@@ -1018,6 +1018,9 @@ Noise can be anything like braces, reserved keywords, etc."
 (defun tide-command:getCodeFixes ()
   (tide-send-command-sync "getCodeFixes" `(:file ,(tide-buffer-file-name) :startLine ,(tide-line-number-at-pos) :startOffset ,(tide-current-offset) :endLine ,(tide-line-number-at-pos) :endOffset ,(+ 1 (tide-current-offset)) :errorCodes ,(tide-get-flycheck-errors-ids-at-point))))
 
+(defun tide-command:getCombinedCodeFix (fixId)
+  (tide-send-command-sync "getCombinedCodeFix" `(:scope (:type "file" :args (:file ,(tide-buffer-file-name))) :fixId ,fixId)))
+
 (defun tide-get-fix-description (fix)
   (plist-get fix :description))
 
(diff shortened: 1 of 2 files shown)


josteink:
It's not clear from the description, but if I were to test this... Would the simplest way to do so, would be with a TS-file with multiple unused imorts (as per https://github.com/Microsoft/TypeScript/issues/14549)?

What other kinds of use-cases would be useful to verify?

tide.el
```diff
@@ -2072,6 +2091,34 @@ timeout."
...
+      (insert "tsserver version: ")
+      (insert (propertize version 'face '(success bold)))
+      (insert "\n\n")
+      (insert "config file path: ")
+      (insert (propertize config-file-name 'face 'success)))
+    (special-mode)
+    (display-buffer (current-buffer) t)))
+
+(defun tide-project-info ()
+  "Show the version of tsserver."
```
josteink: I think this name is (currently) pretty misleading. It doesn't show much info about the *project* at all. What did you intend for this function to be used for? I think a naming based on that would work better.

For instance, if this is meant to aid diagnostics, we should call it `tide-show-diagnostics` or something similar.

tide.el
```diff
@@ -1018,29 +1018,48 @@ Noise can be anything like braces, reserved keywords, etc."
...
-(defun tide-apply-codefixes (fixes)
+(defun tide-apply-codefix-for-all-in-file (fix)
+  (tide-apply-codefix fix)
+  (-when-let* ((fix-id (plist-get fix :fixId))
+               (response (tide-command:getCombinedCodeFix fix-id)))
+    (tide-on-response-success response nil
+      (tide-apply-codefix (plist-get response :body)))))
+
+(defun tide-apply-codefixes (fixes apply-codefix)
   (cond ((= 0 (length fixes)) (message "No code-fixes available."))
```
josteink: I think variable name `apply-codefix` is not very good. To me that name almost seems to imply this is a boolean, and IMO is a source of needless confusion or mental overhead.

`codefix-function` or something which hints at its type would be better.

tide.el
```diff
@@ -1018,29 +1018,48 @@ Noise can be anything like braces, reserved keywords, etc."
...
+        (t (funcall
+            apply-codefix
             (tide-select-item-from-list "Select fix: " fixes #'tide-get-fix-description (tide-can-use-popup-p 'code-fix))))))
 
-
-(defun tide-fix ()
-  "Apply code fix for the error at point."
-  (interactive)
+(defun tide-code-fix (apply-codefix)
   (unless (tide-get-flycheck-errors-ids-at-point)
```
josteink: Same with the variable name `apply-codefix` here.

ananthakumaran:
These are the couple of cases I have tested. 

```javascript
// @ts-check

class C {
    constructor() {

    }
    method() {
      var x = this.x;
      x += this.y();
      x += this.x;
      return x;
    }
}
```

```typescript
export interface Shape {
    getArea(): number;
}

export class Circle implements Shape {
}

class Rectangle implements Shape {
}

class Square implements Shape {
}
```

tide.el
ananthakumaran: Yeah, it's mostly for debugging. `project-info` may be inaccurate. What about `tide-verify-setup`?. I am not sure about `show-diagnostics`

tide.el
ananthakumaran: will do

tide.el
josteink: I'm OK with that. That's still much better than `tide-project-info`.

josteink:
Not a big difference in terms of code, but I still think that makes it much more readable and apparent what things are about :)

ananthakumaran:
have you had a chance to test this out?

josteink:
Sorry, but I simply haven't had time for that. Busy as heck at work.

If it works for the use-cases you have intended, I say merge ahead. I have another few patches on another few projects which will probably get my attention first (which are also pending a proper review).

Issue #71, opened 2016-10-27
tide + tramp


skormel:
hi all,

Congrantz for this great module it works fantastic.

Only I have an issue.  My angular project is not in local, I access it via TRAMP (SSH), the code is decoracted, but I can't use any of the tide-commands, It always return:

> eldoc error: (Server does not exist. Run M-x tide-restart-server to start it again)

If I run the proposed command. It dumps this:

> (client) tsserver exists: exited abnormally with code 1.

Because tsserver must be running in local machine, but not in remote.

In remote machine I've got emacs + tide configured and 100% working.

Exists any way to fix this situation?

Thanks.

Best Regards.
Antonio

ananthakumaran:
I myself don't use tramp and don't have time now to debug the issue. tide uses `start-file-process` and default directory is set to the return value of `(tide-project-root)`. It looks like `tide-project-root` needs to be modified to point to remote url https://www.gnu.org/software/emacs/manual/html_node/tramp/Remote-processes.html

skormel:
I'll try to solve it but I'm completely noob with lisp/elisp (I know,  I've got to study lisp). 

If I could fix it, I'd create a pull request.

Thanks for all

Best Regards.
Antonio.

FieryCod:
Is there any progress in making tide work correctly with tramp?

josteink:
Not that I know. Whoever wants it, will have to send in a patch :)

PR #235, opened 2018-02-01
Support remote file access

cuspymd:
If `tide-project-root` is on remote host, `tsserver` runs on remote host by tramp-mode.
So when `tide` passes filename to `tsserver`, filename should be changed as local file format.
And when `tide` receives filename to `tsserver`, it should be changed as full(tramp) file format.
  
* Related issue #71

### DIFF
--- a/tide.el
+++ b/tide.el
@@ -37,6 +37,7 @@
 (require 'flycheck)
 (require 'imenu)
 (require 'thingatpt)
+(require 'tramp)
 (require 'tide-lv)
 
 ;; Silence compiler warnings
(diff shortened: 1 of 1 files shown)


ananthakumaran:
Thanks for the PR. I haven't gone through the details, but I have couple of comments about the general approach

1) the way I see it, there are three primitive functions
```elisp
(defun tide-local2remote (file))
(defun tide-remote2local (file))
(defun tide-plist-transform (plist fn &rest keys))
```
The code seems to be mixing up these three concerns(specially the transforms). We could just write a general data transform function and use it everywhere instead of using plist/map etc in a adhoc way. If you want to transform something inside list we could do something like

```elisp
(tide-plist-transform response #'tide-local2remote :body :refs #'-map :file)

(defun tide-plist-transform (plist fn &rest keys)
  ...
  (let ((key (car keys)))
    (if (functionp key)
        (funcall key (lambda (plist) (tide-list-transform plist (rest keys))) plist)
      ...)))
```

2) if possible, we should limit the changes to two places.

if tramp mode is active (we shouldn't do data transform for non tramp users, because it could cause performance problem. Sometimes we get more than 100 KB data from tsserver)

a) transform request before sending data to tsserver
b) transform response after receiving data from tsserver

we already seem to do a, probably we should do b also. The structure of data returned by tsserver is well documented at https://github.com/Microsoft/TypeScript/blob/master/src/server/protocol.ts. With the name of the command, this would be straight forward to write I believe.

cuspymd:
@ananthakumaran Thanks for your review.
I agree with your opinion about a general data transform function. But the function `tide-plist-transform` is not easy to understand.

Frist parameter `plist`, it looks like a propery list but can be a propery list or a normal list.
Third parameter `keys`, it looks like a list of poperty symbols but can contains a symbol or a function.
And any function can be passed to the parameter `keys`, but in function implementations it is assumed to be the function `-map`.

I can not be sure that this function is well designed. Do you think this is a really good design to stick with at the cost of readability?

And there is something to be said about your performance concern.
If `tide-project-root` exits on local PC(in case of  'noon tramp users' you said),
this patch dose string search once for only the path of `tide-project-root` regardless of the size of the received file.

Thank you for your review. I'll take a look at the typescrpt protocol.

ananthakumaran:
The `tide-plist-transform` function I had in mind was somewhat similar to `update_in` in elixir language https://hexdocs.pm/elixir/master/Access.html#module-accessors. Our implementation need not be thorough and for simplicity, we can for now only implement the `-map` type functions, nothing else needs to work (I am also ok with using a special symbol like `:*` to specify list traversal). 

> I can not be sure that this function is well designed. Do you think this is a really good design to stick with at the cost of readability?

Maybe, but that expresses the intent of the code clearly, though there is implementation complexity at the transform function. But, as it will be used many times, probably the added complexity is ok. My view may also be affected by my familiarity with elixir's `update_in` function.

cuspymd:
@ananthakumaran Thanks for your reply.
I will refer to your link.

josteink:
Anyone have the status on this one? Is this PR still good? Are there outstanding work?

It seems such a shame to me to not to be able to fully support tramp, if this is all the work that is needed.

EdwardIII:
Hey @cuspymd how did you get on with this? Anything in your PR I can help with? Would love to use tide over tramp.

cuspymd:
@EdwardIII Thanks for your interest.
I forgot about this PR for a long time. First of all, it's been a while, so I guess I'll have to check if these fixes are still valid.

PR #237, opened 2018-02-04
verify filename before applying highlights

### LINKED ISSUE #236: tide-hl-identifier-mode broken due to tsserver 2.7 apparent bug
tsserver since 2.7 is no longer filtering on filesToSearch while performing the documentHighlights request.  This often results in tide attempting to highlight a bad location in the current buffer as it's assuming that the response's first file is the requested one.  This is now sometimes not the case.

I believe this is a Typescript bug, but it could be worked around in tide by checking that the returned filename is in fact the requested buffer.

Typescript bug filed at Microsoft/TypeScript#21604.

ananthakumaran:
fixes #236 

@bdowning I am not able to reproduce it locally with version `2.8.0-dev.20180127`. But I guess this should fix/workaround the bug. Could it test it out?

### DIFF
--- a/tide.el
+++ b/tide.el
@@ -1960,7 +1960,10 @@ highlights from previously highlighted identifier."
 
 (defun tide--hl-highlight (response)
   "Highlight all instances of the identifier under point."
-  (let ((references (plist-get (car (plist-get response :body)) :highlightSpans)))
+  (-when-let* ((item (-first (lambda (item)
+                               (equal (tide-buffer-file-name) (plist-get item :file)))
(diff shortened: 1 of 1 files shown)


josteink:
LGTM.

bdowning:
This does appear to fix things for me at first glance.  I posted the test case I used over in #236 if you want to verify.

ananthakumaran:
Verfied, tsserver doesn't send us any highlights for the requested file in this case, so just not throwing errors is the only thing that could be done here I guess.

PR #240, opened 2018-02-08
add support for completion with inserText

ananthakumaran:
add support for the two cases mentioned in [1](https://code.visualstudio.com/updates/v1_20#_bracket-property-suggestions) and [2](https://code.visualstudio.com/updates/v1_20#_automatic-member-property-suggestions)


josteink:
Tested and works. Awesome feature. Good job :)

PR #241, opened 2018-02-24
add support for organizeImports command

ananthakumaran:
this command is only available on Typescript master branch

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,6 +157,8 @@ point.
 
 <kbd>M-x tide-verify-setup</kbd> Show the version of tsserver.
 
+<kbd>M-x tide-organize-imports</kbd> Organize imports in the file.
+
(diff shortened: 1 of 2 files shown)


josteink:
Woo! Another feature!

Based on what I [read here](https://github.com/Microsoft/TypeScript/pull/21909), it seems all this should do is sort the imports. I've tested that, and it seems to work for me.

Besides that, the code changes looks good to me. I especially like the new way of handling compatibility.

I'd preferred it to be split over 2 commits (refactor, add feature), but now that it's already done, I'm not sure it's worth the trouble splitting up.

I'm OK with this getting merged as is. Good job :)

ananthakumaran:
Thanks, I kind of got lazy and did the refactor in the middle.

Issue #246, opened 2018-04-07
We need to provide more guidance for issue reporting.


lddubeau:
We've been having a number of issue reports lately that fell in these categories:

1. Issues reporting bugs that did not include the minimal information that should be present. In my view, common developer courtesy requires that issue reports include at a minimum: the relevant version numbers, a series of steps to reproduce the issue, actual results, expected results. I strive to always include this in my all my bug reports.

2. Problems that might go away by performing a series of minimal diagnostic steps. Or might be traceable to other tools than tide itself through diagnostic steps.

I see two things necessary:

1. A template for submitting issues to this project. (`ISSUE_TEMPLATE.md`). The template will contain a checklist of diagnostic steps to perform before submitting the issue, but I don't want to load that list with lots of details, so this brings up the next point...

2. A wiki page that explains the rationale for the diagnostic steps in greater details. I want to provide those details separately so that people who already know how it all works can just go through the checklist. People who don't already know can follow a link and read why, for instance, we're asking to issue `tide-restart-server` before deciding that there's a bug in tide.

I see the README has a [debugging](https://github.com/ananthakumaran/tide#debugging) section. I'm thinking that kind of debugging should not be part of the initial checklist but instead be brought up by us on a case by case basis.

I'm going to create the wiki page and the `ISSUE_TEMPLATE.md`, and I'm going to assign myself to this issue, which only means I'm not expecting anyone else to take the lead. Comments, concerns, suggestions are all welcome.

josteink:
Great initiative! I agree about the problem and the general framework you’ve proposed as a decent improvement over what we have now. 

If we see that it doesn’t work out, we can tweak it along the way.

PR #247, opened 2018-04-07
Guidance for filing issues.

lddubeau:
This is the PR for the issue template i mentioned in #246.

There are two areas I'm still unsure about:

1. There's a step in the checklist asking to reproduce the problem with the latest version of tide. I'm not sure what specific version we should consider to be "latest" for our purposes here. I know some projects ask submitters to test against the master branch, which in some cases I consider to be an excessive request because checking out the project's master branch and building it to check whether the issue still happens there is a major enterprise. Ever since I've started contributing to tide I've been running tide from the master branch anyway so I've lost track of how often releases are pushed to the package repository (or repositories) and therefore I don't have a clear idea of the stakes here.

2. In the wiki, the section titled ["The Problem does not Belong Elsewhere"](https://github.com/ananthakumaran/tide/wiki/Explanations-about-the-Issue-Reporting-Checklist#the-problem-does-not-belong-elsewhere) can probably benefit from some more rules of thumb. Ultimately, I would not be surprised if we don't get great benefit from that section. If users get wrong results, it could be that `tsserver` is not doing the right thing, or tide is not applying the results correctly. It is hard to know which case is the right one without knowing the internals of tide or `tsserver`.

### DIFF
--- a/ISSUE_TEMPLATE.md
+++ b/ISSUE_TEMPLATE.md
@@ -1,0 +1,31 @@
+When submitting issues to this project, please include the following information.
+
+## Checklist
+
+- [ ] I have searched both open and closed issues and cannot find a duplicate.
+- [ ] I can [reproduce the problem](https://github.com/ananthakumaran/tide/wiki/Explanations-about-the-Issue-Reporting-Checklist#use-the-latest-version-of-the-relevant-packages) with the latest version of the relevant packages.
+- [ ] The problem still occurs after I issued [`M-x tide-restart-server`](https://github.com/ananthakumaran/tide/wiki/Explanations-about-the-Issue-Reporting-Checklist#try-m-x-tide-restart-server) in the buffer where I had the problem.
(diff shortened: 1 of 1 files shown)


josteink:
1. If people use MELPA (and *not* MELPA stable), master is always latest. Advice here could simply be that they should update all packages to latest (because we can’t/won’t debug whatever weird combo their setup they’ve ended up getting curtails). 

2. I’m not sure what the best easy/obvious way to present this idea is. “How to determine which project the issue belongs to” might be better?

josteink:
Regarding the actual commit, I think the checklist is both thourough enough, yet reasonably sparse not to be unreasonable. 

Good job!

ananthakumaran:
This is really great. I had contemplated about doing similar thing but never got around to do it. This is through and clear.

lddubeau:
I'm glad we're in agreement. I've edited in light of @josteink 's comments, fixed the links that became broken after editing the Wiki, and merged with master.

PR #249, opened 2018-04-16
fix invalid tide-on-response-success calls

ananthakumaran:
@lddubeau What do you think about this approach? Using `&optional` and `&rest` in the same function seems like a tricky thing, I am not entirely sure if this will introduce any other issues. 

P.S. Accidentally pushed to master(reverted now)

### DIFF
--- a/tide.el
+++ b/tide.el
@@ -272,6 +272,11 @@ above."
    args
    :initial-value list))
 
+(defun tide-plist-map (fn plist)
+  (when (and plist (not (cl-evenp (length plist))))
+    (error "Invalid plist %S" plist))
+  (-map (-lambda ((key value)) (funcall fn key value)) (-partition 2 plist)))
+
 (defun tide-combine-plists (&rest plists)
   "Create a single property list from all plists in PLISTS.
 The process starts by copying the first list, and then setting properties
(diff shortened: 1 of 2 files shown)


tide.el
```diff
@@ -1120,13 +1130,14 @@ in the file that are similar to the error at point."
...
   (let ((response (tide-command:getEditsForRefactor (plist-get selected :refactor) (plist-get selected :action))))
     (tide-on-response-success response
-      (deactivate-mark)
-      (tide-apply-code-edits (tide-plist-get response :body :edits))
-      (-when-let (rename-location (tide-plist-get response :body :renameLocation))
-        (with-current-buffer (tide-get-file-buffer (tide-plist-get response :body :renameFilename))
-          (tide-move-to-location rename-location)
-          (when (tide-can-rename-symbol-p)
-            (tide-rename-symbol)))))))
+      (progn
```
josteink: If body can now only be a single expression (or wrapped in a progn) this should be documented in a doc-string.

ananthakumaran: strictly speaking, it will accept multiple expressions if options param is passed explicitly, otherwise, it's ambiguous. So we need to wrap it with `progn`. I will add some documentation

PR #253, opened 2018-05-10
Recenter screen after jump

invokesus:
Recenter call added in `tide-jump-to-filespan`. However, condition is not checked whether the definition line was already in view or not.

### DIFF
--- a/tide.el
+++ b/tide.el
@@ -174,6 +174,11 @@ above."
   :type '(choice (const "TS") (const "TSX") (const "JS")(const  "JSX"))
   :group 'tide)
 
+(defcustom tide-no-recenter-after-jump t
+  "Recenter buffer after jumping to definition"
+  :type 'boolean
+  :group 'tide)
+
 (defmacro tide-def-permanent-buffer-local (name &optional init-value)
   "Declare NAME as buffer local variable."
   `(progn
@@ -750,15 +755,35 @@ implementations.  When invoked with a prefix arg, jump to the type definition."
     (tide-move-to-location location)
     (point)))
 
+(defun tide-recenter-p (filespan &optional no-recenter-pref)
+  (if no-recenter-pref
+      nil
+    (let* ((new-file-name (plist-get filespan :file)))
+      (if (string-equal new-file-name (tide-buffer-file-name))
+          (tide-recenter-in-same-buffer-p filespan)
+        t))
+    )
+  )
+
+(defun tide-recenter-in-same-buffer-p (filespan)
+  (let* ((newpos (plist-get (plist-get filespan :start) :line))
+         (line-diff (abs (- (line-number-at-pos) newpos))))
+    (if (> line-diff (count-screen-lines))
+        t
+      nil)))
+
 (defun tide-jump-to-filespan (filespan &optional reuse-window no-marker)
-  (let ((file (plist-get filespan :file)))
+  (let ((file (plist-get filespan :file))
+        (recenter-decision (tide-recenter-p filespan tide-no-recenter-after-jump)))
     (unless no-marker
       (ring-insert find-tag-marker-ring (point-marker)))
     (if reuse-window
         (pop-to-buffer (tide-get-file-buffer file) '((display-buffer-reuse-window display-buffer-same-window)))
       (pop-to-buffer (tide-get-file-buffer file)))
     (tide-move-to-location (plist-get filespan :start))
-    (recenter)))
+    (if recenter-decision
+        (recenter)))
+  )
 
 (defalias 'tide-jump-back 'pop-tag-mark)
 
--- a/tide.el
+++ b/tide.el
@@ -174,7 +174,7 @@ above."
   :type '(choice (const "TS") (const "TSX") (const "JS")(const  "JSX"))
   :group 'tide)
 
-(defcustom tide-recenter-after-jump nil
+(defcustom tide-recenter-after-jump t
   "Recenter buffer after jumping to definition"
   :type 'boolean
   :group 'tide)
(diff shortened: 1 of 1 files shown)


josteink:
So I guess this PR is on hold until you've done some more research then? :)

invokesus:
Will look into this over the weekend.🙂

invokesus:
Didn't come across any easy way to detect whether a line was on the screen or not. So used differences of `line-numbers` along with  `count-screen-lines` to check whether jump is required or not when jump point is in the same buffer.

tide.el
```diff
@@ -750,14 +755,35 @@ implementations.  When invoked with a prefix arg, jump to the type definition."
...
+(defun tide-recenter-p (filespan &optional no-recenter-pref)
+  (if no-recenter-pref
+      nil
+    (let* ((new-file-name (plist-get filespan :file)))
+      (if (string-equal new-file-name (tide-buffer-file-name))
+          (tide-recenter-in-same-buffer-p filespan)
+        t))
+    )
+  )
+
```
josteink: This isn't C. Tidy up those parenthesis 😄

tide.el
```diff
@@ -174,6 +174,11 @@ above."
   :type '(choice (const "TS") (const "TSX") (const "JS")(const  "JSX"))
   :group 'tide)
 
+(defcustom tide-no-recenter-after-jump t
+  "Recenter buffer after jumping to definition"
+  :type 'boolean
+  :group 'tide)
+
```
josteink: It may be a personal preference, but I really dislike negative settings. `If (no = true)` is a bad way to write conditionals when you want the code to clearly communicate *intent*.

And if you ever need to take the negative of that, you now have a messy double-negative and the cognitive load of that should IMO really not ever be needed. 

I vote you make the preference `tide-recenter-after-jump` instead.

Also: Since `tide-recenter-in-same-buffer-p` ensures we don't recenter when things are already visible in the buffer, feel free to keep the default for this new preference `t`.

tide.el
```diff
@@ -750,14 +755,35 @@ implementations.  When invoked with a prefix arg, jump to the type definition."
...
+        t))
+    )
+  )
+
+(defun tide-recenter-in-same-buffer-p (filespan)
+  (let* ((newpos (plist-get (plist-get filespan :start) :line))
+         (line-diff (abs (- (line-number-at-pos) newpos))))
+    (if (> line-diff (count-screen-lines))
+        t
+      nil)))
```
josteink: `(if bool t nil)`?

Can't the entire `if` statement be removed here and be replaced by `(> line-diff (count-screen-lines))` alone?

tide.el
```diff
@@ -750,14 +755,35 @@ implementations.  When invoked with a prefix arg, jump to the type definition."
...
+
 (defun tide-jump-to-filespan (filespan &optional reuse-window no-marker)
-  (let ((file (plist-get filespan :file)))
+  (let ((file (plist-get filespan :file))
+        (recenter-decision (tide-recenter-p filespan tide-no-recenter-after-jump)))
```
josteink: Nitpick about variable name: I don't think it communicates intent well enough. A decision? What *kind* of decision?

I'd prefer for it to be `should-recenter` (`should-recenter-p`?) or something along those lines.

tide.el
```diff
@@ -750,14 +755,35 @@ implementations.  When invoked with a prefix arg, jump to the type definition."
...
     (unless no-marker
       (ring-insert find-tag-marker-ring (point-marker)))
     (if reuse-window
         (pop-to-buffer (tide-get-file-buffer file) '((display-buffer-reuse-window display-buffer-same-window)))
       (pop-to-buffer (tide-get-file-buffer file)))
-    (tide-move-to-location (plist-get filespan :start))))
+    (tide-move-to-location (plist-get filespan :start))
+    (if recenter-decision
+        (recenter)))
+  )
```
josteink: Again: This isn't C. Please clean up the trailing parenthesis 😄

tide.el
```diff
@@ -750,14 +755,35 @@ implementations.  When invoked with a prefix arg, jump to the type definition."
...
+  (let ((file (plist-get filespan :file))
+        (recenter-decision (tide-recenter-p filespan tide-no-recenter-after-jump)))
     (unless no-marker
       (ring-insert find-tag-marker-ring (point-marker)))
     (if reuse-window
         (pop-to-buffer (tide-get-file-buffer file) '((display-buffer-reuse-window display-buffer-same-window)))
       (pop-to-buffer (tide-get-file-buffer file)))
-    (tide-move-to-location (plist-get filespan :start))))
+    (tide-move-to-location (plist-get filespan :start))
+    (if recenter-decision
```
josteink: And `if` form with no `else`-clause should be replaced with a `when`-form instead.

josteink:
That does look much better. I haven't fully *tested* the code yet, but from a glance it mostly looks like it does what it should.

I do have some smaller issues with the *form* of the code though, and I've left some review-comments and nitpicks for the issues I've found.

Let me know what you think, if there's something you wonder about, or if you have a new revision for me to look over!

invokesus:
Thanks for the review. Sorry about the messy code; first time writing elisp 😅. Made changes according to the comments. Let me know if anything else needs to be changed.

josteink:
No worries. That looks much better!

But shouldn’t we set the new ‘defcustom’ by default to be true? That shouldn’t change any existing behavior in a bad way, right?

invokesus:
Turned it on by default 🙂. 

Existing behavior was to **never** recenter, now it will recenter in **all** cases, except when the destination point is also on the screen **and** in the same buffer as the point where jump was invoked.

Behavior is similar to that of `xref`, only with the added check stated above.

josteink:
I've re-ordered and squashed a few of the commits to make the merge cleaner, but all in all it's still your code, and it's now merged to `master`.

Thanks for the patch!

PR #254, opened 2018-05-14
Add tide-add-tslint-disable-next-line

lddubeau:
This is not complete yet because I've not added a test for it to the test suite, but I've been sitting on this since last October. It is high time it be integrated *somewhere*.

This essentially adds a function that allows for automatically turning off `tslint` warnings. The gory details are in the docstring for the new function, but here is an example:

```
if (something.foo("blah"))
```

If you have 5 `tslint` rules giving you warnings at `foo` and you move your point there and issue `tide-add-tslint-disable-next-line` then your code will become:

```
// tslint:disable-next-line:[rule1] [rule2] [rule3] [rule4] [rule5]
if (something.foo("blah"))
```

I feel the function I'm adding here is more `tide`-y than `typescript-mode`-y but maybe there's a good argument to be made that it belongs to `typescript-mode` rather than `tide`.

### DIFF
--- a/tide-tests.el
+++ b/tide-tests.el
@@ -22,6 +22,7 @@
 
 (require 'tide)
 (require 'dash)
+(require 'cl)
 
 (defun tide-plist-equal (a b)
   (and (listp a) (listp b)
@@ -123,6 +124,65 @@
   (should-error (tide-load-tsconfig "test/loop.json" '()))
 
   (should-error (tide-load-tsconfig "test/notfound.json" '())))
+
+;; Adapted from jdee-mode's test suite.
+(defmacro test-with-temp-buffer (content &rest body)
+  "Fill a temporary buffer with `CONTENT' and eval `BODY' in it."
+  (declare (debug t)
+           (indent 1))
+  `(with-temp-buffer
+     (insert ,content)
+     (tide-mode)
+     (goto-char (point-min))
+     ;; We need this so that tests that simulate user actions operate on the right buffer.
+     (switch-to-buffer (current-buffer))
+     ,@body))
+
+(defun test-tide-add-tslint-disable-next-line (mock initial goto expected)
+  (test-with-temp-buffer
+   initial
+   (search-forward goto nil t)
+   (goto-char (match-beginning 0))
+   (cl-letf (((symbol-function 'tide-get-flycheck-errors-ids-at-point)
+              (lambda () mock)))
+     (tide-add-tslint-disable-next-line)
+     (should (string= (buffer-string) expected)))))
+
+(ert-deftest tide-add-tslint-disable-next-line/no-errors ()
+  "If there are no flycheck errors at point, it should be a no-op."
+  (let ((initial "// tslint:disable-next-line:previous\nconst x = 1;\nconst y = 2;\n"))
+    (test-tide-add-tslint-disable-next-line
+     ()
+     initial
+     "const y"
+     initial)))
+
+;; The first line of the buffer consitutes an edge condition. An
+;; earlier version of the function failed in this case.
+(ert-deftest tide-add-tslint-disable-next-line/first-line ()
+  "Create a new tslint flag if needed. (First line of buffer)"
+  (let ((initial "const x = 1;\nconst y = 2;\n"))
+    (test-tide-add-tslint-disable-next-line
+     '("err1" "err2")
+     initial
+     "const x"
+     (concat "// tslint:disable-next-line:err1 err2\n" initial))))
+
+(ert-deftest tide-add-tslint-disable-next-line/subsequent-lines ()
+  "Create a new tslint flag if needed. (Subsequent lines of buffer)"
+  (test-tide-add-tslint-disable-next-line
+   '("err1" "err2")
+   "const x = 1;\nconst y = 2;\n"
+   "const y"
+   "const x = 1;\n// tslint:disable-next-line:err1 err2\nconst y = 2;\n"))
+
+(ert-deftest tide-add-tslint-disable-next-line/adds ()
+  "If there was already a disable-next-line flag, add to it."
+  (test-tide-add-tslint-disable-next-line
+   '("err1" "err2")
+   "// tslint:disable-next-line:previous\nconst x = 1;\nconst y = 2;\n"
+   "const x"
+   "// tslint:disable-next-line:previous err1 err2\nconst x = 1;\nconst y = 2;\n"))
 
 (provide 'tide-tests)
 
--- a/tide.el
+++ b/tide.el
@@ -1187,6 +1187,49 @@ in the file that are similar to the error at point."
            (tide-select-refactor body))
         (message "No refactors available.")))))
 
+;;; Disable tslint warnings
+
+(defconst tide-tslint-disable-next-line-regexp
+  "\\s *//\\s *tslint\\s *:\\s *disable-next-line\\s *:\\(.*\\)"
+  "Regexp matching a tslint flag disabling rules on the next line.")
+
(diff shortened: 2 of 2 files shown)


josteink:
> This is not complete yet because I've not added a test for it to the test suite

As such I'll refrain from any *deep* code-review until later. Hope you don't mind.

> I feel the function I'm adding here is more tide-y than typescript-mode-y but maybe there's a good argument to be made that it belongs to typescript-mode rather than tide.

Right now the implementation depends on pre-existing code which is only found in `tide`, which I think (for now) is a good guide for where this should go.

That said... And this is going to be a bit of a meta-discussion... But you started it! 😃 

Every time I have this uneasy feeling about *which* project the code should be added to, I keep asking myself do we really need *two* different packages for this?

While I can appreciate the idea about separation of concerns (syntax highlighting vs code-intelligence) and that `tide` can be used on plain JavaScript-files too... There's always going to be some blurred lines.

(For example most major-modes which don't do fancy code intelligence stuff, still does basic code-navigation through `imenu`).

I don't really think there's a single Emacs `typescript-mode`-user which is *against* having the features `tide` provides, nor do I think it's very common to *only* install `typescript-mode`.

Are we maintaining a needless separation over purely philosophical concerns?

Right now we also have pretty steep divide between users contributing to improving `tide` (which IMO is doing pretty well) and users contributing to improving `typescript-mode` (which isn't doing near as well).

Would joining these projects benefit both, or would we be harming `tide` in our attempt to help lift `typescript-mode`?

I don't know. Do you have any opinions on the matter?

lddubeau:
> As such I'll refrain from any deep code-review until later. Hope you don't mind.

I don't mind.

> That said... And this is going to be a bit of a meta-discussion... But you started it!

That's absolutely fine.

It depends what form "merging" would take. As we speak, if someone is in a situation where they want to have TS editing support but cannot or do not want to have a `tsserver` started. It is trivial: they use `typescript-mode` without `tide`. Both may be *installed* but if they have a need to use `typescript-mode` in a buffer without `tide`, they can do it. Sometimes people need to edit fragments of TS code embedded in a larger context which is not TS code. For instance if a developer is writing documentation for a documentation system that knows nothing about TS. Right now, using `typescript-mode` without `tide` for those fragments is easy.

It would be possible to merge `tide` and `typescript-mode` into a single project that provides the two modes side-by-side, and then the possibility to use one without the other would still be preserved. Or we could actually merge the modes into a single one, but I'd argue for having the capability to start the mode with or without a `tsserver`, with the understanding that if `tsserver` is not present, then the functions that depend on it are not available.

> Right now we also have pretty steep divide between users contributing to improving tide (which IMO is doing pretty well) and users contributing to improving typescript-mode (which isn't doing near as well).

> Would joining these projects benefit both, or would we be harming tide in our attempt to help lift typescript-mode?

Hmm... I cannot see a merge substantially helping `typescript-mode` or hurting `tide`. Suppose we merge the two modes into a single mode. I'd expect that post-merge the parts that came from `typescript-mode` would still receive comparatively less contributions than those parts that came from `tide`. So I'd expect the disparity to apply to different parts of a single project just the same as it applies now to two different projects. I don't know that anyone would not contribute code for a new `tsserver` command because they cannot figure out the fontification code.

I think it is hard to contribute to `typescript-mode` because there are substantial parts of it that require  very Emacs-specific knowledge. The fontification code is an example of that. During review of the fontification PR to add special highlighting for class names, I mentioned one change that was affecting fontification *levels*. The way the API for fontification is set up it is hard to know that there's even a notion of fontification levels. It is possible to discover it through reading the documentation but, as I recall, it is not particularly easy to discover, short of reading everything about fontification. I could name syntax tables as another bit that is very Emacs-specific.

----

I also suspect that `tsserver` is improved we'll get more and more issues that challenge the division between `tide` and `typescript-mode`.

lddubeau:
I've fixed a bug and added tests. So if anyone wants to review the new functionality, it's ready for that!

@ananthakumaran Opinions about the larger [questions raised by josteink](https://github.com/ananthakumaran/tide/pull/254#issuecomment-389075691)?

ananthakumaran:
There are two main arguments against merging tide/typescript-mode

1) tide supports js/jsx/tsx and not all of them use/need typescript-mode. As tsserver has plugin architecture, support for more types could be expected in the future.

2) Not everybody need battery loaded setup, sometimes I just downgrade to fundamental mode because the mode/plugin doesn't allow me to do what I want.


As for whether this one should go to tide/typescript-mode, I don't have any strong opinions on either way, so I am ok with any decision.

josteink:
Right. Sorry about the long delay in commenting on this. Been busy prepping moving and what not.

W.r.t. typescript-mode and tide being different packages and all that... @ananthakumaran brings up some good points, but I guess it's not really relevant to this PR. It's OK for me to take that issue down the road should it become more pressing.

Leaving that behind, let's talk about the PR itself.

As for this PR, or any feature, I have 3 main considerations:

* The code itself
* Test-coverage
* The user-oriented part of the feature (how does it *flow*?)

The code-itself looks OK. We have test-coverage so that makes me happy.

That leaves the part of how the feature "flows"... In which case I have some ideas... Which we may or may not decide to act on right now.

In Idea/Rider/VisualStudio (and VSCode and more I'm sure), there's often this one general dispatch-function with a single, standard key-binding to just *act* on ... anything on a certain line.

This action can be anything, but usually involves fixing/improving stuff:

* Importing namespaces/packages
* Simplifying code
* Available refactorings
* Fixes for or ability to disable compiler/linting-warnings.
* Etc

Basically, in VS or IntelliJ, I can just `alt-enter` to get a popup of whatever the IDE thinks may be useful for me right now. That's extremely smooth and makes discoverability of built in IDE features much easier.

Right now this feature is currently exposed via a single defun `tide-add-tslint-disable-next-line` and it is left for the user to discover for himself. The *spirit* of this feature however seems to plays very nicely along with the spirit of `tide-code-fix`.

Which kinda gives us at least *two* tide functions for "please help me fix this code at this point". And I don't want to have 2 different key-bindings for that!

If we keep adding stuff like this, we should consider adding a single general dispatch-function which investigates the current state of the code at point, and then provides useful suggestions, and possibly also a default-binding for this function.

Opinions? Disagreements?

If you guys agree with me that this would on overall be an improvement, should we:

* Also consider adding that general-dispatch mechanisms to this PR (You broke the single function-fix-paradigm. You go fix it!)
* Or does it make more sense to leave it for later? (General dispatch-function is a feature of its own, best discussed on its own in its own a PR)

Either approach is fine by me, so I'm curious about what you guys think.

lddubeau:
@josteink I agree that having a general dispatch for "fix something here" would be desirable. I'm absolutely in favor of it. I suppose it would be something that queries `tsserver` for possible fixes like `tide-fix` does but also adds options that are proper to tide like `tide-add-tslint-disable-next-line`.

However, I'd rather see this PR be merged without the *precondition* that such "general dispatch" be created. My PRs scratch the specific itches I've been having. This PR is no exception. I've been itching for a function for turning off contextually spurious `tslint` errors. I'm not itching yet for having a "general dispatch" functionality. So if it becomes a precondition, it will take a while before the PR gets finalized.

ananthakumaran:
I agree, we could handle the unified fix in another PR

josteink:
That's 2 votes for "it sounds nice, but lets do it in another PR".

As such, I have no objections to this PR. Merge at will.

lddubeau:
Of course, two seconds *after* I merge this, it occurs to me that the new function should be documented in the README, like functions that perform similar jobs... you know,  "discoverability" and all that jazz, as we've *just been discussing*. oops...

I've pushed a commit to `master` that fixes that omission.

Issue #184, opened 2017-08-09
Emacs hangs on large outputs from tsserver


domachine:
Hi,

thanks for this amazing project. One thing keeps bugging me though. From time to time emacs starts freezing and a message pops up saying that the undo buffer for the `*tide-server*` buffer becomes too large. From that time on I started to manually call `buffer-disable-undo` on the `*tide-server*` buffer once the ts server started up. This stops emacs from freezing. I wondered if it's safe to disable `undo` for this buffer automatically once the mode is loaded. I'll happily hacking together a pull-request if this an option that you're fine with.

Thank you, for sacrificing your spare time!

ananthakumaran:
go ahead and open a PR. I find no reason to enable undo in `*tide-server*` buffer.

domachine:
Ok on further investigation, I have to admit this still occurs :-(. I just thought it was related to the undo buffer. This mainly occurs when working with the `react-apollo` library. I already found an issue on their repo (https://github.com/apollographql/react-apollo/issues/786) that I think is related. The typings seem so complex that this slows down the typescript language server in general. I'm not very familiar with the inner workings of `tide` but is there any way to handle this in a way that doesn't let emacs hang several seconds? And with hanging I mean the cursor freezes and the editor is not even responsive to `ctrl+g`.

Oh, and this is my setup by the way:

- `GNU Emacs 25.1.1 (x86_64-apple-darwin16.0.0, NS appkit-1504.00 Version 10.12 (Build 16A323)) of 2016-10-27`
- `macOS Sierra 10.12.5`

ananthakumaran:
> doesn't let emacs hang several seconds?
This is a bad thing. Can you provide more details
1) The command which causes/triggers the problem
2) The request/response log related to that command see https://github.com/ananthakumaran/tide#debugging for details on how to capture it.
3) You can use `toggle-debug-on-quit` to get the stack trace when you quit using ctrl-g

domachine:
This isn't actually triggered by a special command. It more or less occurs now and then when I work with a file (moving the cursor, normal typing). I've got a feeling that this happens once I pass a complex function (like `graphql` from `react-apollo`) with the cursor and that the generation of the documentation shown in the minibar is the trigger of that problem. But I currently see no deterministic way to reproduce. All I can say is that this is almost every time combined with the undo buffer message (which I now have suppressed using my configuration). Maybe it's triggered by a huge parsing overhead of the tsserver response.

[...]

PR #257, opened 2018-05-21
Fix javascript code completion

vshender:
This PR solves the issue with JavaScript code completion.

The problem is that tsserver often returns a lot of irrelevant completions. A distinctive feature of such completion items is a `:kind` property with a value of `"warning"`. Moreover, I've paid attention to one more difference: for all relevant completions the value of `:sortText` property is `"0"`, and for all irrelevant is `"1"`.

As it follows from [here](https://github.com/Microsoft/TypeScript/blob/master/src/server/protocol.ts#L1857), this property is intended to be used for sorting completion items. The single place where completions with `:sortText` property equal to `"1"` are returned is [`getJavaScriptCompletionEntries()`](https://github.com/Microsoft/TypeScript/blob/master/src/services/completions.ts#L160) function that just returns our irrelevant `"warning"` completions. Moreover, if you look at the [development history](https://github.com/Microsoft/TypeScript/commit/f583fd451e312f796647c31137cdd50887bbb867#diff-233e1126c0abc811c4098757f9e4516eR3159), you can find that these completions are indeed intended to go after all others.

Therefore, I've decided to implement this behavior in Tide, and now completion list looks much better. (BTW, the same approach is used in [Vim plugin](https://github.com/Quramy/tsuquyomi/blob/master/autoload/tsuquyomi.vim#L292)).

Example before the changes:
test js 2018-05-21 14-25-26

And after:
test js 2018-05-21 14-37-33

Example after the changes with completion items by kind sorting enabled:
test js 2018-05-21 14-30-45

I want to note that this sorting will not affect the TypeScript code completion, since the elements with `:sortText` property equal to `"1"` are only returned for JavaScript.

I've also added an option to remove results with `"warning"` `:kind` property from a completion list. I've decided to not enable this behavior based on any heuristics such as file extension, etc. Instead this behavior is configured by `tide-filter-out-warning-completions` variable. In the simplest case you can just bind it locally for buffer in a javascript-mode hook.

Example using warning completions filtering:
test js 2018-05-21 14-32-13


Nek:
@vshender thank for thorough analysis and quick fix. The solution looks great to me.

tide.el
```diff
@@ -1261,11 +1277,14 @@ in the file that are similar to the error at point."
...
+                          (or (not tide-filter-out-warning-completions)
+                              (not (equal (plist-get completion :kind) "warning")))))
                    completions)))
-     (if tide-sort-completions-by-kind
-         (-sort 'tide-compare-completions filtered)
-       filtered))))
+     (let ((sorted (-sort 'tide-compare-completions-basic filtered)))
+       (if tide-sort-completions-by-kind
+           (-sort 'tide-compare-completions-by-kind sorted)
+         sorted)))))
```
josteink: If I'm reading this correctly this code

1. first sorts the results
2. checks the `tide-sort-completions-by-kind` preference
3. and if the preference is true, sorts the sorted results once again, before returning.

Wouldn't it be better to just check the preference and sort once?

This *is* after all user-interactive code and I think doing twice the work we need before returning seems wasteful.

Or am I missing something?

josteink:
All in all the change looks mostly good. 

I appreciate you putting in work on updating the test-cases.

These changes should *logically* have a very local impact, and as I can see you've constrained the changes to a very small, isolated part of the code-base.

That said... I have a little nitpick on change none the less, and left a comment in the code near that section.

Once we've cleared that bit up, I don't see any significant issues with getting this merged.

tide.el
vshender: The reason is that if `tide-sort-completions-by-kind` is true then completions should be sorted using both `tide-compare-completions-basic` and `tide-compare-completions-by-kind`. The internal sorting algorithm is *stable*, so sorting completions twice works as intended (completions with `sortText` equal to `"1"` remain last in the completion list).

As an alternative solution I can modify `tide-compare-completions-by-kind` to additionally compare completions by `sortText` property and sort completions only once using this comparator if `tide-sort-completion-by-kind` is true:

    (if tide-sort-completions-by-kind
        (-sort 'tide-compare-completions-by-kind filtered)
      (-sort 'tide-compare-completions-basic filtered))

Actually, I've done it in the first version of the fix, but this solution makes `tide-compare-completions-by-kind` comparator more complex, and I don't like that this more complex comparator includes the logic of another one (we can't just call `tide-compare-completions-basic` from`tide-compare-completions-by-kind` in this case). I believe that slight performance improvement is not worth it (completion list is unlikely to contain even thousands of elements, so sorting of such a list is really fast operation), but if you think that this approach is better, I will certainly make the necessary corrections.

ananthakumaran: > completion list is unlikely to contain even thousands of elements, so sorting of such a list is really fast operation

this is not true,  tsserver is known to return #184 so much data that we can't even parse without causing significant delay. Also we take some special care during completion request (always send a prefix except for method completion) so that tsserver won't send back too much data. Triggering auto-complete from the top level scope with browser libs enabled usually results in a large response

josteink: > but this solution makes tide-compare-completions-by-kind comparator more complex, and I don't like that this more complex comparator includes the logic of another one

Agreed. I'm against code-repetition too.

> we can't just call tide-compare-completions-basic fromtide-compare-completions-by-kind in this case

OK. That would have been my second suggestion, but I guess that's not an option then.

> if you think that this approach is better, I will certainly make the necessary corrections.

Given your explanation, I'm mostly alright with the current code-structure then, but I think from the outside looking in, this particular piece of code either needs:

1. some better semantics (to make it obvious what is happening and why), maybe store like storing it in variable-names with more descriptive names?
2. or some comments or doctrings to explain why we are "sorting things twice".

Basically just to aid a future maintainer, contributor, etc to avoid breaking this piece of code in the future.

Does that sound reasonable?

vshender: Yes, it's absolutely reasonable. But now, given that completion lists can be really large, I'm not sure that sorting the list twice is a good idea :(

ananthakumaran: Would it work if we just compose two comparator? [java stdlib](http://grepcode.com/file/repository.grepcode.com/java/root/jdk/openjdk/8-b132/java/util/Comparator.java#Comparator.thenComparing%28java.util.Comparator%29)

vshender: I can write something like this:

    (defun compose-comparators (cmp1 cmp2)
      "Compose two comparators."
      (lexical-let ((cmp1 cmp1) (cmp2 cmp2))
        (lambda (a b)
          (or (funcall cmp1 a b)
             (if (not (funcall cmp1 b a))
                 (funcall cmp2 a b))))))

    ...

    (let ((compare-completions
           (if tide-sort-completions-by-kind
               (compose-comparators 'tide-compare-completions-basic
                                    'tide-compare-completions-by-kind)
             'tide-compare-completions-basic)))
      (-sort compare-completions filtered))

ananthakumaran: LGTM, I guess lexical-let might not be necessary as we set `-*- lexical-binding: t -*-` in the file

josteink:
````
(defun compose-comparators (cmp1 cmp2)
````

No problem with the rest of that approach, but this function should probably be namespaced properly with a tide-prefix.

Comment (vshender) [reply tide.el]: It was just a sample, the real function name has the prefix :)

**Comment by vshender** [comment]
Is the current version of fix good enough to be merged?

File: tide.el [hunk 2]
```diff
@@ -1261,11 +1284,16 @@ in the file that are similar to the error at point."
...
+                     (and (string-prefix-p prefix (plist-get completion :name))
+                          (or (not tide-filter-out-warning-completions)
+                              (not (equal (plist-get completion :kind) "warning")))))
                    completions)))
-     (if tide-sort-completions-by-kind
-         (-sort 'tide-compare-completions filtered)
-       filtered))))
+     (let ((completions-comparator
+            (if tide-sort-completions-by-kind
+                (tide-compose-comparators 'tide-compare-completions-basic
```
josteink: It’s considered good Emacs lisp convention to quote functions using #' instead of just '.

vshender: I agree, but the convention with just a single `'` have already been used in the [source code](https://github.com/ananthakumaran/tide/blob/master/tide.el#L1267), that I intended to fix, so I've decided to follow it for uniformity.

I see that you mostly use `'#` convention, but there are some places, where just a single `'` is used. I can fix them too.

josteink: It’s not *that* big a deal, I guess. Leave it be and let’s see what mr actual repo-owner thinks instead :)

PR #258, opened 2018-05-24
add tide-rename-file command

ananthakumaran:
NOTE: it only works with the master branch of TypeScript repo.


tide.el
```diff
@@ -1579,6 +1579,35 @@ number."
...
+  (interactive)
+  (let* ((name (buffer-name))
+         (old (tide-buffer-file-name))
+         (basename (file-name-nondirectory old)))
+    (when (not (and old (file-exists-p old)))
+      (error "Buffer '%s' is not visiting a file." name))
+    (let ((new (read-file-name "New name: " (file-name-directory old) basename nil basename)))
+      (when (get-buffer new)
+        (error "A buffer named '%s' already exists." new))
+      (when (file-exists-p new)
```
josteink: How will this handle cases where the same filename is re-used in several sub-folders?

Example:

````
src
|--network
|  |--stringUtil.ts
|  \-- etc
|--controls
|  |-- textbox.ts
|  |-- label.ts
````

Let's say `stringUtil` is open in a buffer. And you want to rename another buffer (in `src` or `controls`) or whatever to stringUtil too. That's technically valid, but will probably be stopped by this code. 

Shouldn't we let `tsserver` handle the error-checking for us here?

And what about buffer-names when we have several buffers open based on the same filename... will `(get-buffer old)` be reliable or sufficient?

I think we may have to look into better ways to handle cases like this in a reliable fashion.

ananthakumaran: > Let's say stringUtil is open in a buffer. And you want to rename another buffer (in src or controls) or whatever to stringUtil too. That's technically valid, but will probably be stopped by this code.

I tried this, it works without any issue. If there is an existing buffer with the same name in a different folder, the folder name is added to both the new & old buffer (I guess uniquify kicks in does the work). This throws an error if you try to do rename to existing buffer in the same folder though.

ananthakumaran: switched `get-buffer` to `get-file-buffer` which should do the search based on filename

josteink: OK. Fair enough. Just checking. If that's the case, this does look mostly reasonable. 

That said, I haven't actually *tested* the code by running it yet. (Don't have time to setup with master TypeScript repo right now)

ananthakumaran:
this should work with 2.9, we could merge if somebody could confirm it's working

PR #262, opened 2018-06-08
support move-to-file refactor

ananthakumaran:
This [refactor](https://code.visualstudio.com/updates/v1_24#_move-to-new-file-refactoring) is only enabled if `allowTextChangesInNewFiles` flag is
set on the configure request. Handle code edits for non existent file.

### DIFF
--- a/tide.el
+++ b/tide.el
@@ -773,7 +773,7 @@ implementations.  When invoked with a prefix arg, jump to the type definition."
       (widen)
       (goto-char (point-min))
       (forward-line (1- line)))
-    (when (not (eql 1 (point)))
+    (when (not (and (= offset 0) (= line 0)))
       (forward-char (1- offset)))))
 
 (defun tide-location-to-point (location)
--- a/tide.el
+++ b/tide.el
@@ -98,6 +98,15 @@ above."
   :type '(plist :value-type sexp)
   :group 'tide)
 
+(defcustom tide-user-preferences '(:includeCompletionsForModuleExports t :includeCompletionsWithInsertText t :allowTextChangesInNewFiles t)
+  "User preference plist used on the configure request.
+
+Check
+https://github.com/Microsoft/TypeScript/blob/17eaf50b73c1355d2fd15bdc3912aa64a73483dd/src/server/protocol.ts#L2684
+for the full list of available options."
+  :type '(plist :value-type sexp)
+  :group 'tide)
+
 (defcustom tide-completion-ignore-case nil
   "CASE will be ignored in completion if set to non-nil."
   :type 'boolean
@@ -314,6 +323,17 @@ ones and overrule settings in the other lists."
   (if (equal file (tide-buffer-file-name))
       (current-buffer)
     (find-file-noselect file)))
+
+(defun tide-get-or-create-file-buffer (file)
+  (cond
+   ((equal file (tide-buffer-file-name)) (current-buffer))
+   ((file-exists-p file) (find-file-noselect file))
+   (t (let ((buffer (create-file-buffer file)))
+        (with-current-buffer buffer
+          (set-visited-file-name file)
+          (basic-save-buffer)
+          (display-buffer buffer t))
+        buffer))))
 
 (defun tide-response-success-p (response)
   (and response (equal (plist-get response :success) t)))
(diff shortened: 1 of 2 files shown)


tide.el
```diff
@@ -757,7 +773,8 @@ implementations.  When invoked with a prefix arg, jump to the type definition."
       (widen)
       (goto-char (point-min))
       (forward-line (1- line)))
-    (forward-char (1- offset))))
+    (when (not (and (= offset 0) (= line 0)))
```
josteink: Couldn't this be simplified to `(not (eql 0 (point)))` instead?

tide.el
```diff
@@ -307,13 +316,20 @@ ones and overrule settings in the other lists."
...
+  (cond
+   ((equal file (tide-buffer-file-name)) (current-buffer))
+   ((file-exists-p file) (find-file-noselect file))
+   (new-file (let ((buffer (create-file-buffer file)))
+               (with-current-buffer buffer
+                 (set-visited-file-name file)
+                 (basic-save-buffer)
+                 (display-buffer buffer t))
+               buffer))
+   (t (error "Invalid file %S" file))))
```
josteink: This line introduces a failure-mode when file doesn't exist, where it before would instead fall back to `find-file-nosolect`.

Is there any risk this might break other features calling this function?

josteink:
Whee! More refactorings!

Sorry about the slow response (been busy, etc etc), but now I've tested it and I can confirm it works.

One thing I notice though is that this refactoring just moves our code out to a new file immediately, and then it's done without asking what file name or location we'd prefer. Is this unchangeable `tsserver`-behaviour, or is this something we can provide if we like to?

The reason I ask is that conceptually I see this is new refactoring as a `extract-to-target` type of refactoring. And we already have (at least) 2 of those: extract function (to module scope & local scope), and they behave differently.

When we do a extract function refactoring, we are immediately asked about what *name* we want the function to have.

This new refactoring should ideally behave consistently with those,  so after we've extracted the class/function/whatever to a new file, we should immediately ask for a new filename and apply the rename-file refactoring too.

What do you think?

tide.el
ananthakumaran: will do

tide.el
ananthakumaran: actually no, before this tsserver always returned existing file, this case should not happen under any case.

josteink: OK. Then I'm fine with this. 👍

ananthakumaran:
Let me think, what you say should work as long as there is only one new file. This also means we are now doing something without any kind of hint from tsserver. `extra-to-target` refactor hints the editor to trigger the rename by supplying the rename location and filename. Effectively, we have to figure out it's a move to new file refactor (we don't try to do this anywhere) and then go through the changes, find the new files and if there is only one new file, then trigger a rename. VS code doesn't seem to ask for the new filename either from what I can see from their screenshots

josteink:
>  This also means we are now doing something without any kind of hint from tsserver. `extra-to-target` refactor hints the editor to trigger the rename by supplying the rename location and filename.

Ok, Fair enough. I guess that means that if want to solve this in a clean/proper way in `tide`, we must first lobby for `tsserver` to export this kind of meta-data or hints, so we have something to latch on to in a consistent way.

I'm terrible at following up issues these days... So I really don't think I'd be the best guy to launch this proposal in the typescript-repo.

If you don't think this is critical and don't want to follow this up with the TypeScript-team, I'm not going to push further on this matter.

It would have been *nice* to have, but I can live without this feature being 100% consistent with the rest of the refactorings since this inconsistency is clearly not the result of shoddy work on our end.

ananthakumaran:
There is also a request in vscode for similar flow https://github.com/Microsoft/vscode/issues/50685. Let's see how it goes.

ananthakumaran:
I am going to merge this for now, as the issue in vscode seems to go nowhere.

josteink:
OK. Fair enough.  We waited. We tried to do the right thing.

This is clearly better than not supporting the feature at all.

tide.el
ananthakumaran: thinking again, this is not same as `(not (eql 0 (point)))`. The guard is to prevent movement backward from point 1. I guess tsserver is breaking the contract here and sending 0 instead of 1 based offset.

Issue #265, opened 2018-06-15
tide-nav should navigate to all symbols, not just types


sandersn:
In a program like this, tide-nav should allow me to navigate to `f` as well as `A`.

```ts
interface A {
  a: number
}
function f(a: A) {
  return a.a + 1
}
```
## Expected Behavior

tide-nav returns `A` and `f` (plus possibly other symbols)

## Actual Behavior

tide-nav returns `A`

Note that commenting out the filter expression in tide-nav (`(setq navto-items (-filter ...))`) gives me the behaviour I want. It makes sense to query just for types sometimes, but I don't think it should be the default.

## Checklist

- [x] I have searched both open and closed issues and cannot find a duplicate.
- [x] I can [reproduce the problem](https://github.com/ananthakumaran/tide/wiki/Explanations-about-the-Issue-Reporting-Checklist#use-the-latest-version-of-the-relevant-packages) with the latest version of the relevant packages.
- [x] The problem still occurs after I issued [`M-x tide-restart-server`](https://github.com/ananthakumaran/tide/wiki/Explanations-about-the-Issue-Reporting-Checklist#try-m-x-tide-restart-server) in the buffer where I had the problem.
- [x] I verified that the version and the configuration file path reported by [`M-x tide-verify-setup`](https://github.com/ananthakumaran/tide/wiki/Explanations-about-the-Issue-Reporting-Checklist#check-the-information-reported-by-m-x-tide-verify-setup) are correct.
- [x] I am [positive](https://github.com/ananthakumaran/tide/wiki/Explanations-about-the-Issue-Reporting-Checklist#how-to-determine-which-project-the-problem-belongs-to) the problem does not belong to `typescript-mode` or `tsserver`.

## Relevant Version Numbers

- Tide:20180212.2229 from elpa, but the relevant code is the same in master
- TypeScript: 3.0, built from master
- Emacs: 25.2.2

ananthakumaran:
I guess the reasoning was not to show too many items, though it's not in the readme,  the function is documented as `Search and navigate to named types.`.

I guess we could revisit the decision. Some of the approaches are

1) make the filter configurable and set a sane default (my vote is for container elements like class/module/type/interface/enum)
2) we could change the behavior when called with a prefix arg (aka disable the filter)


```elisp
(-filter
 (lambda (navto-item) 
   (member (plist-get navto-item :kind) '("class" "interface" "type" "enum")))
 navto-items)
```

```typescript
    /**
[...]

PR #267, opened 2018-06-28
Make tide-nav filter customizable

sandersn:
Based on the discussion in #265, I made a simple customize entry for tide-nav's item filtering.

It's not fancy right now, for two reasons:
1. I'm not confident with defcustom (or emacs develoment in general), so I tried the simplest thing that would work.
2. I think that All and Types are the only two choices that make sense, so something fancier like a big checklist would only confuse people. In addition, implementing All by checking the entire checklist would also be slow, since it would be `member` of a medium-ish list filtered over a large list.

The biggest obstacle, however, is (1), so please do suggest improvements if you see them.

### DIFF
--- a/tide.el
+++ b/tide.el
@@ -113,9 +113,9 @@ above."
   :type 'boolean
   :group 'tide)
 
-(defcustom tide-nav-item-filter 'types
-  "The filter for items returned by tide-nav"
-  :type '(choice (const none) (const types))
+(defcustom tide-navto-item-filter #'tide-navto-item-filter-default
+  "The filter for items returned by tide-nav. Defaults to class, interface, type, enum"
+  :type 'function
   :group 'tide)
 
 (defface tide-file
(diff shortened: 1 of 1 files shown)


tide.el
```diff
@@ -854,10 +859,11 @@ Noise can be anything like braces, reserved keywords, etc."
                     (let ((response (tide-command:navto prefix)))
                       (tide-on-response-success response
                         (-when-let (navto-items (plist-get response :body))
-                          (setq navto-items
-                                (-filter
-                                 (lambda (navto-item) (member (plist-get navto-item :kind) '("class" "interface" "type" "enum")))
-                                 navto-items))
+                          (if (eq tide-nav-item-filter 'types)
```
josteink: `if` forms without any else-clause is usually best expressed using `when`.

ananthakumaran:
yes, I second @josteink. otherwise, it won't be possible to edit the filter-list without making further changes to tide.

josteink:
Thanks for the update. I like this much better!

If I were to nitpick anything (or bring something up for discussion) it's if we should have a *filter*-function or a *predicate*-function.

Right now we have this built in filter:

````elisp
(defun tide-navto-item-filter-default (navto-items)
  (-filter
   navto-items))
````

Invoked like this:

````elisp
(setq navto-items (funcall tide-navto-item-filter navto-items))
````

In this case we have a customizable *filter*-function.

My guess (and this is a guess), is that most filter functions is going to look like this:

````elisp
(defun my-custom-tide-filter (navto-items)
  (-filter
  ;; my custom lambda/predicate
   (lambda (navto-item)
    (member (plist-get navto-item :kind) '("class" "function" "interface" "type" "enum"))
   navto-items))
````

Which leaves a bit of overhead/redundancy which could be avoided if we instead made the defcustom a *predicate*-function which we invoked like this:

````elisp
(setq navto-items
  (-filter #'my-custom-tide-filter
    (lambda (navto-item) (funcall tide-navto-item-filter navto-item))
    navto-items))
````

Then the filter wouldn't need all that repeated boilerplate and merely be the filtering predicate:

````elisp
(defcustom my-custom-tide-filter (item)
    (member (plist-get item :kind) '("class" "function" "interface" "type" "enum"))
````

Given that the amount of overhead per person is going to be minimal (I don't assume many people will write filters, and the amount of filters they write will probably be few), and that this implementation gives less power to the customization (it can't change the content of the list, etc)...

I don't necessarily think it's a change for something universally better... But it's a different way to solve it, so might as well discuss it :smile: 

That said... Unless anyone is strongly in favour of such a solution instead, I'm perfectly happy with the PR as it is now.

What do you guys think?

sandersn:
I realise performance isn't much of a consideration at this level of the editor, but navto-items can be a very large list, so I like the ability to make my filter function `(lambda (x) x)` instead of making my predicate `(lambda (x) t)`. That way I can avoid copying the list when not filtering at all.

And of course @josteink you are right that few people will need to write this code, so parsimony isn't worth *that* much.

josteink:
I see no objections, so consider this merged!

Issue #255, opened 2018-05-18
C-u C-u M-. to cause tide-jump-to-definition in new window irrespective or reuse-window variable


invokesus:
## Checklist

- [X] I have searched both open and closed issues and cannot find a duplicate.

I don't think currently there any way to jump to definition` in a new window if `tide-jump-to-definition-reuse-window` is true.

And since the prefix argument is used to jumping to type definition, I would like to add a feature where `C-u C-u M-.` jumps to definition in a new window.

I have `reuse-window` set to `t` since I often need just a glance at the definition but sometimes I need the definition open in a window beside the current one I am working on. This modification will help in that use case.

invokesus:
Another variation could be for `C-u C-u M-.` to perform the opposite behavior of what `reuse-window` variable states.

ananthakumaran:
I will wait for input from others, but my opinion is, this goes into the too specific feature area. Not sure how many people would use it. Also this could be easily solved by

```elisp
(defun my-jump ()
  (let ((tide-jump-to-definition-reuse-window (my-reuse-logic))
    (tide-jump-to-definition))
```

josteink:
I don’t really have an opinion on this either way.

Does other major or minor-modes have any conventions for this kind of navigation and frame management?

invokesus:
[Elpy](https://elpy.readthedocs.io/en/latest/ide.html#navigation) has quite a few.

ananthakumaran:
Should we consider add support for `xref`, then we don't have to worry about the UX and it would be consistent with other packages. Though it's still marked as experimental as of 26.1

```
;; NOTE: The xref API is still experimental and can change in major,
;; backward-incompatible ways.  Everyone is encouraged to try it, and
;; report to us any problems or use cases we hadn't anticipated, by
;; sending an email to emacs-devel, or `M-x report-emacs-bug'.
```

ananthakumaran:
I think in the short term, we can add support for xref without breaking existing functionalities. This means, if xref is available then xref-* functions will automatically start to work. If it works out, then we can consider dropping support for current navigation functionalities later. But all this depends on whether xref api works nicely with our backend etc. I haven't looked at the api closely.

josteink:
Right. Given that everything works across versions and without any need for manual intervention by the user, I'd be happy with such a change.

nrxus:
[...]

PR #278, opened 2018-08-30
tide-references: make entire line clickable

sandersn:
Note that the presentation doesn't change; only the symbol in question is highlighted, but you can now click anywhere in the line to jump to the reference location.

I made this change for two reasons:
1. I am bad at clicking things, especially small ones. 
2. I use hiwin to grey out non-active windows, which cancels tide-reference's highlights when it's not active.

However, I think this change will benefit everybody by making the click target bigger.

### DIFF
--- a/tide.el
+++ b/tide.el
@@ -1521,7 +1521,7 @@ This function is used for the basic completions sorting."
     (put-text-property start end 'face 'tide-match line-text)
     (put-text-property start end 'mouse-face 'highlight line-text)
     (put-text-property start end 'help-echo "mouse-1: Visit the reference." line-text)
-    (put-text-property start end 'tide-reference reference line-text)))
+    (put-text-property 0 (length line-text) 'tide-reference reference line-text)))
 
 (defun tide-insert-references (references)
   "Create a buffer with the given REFERENCES.


sandersn:
@lddubeau I like your idea. Two questions: Should we collapse multiple hits into one line? Do you think it's all right for 'n' and 'p' to just move from line to line in that case?

ananthakumaran:
> If we are to change this behaviour to be more in line with how other "common" Emacs-functions work, it would really be great to have it work similarly at the core, not just at the appearance-level.

We discussed about the UI change in the [past](https://github.com/ananthakumaran/tide/issues/255). If somebody interested in spending time on this area, probably look at adding support for [Xref](https://www.gnu.org/software/emacs/manual/html_node/emacs/Xref.html). Emacs comes with a default UI and other frameworks could hook into xref and customize like https://github.com/brotzeit/helm-xref, https://github.com/alexmurray/ivy-xref

lddubeau:
@sandersn Regarding your second question, with what I have in mind `n` and `p` would not change from their current behavior. Regarding your first question, I'm not sure what you are asking there. I'm going to describe what I have in mind in other words because I think @josteink and I are not on the same page as to what is being proposed here. If nothing I'm writing answers your question, I'm going to need details about what you're asking.

The change I proposed in response to the initial PR would not include any cosmetic change: the `*tide-references*` buffer would look exactly like it does now. It moreover would not change anything to existing user interactions that are already supported (so `n` and `p` would work the same as they do now). The only difference is that if someone goes into `*tide-references*` and hits the enter key somewhere else than right inside a hit, Emacs would follow the reference to the corresponding position into the source file references. (What I mean by "corresponding position" is that if searched for `foo` and the `*tide-references*` buffer has a line `   23: const bar = baz(foo) + moo(foo)` and I hit enter in `*tide-references*` while `point` is on the `+` symbol, then I should get onto the `+` symbol into the source file that contained the hit.)

The above was my initial "off the top of my head" idea. I can see arguments for some amendments to that idea. For instance, a case could be made that although `*tide-references*` *should* respond to enter being hit anywhere in a line, it should not jump between hits but should instead jump to the closest hit.

@josteink I agree that `*tide-references*` should harmonize as much as possible with what we get with `grep` or similar tools. It seems to me what @sandersn initially proposed and the change I proposed in response to the PR is one step towards that goal. When you are in a `*grep*` buffer, you can hit enter everywhere and jump to a reasonable position. Right now, you cannot do the same in `*tide-reference*`. If what I'm proposing is implemented, then `*tide-references*` will be handling enter better than `*grep*` because `*grep*` mishandles multiple hits on the same line. Using the same example I gave in an earlier comment, if I `M-x grep-find` the string `db` and hit enter on the 3rd instance of `db` in the line `*grep*` takes me right onto the 1st instance! In my view, that's a bug.

I realize the proposal here does nothing with regards to addressing bindings to `next-error`, etc.. If @sandersn wants to address that *too*, that's great. But I don't think @sandersn needs to address that in this PR. And I don't see anything proposed here as somehow impeding the changes you mentioned.

josteink:
> When you are in a *grep* buffer, you can hit enter everywhere and jump to a reasonable position. Right now, you cannot do the same in *tide-reference*

Agreed. And I think that should be fixed too.

> If what I'm proposing is implemented, then *tide-references* will be handling enter better than *grep* because *grep* mishandles multiple hits on the same line. 

Which is good.

> I realize the proposal here does nothing with regards to addressing bindings to next-error

Yeah. That's just me wishing. I don't strictly see it as a *requirement*, but if we could get it working while redoing this porting of the code, I just thought that would be *nice*, because then it would be even more in line with how grep, occur and similar work in Emacs.

elibarzilay:
I ran into this annoyance too, and some more; and @lddubeau's idea seemed appealing so I implemented it.  It doesn't change `n`/`p`, and doesn't implement a `next-error` thing, and it doesn't change any of the existing code, but it works fine in a way that sucks less:

```emacs-lisp
(defun tide-goto-reference-line ()
  "Jump to reference location in the file."
  (interactive)
  (-when-let* ((bol  (line-beginning-position))
               (line (1+ (next-single-property-change bol 'face))) ; after the ":"
               (col  (max 0 (- (point) line)))
               (ref  (next-single-property-change bol 'tide-reference))
               (ref* (get-text-property ref 'tide-reference)))
    (tide-jump-to-filespan ref* nil t)
    (goto-char (+ (line-beginning-position) col))))
```

Making this a proper PR shouldn't take too much: make it put text properties that are more reliable than relying on where the `face` changes (to find the text BOL point), or even better -- make the line number displayed without being actual text so it's not counted as contents of the refs buffer (and you can't navigate into it).

[I also ran into more related annoyces: a bug in `p` (use it from inside an identifier to see it), the lack of `TAB` binding, and not cycling matches, so I fixed the bug and made another function that cycles through the refs to bind to `TAB`.]

josteink:
> Making this a proper PR shouldn't take too much: make it put text properties that are more reliable than relying on where the face changes (to find the text BOL point), or even better -- make the line number displayed without being actual text so it's not counted as contents of the refs buffer (and you can't navigate into it).

Please do! 😃

PR #279, opened 2018-08-31
WIP: server management

lddubeau:
I'm opening a PR so that this can be discussed before it is merged. Or so that it can be objected to, if there are objections.

The explosion of `tsserver` processes has been a bit of an issue for me. I can do `ps` at the command line but that's rather limited in usefulness. I can also use `M-x list-processes` but it shows things I don't care about and conversely does not readily provide information I care about when I'm inspecting `tsserver` processes. So I've worked on a `tide-list-servers` command which can help inspect the instance that tide has started. I've derived it from how `list-processes` does its job.

Things that can be done with it:

1. See the servers running.
2. Kill a server by hitting `d` in the `*Tide Server List*` buffer.
3. Verify the configuration of a server by going on its project name and hitting enter.

This is something I've just put up together, which I'm sure could be improved. Feedback welcomed.

### DIFF
--- a/tide.el
+++ b/tide.el
@@ -2384,18 +2384,8 @@ timeout."
                     project-name ,project-name
                     action tide--list-servers-verify-setup))
                  (name (process-name p))
-                 (buf-label (if (buffer-live-p buf)
-                                `(,(buffer-name buf)
-                                  face link
-                                  help-echo ,(format-message
-                                              "Visit buffer `%s'"
-                                              (buffer-name buf))
-                                  follow-link t
-                                  process-buffer ,buf
-                                  action process-menu-visit-buffer)
-                              "--"))
                  (cmd (mapconcat 'identity (process-command p) " ")))
-            (push (list p (vector project-name-label name status buf-label cmd))
+            (push (list p (vector project-name-label name status cmd))
                   tabulated-list-entries)))))))
 
 (defun tide--server-list-kill-server ()
@@ -2417,7 +2407,6 @@ timeout."
   (setq tabulated-list-format [("Project Name" 15 t)
                                ("Process" 15 t)
 			       ("Status"   7 t)
-			       ("Buffer"  15 t)
 			       ("Command"  0 t)])
   (setq tabulated-list-sort-key (cons "Project Name" nil))
   (add-hook 'tabulated-list-revert-hook 'tide--list-servers-refresh nil t)
--- a/tide.el
+++ b/tide.el
@@ -396,15 +396,14 @@ ones and overrule settings in the other lists."
 
 (defun tide-any-buffer (project-name fn)
   "Callback FN for the first buffer within PROJECT-NAME with tide-mode enabled."
-  (let* ((buffer (-any (lambda (buffer)
-                         (with-current-buffer buffer
-                           (when (and (bound-and-true-p tide-mode)
-                                      (equal (tide-project-name) project-name))
-                             buffer)))
-                       (buffer-list))))
-    (when buffer
-      (with-current-buffer buffer
-        (funcall fn)))))
+  (-when-let (buffer (-any (lambda (buffer)
+                             (with-current-buffer buffer
+                               (when (and (bound-and-true-p tide-mode)
+                                          (equal (tide-project-name) project-name))
+                                 buffer)))
+                           (buffer-list)))
+    (with-current-buffer buffer
+      (funcall fn))))
 
 (defun tide-line-number-at-pos (&optional pos)
   (let ((p (or pos (point))))
--- a/tide.el
+++ b/tide.el
@@ -38,6 +38,7 @@
 (require 'imenu)
 (require 'thingatpt)
 (require 'tide-lv)
+(require 'tabulated-list)
 
 ;; Silence compiler warnings
 
@@ -392,6 +393,18 @@ ones and overrule settings in the other lists."
         (when (and (bound-and-true-p tide-mode)
                    (equal (tide-project-name) project-name))
           (funcall fn))))))
+
+(defun tide-any-buffer (project-name fn)
+  "Callback FN for the first buffer within PROJECT-NAME with tide-mode enabled."
+  (let* ((buffer (-any (lambda (buffer)
+                         (with-current-buffer buffer
+                           (when (and (bound-and-true-p tide-mode)
+                                      (equal (tide-project-name) project-name))
+                             buffer)))
+                       (buffer-list))))
+    (when buffer
+      (with-current-buffer buffer
+        (funcall fn)))))
 
 (defun tide-line-number-at-pos (&optional pos)
   (let ((p (or pos (point))))
@@ -2340,6 +2353,89 @@ timeout."
     (delete-process server))
   (tide-start-server)
   (tide-each-buffer (tide-project-name) #'tide-configure-buffer))
+
+(defun tide--list-servers-verify-setup (button)
+  "Invoke `tide-verify-setup' on a tsserver displayed in the list of server."
+  (tide-any-buffer (button-get button 'project-name) #'tide-verify-setup))
+
+;; This is modeled after list-process--refresh but we do not call delete-process
+;; on exited or signaled processe. That seems inappropriate for a function
+;; designed to *report* information.
+(defun tide--list-servers-refresh ()
+  "Recompute the list of processes for the buffer displayed by
+`tide-list-servers'."
+  (setq tabulated-list-entries nil)
+  (let* ((tsservers (hash-table-values tide-servers)))
+    (dolist (p (process-list))
+      (let* ((project-name (process-get p 'project-name)))
+        ;; Testing for the presence of project-name is a quick way to eliminate
+        ;; processes we don't care about: if it is absent, then the process is
+        ;; definitely not a tide process. If present, the the process could
+        ;; still be from another mode which happens to use the same property. So
+        ;; we also check against the list of servers.
+        (when (and project-name (memq p tsservers))
+          (let* ((buf (process-buffer p))
+                 (status (symbol-name (process-status p)))
+                 (project-name-label
+                  `(,project-name
+                    face link
+                    help-echo ,(format-message "Verify setup of `%s'"
+                                               project-name)
+                    follow-link t
+                    project-name ,project-name
+                    action tide--list-servers-verify-setup))
+                 (name (process-name p))
+                 (buf-label (if (buffer-live-p buf)
+                                `(,(buffer-name buf)
+                                  face link
+                                  help-echo ,(format-message
+                                              "Visit buffer `%s'"
+                                              (buffer-name buf))
+                                  follow-link t
+                                  process-buffer ,buf
+                                  action process-menu-visit-buffer)
+                              "--"))
+                 (cmd (mapconcat 'identity (process-command p) " ")))
+            (push (list p (vector project-name-label name status buf-label cmd))
+                  tabulated-list-entries)))))))
+
+(defun tide--server-list-kill-server ()
+  "Kill a tsserver instance."
+  (interactive)
+  (let* ((process (tabulated-list-get-id)))
+    (delete-process process)
+    (tide-each-buffer (process-get process 'project-name)
+                      (lambda () (tide-cleanup-buffer)))
+    (revert-buffer)))
+
+(defvar tide-server-list-mode-map
+  (let ((map (make-sparse-keymap)))
+    (define-key map [?d] 'tide--server-list-kill-server)
+    map))
+
+(define-derived-mode tide-server-list-mode tabulated-list-mode "tide-server-list-mode"
+  "Major mode for listing tsserver processes."
+  (setq tabulated-list-format [("Project Name" 15 t)
+                               ("Process" 15 t)
+			       ("Status"   7 t)
+			       ("Buffer"  15 t)
+			       ("Command"  0 t)])
+  (setq tabulated-list-sort-key (cons "Project Name" nil))
+  (add-hook 'tabulated-list-revert-hook 'tide--list-servers-refresh nil t)
+  (tabulated-list-init-header))
+
+;; This is modeled after list-processes.
+(defun tide-list-servers (&optional buffer)
+  "Lists the tsserver processes known to tide."
+  (interactive)
+  (unless (bufferp buffer)
+    (setq buffer (get-buffer-create "*Tide Server List*")))
+  (with-current-buffer buffer
+    (tide-server-list-mode)
+    (tide--list-servers-refresh)
+    (tabulated-list-print))
+  (display-buffer buffer)
+  nil)
 
 (defun tide-command:status ()
   (tide-send-command-sync "status" '()))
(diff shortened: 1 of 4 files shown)


tide.el
```diff
@@ -389,6 +390,18 @@ ones and overrule settings in the other lists."
                    (equal (tide-project-name) project-name))
           (funcall fn))))))
 
+(defun tide-any-buffer (project-name fn)
+  "Callback FN for the first buffer within PROJECT-NAME with tide-mode enabled."
+  (let* ((buffer (-any (lambda (buffer)
```
ananthakumaran: -when-let could be used here

tide.el
```diff
@@ -2310,6 +2323,89 @@ timeout."
...
+  (let ((map (make-sparse-keymap)))
+    (define-key map [?d] 'tide--server-list-kill-server)
+    map))
+
+(define-derived-mode tide-server-list-mode tabulated-list-mode "tide-server-list-mode"
+  "Major mode for listing tsserver processes."
+  (setq tabulated-list-format [("Project Name" 15 t)
+                               ("Process" 15 t)
+			       ("Status"   7 t)
+			       ("Buffer"  15 t)
```
ananthakumaran: any specific reason for listing Process and Buffer. I can't think of any useful info out of both of these. Showing project root directory might be useful in cases where there are many projects with same

tide.el
```diff
@@ -2310,6 +2323,89 @@ timeout."
...
+            (push (list p (vector project-name-label name status buf-label cmd))
+                  tabulated-list-entries)))))))
+
+(defun tide--server-list-kill-server ()
+  "Kill a tsserver instance."
+  (interactive)
+  (let* ((process (tabulated-list-get-id)))
+    (delete-process process)
+    (tide-each-buffer (process-get process 'project-name)
+                      (lambda () (tide-cleanup-buffer)))
```
ananthakumaran: I am not sure about calling from here, it also tries to call tsserver `tide-command:closefile`, but the process is already dead. Perhaps the tmp-file cleanup should be moved to `tide-cleanup-project` method? I would prefer the cleanup done in a single place

tide.el
lddubeau: Both columns were directly inherited from `list-processes`. I was thinking about removing "Process" and "Buffer" but I was a bit concerned about missing some use-case scenarios for them.

Someone who wants to do something not supported by `tide-list-servers` may want to have the process name in order to recover the process. However, this information could be obtainable through a key binding rather than through a column.

Similarly if there's a reason to ever go to a process' buffer then that could be provided by a key binding rather than a column. Or I could just not bother with providing any such facility.

And I agree with adding a root directory column.

tide.el
lddubeau: Ouch! This is a mistake. I meant to call `tide-cleanup-project`. I must have tripped over auto-completion and not noticed.

lddubeau: You're right. There's no need for cleaning up there.

lddubeau:
I've just pushed my latest experiments. I've reduce the columns shown to:

* Project name
* CPU (cpu usage)
* Either the project root *or* the command line.

Showing the project root and command line together is not easy to do when the paths are long (which is my case and I suspect is the case for most people). The table framework does not do well when more than one column in a row has to hold long values. I never got it to work nicely. So I've added a keybinding (`/`) to switch what is shown in the last column.

josteink:
Just tested it and it looks nice. I can't seem to get project-root to list correctly though.

Anything obvious I'm doing wrong?

lddubeau:
@josteink I can think of one thing. If Emacs was already running, and you got the PR's code, eval'ed the buffer, and then tried to list processes, then project-root won't be right.  The project root associated with a process is equal to the process' current directory but I did not find a function like `(get-cwd process)`. I suppose the process' buffer has its `default-directory` set to the process' current directory but switching through buffers to get those values does not appeal to me. So the PR currently adds a `project-root` field to the processes.

So if you use the PR's code with processes that were created before the PR, then their project-root won't be right. The processes have to be created with the PR's code to display right.

That's the only thing I can think of that would explain problems with project-root.

josteink:
Yeah. After restarting Emacs and loading tide before any other typescript-project it behaves like you said it should. So from a end-user point of view, this looks like a nice addition. Definitely thumbs up!

I'll try to look over the code and provide some feedback on that too.

josteink:
Yeah. After restarting Emacs and loading tide before any other typescript-project it behaves like you said it should. So from a end-user point of view, this looks like a nice addition. Definitely thumbs up!

I'll try to look over the code and provide some feedback on that too.

josteink:
Yeah. After restarting Emacs and loading tide before any other typescript-project it behaves like you said it should. So from a end-user point of view, this looks like a nice addition. Definitely thumbs up!

I'll try to look over the code and provide some feedback on that too.

tide.el
```diff
@@ -389,6 +390,17 @@ ones and overrule settings in the other lists."
                    (equal (tide-project-name) project-name))
           (funcall fn))))))
 
+(defun tide-any-buffer (project-name fn)
+  "Callback FN for the first buffer within PROJECT-NAME with tide-mode enabled."
+  (-when-let (buffer (-any (lambda (buffer)
```
josteink: This may be nitpicking terminology but I dislike the use of `any` here, and find it slightly confusing, forcing me to read the code to ensure that the implementation actually matches the doc-string.

From my previous experience `any` can be used like a predicate with a filter function:

````c#
if (collection.Any(i => !i.Deleted))
{
    // do stuff
}
````

In this case it represents a `collection -> bool`-operation, and that is OK. While the implementation basically returns `true` after the first match is found, there's nothing conceptually about "any" which ties it to the conceptual notion of "first".

I find "any" when used as a function which invokes an *action* (like in your code) to immediately suspect it might invoke this function for *every* buffer it finds with that project. Basically the semantics of "any" is a bit too vague for my taste.

I think this code would be a lot clearer if we separated the retrieval of the buffer from the invocation upon it.

And I'd prefer if the retrieval of the buffer was in a function which used a phrase matching the intent, like `tide-get-first-buffer-for-project` or something.

Nitpick over. Press any key to continue :)

**Comment by josteink** [comment]
I just tried restarting Emacs and loading tide before any typescript-project. In that case it behaved like you said it should.

I think this is a nice addition to tide, so definitely thumbs up.

I'll try to provide some review on the code too, but I'll do that separately.

**Comment by josteink** [comment]
I just tried restarting Emacs and loading tide before any typescript-project. In that case it behaved like you said it should.

I think this is a nice addition to tide, so definitely thumbs up.

I'll try to provide some review on the code too, but I'll do that separately.

**Comment by josteink** [comment]
I just tried restarting Emacs and loading tide before any typescript-project. In that case it behaved like you said it should.

I think this is a nice addition to tide, so definitely thumbs up.

I'll try to provide some review on the code too, but I'll do that separately.

**Comment by josteink** [comment]
I just tried restarting Emacs and loading tide before any typescript-project. In that case it behaved like you said it should.

I think this is a nice addition to tide, so definitely thumbs up.

I'll try to provide some review on the code too, but I'll do that separately.

**Comment by josteink** [comment]
I just tried restarting Emacs and loading tide before any typescript-project. In that case it behaved like you said it should.

I think this is a nice addition to tide, so definitely thumbs up.

I'll try to provide some review on the code too, but I'll do that separately.

File: tide.el [hunk 5]
```diff
@@ -2310,6 +2323,105 @@ timeout."
   (tide-start-server)
   (tide-each-buffer (tide-project-name) #'tide-configure-buffer))
 
+(defun tide--list-servers-verify-setup (button)
+  "Invoke `tide-verify-setup' on a tsserver displayed in the list of server."
+  (tide-any-buffer (button-get button 'project-name) #'tide-verify-setup))
+
```
josteink: I can't seem to find the `tide-verify-setup` function anywhere. Is this code you forgot to clean out? Or add?

tide.el
```diff
@@ -2310,6 +2323,105 @@ timeout."
...
+(defun tide--server-list-cycle-last-column ()
+  "Cycle through the possible values for the last column."
+  (interactive)
+  (setq tide--server-list-mode-last-column
+        (or (cadr (or (memq tide--server-list-mode-last-column
+                            tide--server-list-last-column-choice-list)
+                      (error "%s is not a possible choice of last column."
+                             tide--server-list-mode-last-column)))
+            (car tide--server-list-last-column-choice-list)))
```
josteink: I think this whole `setq` is a bit more intricate than it needs to ne.

Also: Why do we need error-handling here? Are we afraid someone will manually override the `defvar` to an invalid state? Have you encountered errors here yourself during development since you added this?

tide.el
```diff
@@ -2310,6 +2323,105 @@ timeout."
...
+  (tide--setup-list-mode)
+  (revert-buffer))
+
+(defvar tide-server-list-mode-map
+  (let ((map (make-sparse-keymap)))
+    (define-key map [?d] 'tide--server-list-kill-server)
+    (define-key map [?/] 'tide--server-list-cycle-last-column)
+    map))
```
josteink: Should we also add some basic key for navigation? Like next and previous?

Or better yet... Inherit the keymap from the mode we're deriving from? Like setting these things using `(define-key)` directly on `tide-server-list-mode-map` after the derived mode has been created?

If this code results in the same behaviour, it's certainly not obvious.

tide.el
lddubeau: The meaning of  `any` ultimately comes from `dash`. It also provides `-some` as an alias to `-any` but in my view `some` is not clearer than `any`.

I have to say I'm not super familiar with dash. I'm not sure why I landed on `-any` first, but I found there's a `-first` function which is a better fit. So I'm definitely going to replace `-any` with `-first` and rename the tide function accordingly. (But if we want to nitpick there's material for nitpicking "first" too. When *I* read "first", it implies to me some sort of *definite and guaranteed* order between the elements of the list. First alphabetically, first numerically, first by age, etc. Here though the order is not definitely not guaranteed.)

`tide-any-buffer` is modeled after the preexisting tide function `tide-each-buffer`, which appears just before it. `tide-each-buffer` applies a function to all buffers of a specific project, whereas `tide-any-buffer` applies a function to any buffer of a specific project, but only one. Whatever objections you have against the new function apply just as well to its older sibling.

So if we're going to do something, both functions should be renamed:

`tide-each-buffer` -> `tide-apply-to-each-buffer-in-project`

`tide-any-buffer` -> `tide-apply-to-first-buffer-in-project`

I disagree about separating retrieval from invocation. *If* that's needed, *then* the code can be refactored.

tide.el
lddubeau: That's a preexisting function:

https://github.com/ananthakumaran/tide/blob/46424cae4a0d49981dce575bdfca19fedf953a58/tide.el#L2440

tide.el
lddubeau: Regarding, your first sentence, I'm not seeing the unnecessary intricacy.

Regarding your "also", when I write code that is supposed to do something for *every possible value of a finite set of values*, I always include a branch executed if no other branch is executed. This additional branch raises a explicit error. It helps trap bugs early. Not explicitly and immediately raising an error usually results in either:

1. An error message raised elsewhere which is definitely more obscure than the custom error I raise. I'm not sure here what would happen but I suspect some code down the line would complain about a `nil` value. Tracing back where the `nil` came from is time consuming.

2. A coding problem goes unnoticed, even in testing. (Yes, I've seen it happen.)

tide.el
lddubeau: > If this code results in the same behaviour, it's certainly not obvious.

Huh? `define-derived-mode` sets the parent of the keymap of the child mode to be the keymap of the parent mode. This is exactly as "not obvious" as just about all other modes created in tide or typescript-mode.

And thus there are already "next" and "previous" keybindings available for you to use, inherited from `tabulated-list-mode`.

tide.el
josteink: OK. If that's the intent, I'm all for it. Catching bugs in their infancy is IMO a good thing.

tide.el
josteink: If so my confusion steams from not settings directly on the derived mode's map, but instead using `(make-sparse-keymap)`.

That makes a *new* map, right?

lddubeau: Yes, `make-sparse-keymap` make a new map. Then `define-derived-mode` sets the parent of this new map to be the keymap of the parent mode.

lddubeau:
@josteink I think I've covered everything we discussed.

josteink:
Agreed. The CI for the PR went red though, so I queued a new build.

If everything is green after that, feel free to merge to master :)

lddubeau:
I'd like to have some tests in the suite to cover the new capabilities. It would mean though that the suite would start `tsserver` processes. I believe this would be the first time that the suite incorporates code that actually starts `tsserver` processes. 

So before I cross that bridge, any objections?

josteink:
Absolutely none. Sounds great to me!

lddubeau:
Good thing I added tests. The PR is failing the test suite because I'm using functions that did not exist in the earlier versions of Emacs we're still testing for.

I've already made some modifications to remedy this, but before I go further down that road, are there versions of Emacs that are old enough that it is not worth our time making tide work on them?

josteink:
If you were to have my vote, from a practical point of view, it would be ok to obsolete anything older than what’s shipped in Ubuntu 16.04 LTS. 

That’s 24.5, right?

lddubeau:
As far as I can tell, yes, Ubuntu 16.04 LTS ships 24.5. I don't use Ubuntu though. I'm on Debian. Debian stable (aka Stretch, aka Debian 9) ships 24.5.

I'd be totally fine with limiting the oldest version supported to 24.5.

lddubeau:
The suite finally passes on all versions of Emacs that are supported. Yay! There are a few things to highlight in the latest commits, however.

1. 3550ff5 deals with cases where the CPU usage of a process is reported by Emacs as NaN. I ran into that case after switching to a newer laptop. I have no idea what may cause NaN to be reported and whether it has any special meaning, or it is just a bug in Emacs. At any rate, getting NaN caused the code to crash. I've decided to represent it as `--` in the process list, and when sorting, it is considered to be less than any number.

2. c6731e9 adds the tests. As far as I can recall, this is the first time I write an asynchronous test for Emacs. Here, I'm polling for the condition I'm testing for. Usually, I prefer listening to events, but I'm not familiar with the logic I'd have to tap into, and I cannot budget time to figure it out.

3. a2b2926 deals with the code that does not work on older versions of Emacs. The strategy I've used is to test whether the new functions are actually present, and use these functions when possible. I would have been possible to just handle older versions by replacing the problematic code with just the else branch of the condition. So, for instance, instead of this:

   ```
             (cpu
              ;; alist-get is missing from Emacs prior to version 25. This code should be
              ;; simplified to just call alist-get once we no longer support versions that
              ;; lack it.
              (if (functionp 'alist-get)
                  (alist-get 'pcpu (process-attributes pid))
                (cdr (assq 'pcpu (process-attributes pid))))))
   ```

   I could have just this, which works on all the versions we support:

   ```
             (cpu (cdr (assq 'pcpu (process-attributes pid))))
   ```

   However, I much much prefer `(alist-get ...)` over `(cdr (assq ...))` and I'd like to get rid of the latter as early as I can. With the comment and test, the fact that we can get eventually just use `alist-get` is documented, and the way forward is clear and tested.

lddubeau:
@josteink I think the last two commits I just added cover what you had suggested.

josteink:
Yeah that looks much better.

Unless you have anything more you'd like to add, I say go ahead and merge.

josteink:
After starting with this in August 2018, it's nice to finally have this one landed. 😄 

Good job!

lddubeau:
I'm glad it is done. I've been using the WIP since I opened the pull request.

PR #284, opened 2018-10-24
fetch suggestions

### LINKED ISSUE #283: JSDoc to type annotation codefix not offered
## Checklist

- [x] I am 0.8 certain the problem does not belong to `typescript-mode` or `tsserver`. Tests all pass for my local build, but I haven't looked at the tsserver log after requesting codefixes.

## Relevant Version Numbers

- Tide: tide-20180608.239 and tide-20180810.403
- TypeScript: 3.1 and recent builds of 3.2
- Emacs: 25.2.2

## Steps to Reproduce the Bug

Start with a TS file that has JSDoc types:

```ts
// @Filename: welove.ts
/**
 * @param {String | Number} value
 */
export const f = value => new Number(value)
```

with a tsconfig that has noImplicitAny on:

```json
{
    "compilerOptions": {
        "strict": true,
        "noImplicitAny": true,
        "target": "es5",
        "lib": ["es5", "es2015", "es2016", "es2017", "dom"],
        "allowJs": true,
        "checkJs": true,
        "outDir": "out",
        "jsx": "preserve",
        "esModuleInterop": true,
        "module": "commonjs"
    },
    "include": [
        "ex.d.ts",
        "mod1.js",
        "test.js",
        "welove.ts",
        "first.ts",
        "forever.js"
    ]
}
```

## Expected behavior:

Implicit any error on `value`, plus a codefix that replaces it with `value: string | number`

## Actual behavior:

Error, but no codefix offered.

This codefix works fine on VS Code, and other codefixes work fine, such as creating property declarations. I'm not sure what is different about this one.

Edit: related issue https://github.com/Microsoft/TypeScript/issues/28084

ananthakumaran:
fixes #283 

Suggestions from tsserver are shown as flycheck **info** messages. This fix also enables other types of codefixes like convert to an async function, value is never read etc.

screen shot 2018-10-24 at 9 38 16 am

### DIFF
--- a/tide.el
+++ b/tide.el
@@ -1905,7 +1905,8 @@ code-analysis."
                           (setq resolved t)
                           (funcall cb err))
                       (when (and (plist-member result :syntaxDiag)
-                                 (plist-member result :semanticDiag))
+                                 (plist-member result :semanticDiag)
+                                 (plist-member result :suggestionDiag))
                         (setq resolved t)
                         (funcall cb `(:body (,result) :success t)))))))
       (tide-send-command
@@ -1923,6 +1924,18 @@ code-analysis."
          (if (tide-response-success-p response)
              (setq result (plist-put result :semanticDiag (plist-get response :body)))
            (setq err response))
+         (resolve)))
+
+      (tide-send-command
+       "suggestionDiagnosticsSync"
+       `(:file ,(tide-buffer-file-name))
+       (lambda (response)
+         (cond
+          ((tide-response-success-p response)
+           (setq result (plist-put result :suggestionDiag (plist-get response :body))))
+          ((tide-command-unknown-p response)
+           (setq result (plist-put result :suggestionDiag '())))
+          (t (setq err response)))
          (resolve))))))
 
 (defun tide-parse-error (response checker)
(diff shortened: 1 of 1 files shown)


tide.el
```diff
@@ -2020,13 +2035,16 @@ code-analysis."
     (let ((inhibit-read-only t))
       (erase-buffer))
     (display-buffer (current-buffer) t)
-    (let* ((project-files (-filter (lambda (file-name)
-                                     (not (string-match-p "node_modules/typescript/" file-name)))
+    (let* ((project-files (-reject (lambda (file-name)
+                                     (or (string-match-p "node_modules/typescript/" file-name)
+                                         (string-match-p "tsconfig.json$" file-name)))
```
josteink: Any reason we need to add this additional filter? It's not obvious from the context.

josteink:
This is a really nice addition!

Tested and looks good to me.

tide.el
ananthakumaran: The file list returned by tsserver also contains tsconfig.json -- not sure when this change was introduced. The callback was waiting for diagnostics results of tsconfig.json which would never arrive.

josteink: Right. Not sure if that needs a comment or not.

Otherwise, the changes looks fine, and I'm happy about the improvements.

PR #289, opened 2018-11-23
sort the edits in descending order

### LINKED ISSUE #288: tide-fix and typescript-tslint-plugin

## Checklist

- [x] If [tide *is reporting*](https://github.com/ananthakumaran/tide/wiki/Explanations-about-the-Issue-Reporting-Checklist#the-behavior-of-cli-tools-agrees-with-your-expectations) an error or warning I think *should not be reported*, I can run `tsc` (and `tslint`, if applicable) without the error or warning I'm seeing in tide.
- [x] If [tide is *not reporting*](https://github.com/ananthakumaran/tide/wiki/Explanations-about-the-Issue-Reporting-Checklist#the-behavior-of-cli-tools-agrees-with-your-expectations) an error or warning I think *should be reported*, `tsc` (or `tslint`, if applicable) reports the error or warning I was expecting to see.
- [x] I am [positive](https://github.com/ananthakumaran/tide/wiki/Explanations-about-the-Issue-Reporting-Checklist#how-to-determine-which-project-the-problem-belongs-to) the problem does not belong to `typescript-mode` or `tsserver`.

## Relevant Version Numbers

- Tide: 20181025.1201
- TypeScript: 3.1.6
- Emacs: 26.1

## Steps to Reproduce the Bug

I have been using this [plugin](https://github.com/Microsoft/typescript-tslint-plugin) to view TSLint violations in different editors, and for me it works better with a project over Flycheck. I noticed that there is a `tide-fix` command and when I run it I do get options to fix the command under the cursor.

screen shot 2018-11-22 at 9 07 30 pm

If I select `a` or `s` then the fixes work as expected.

screen shot 2018-11-22 at 9 07 40 pm

However, in the scenario below where I have multiple violations:

screen shot 2018-11-22 at 9 09 29 pm

When I try to fix violations I get unexpected results:

screen shot 2018-11-22 at 9 09 39 pm

I have not gotten this issue with other editors so I can't say it's necessarily it's because of the typescript-tslint-plugin. I'm suspecting that tide own formatting fixes could be interfering.

Also, it would be nice if I can invoke these fixes without having to move my cursor under an error in question.

ananthakumaran:
fixes #288

> Applying the edit instructions in reverse to file will result in correctly reformatted text.

One of the comments in TypeScript repo says the edits should be applied
in reverse order. But, I am not sure all the plugins follows this? Doing
a manual sort does seem to fix the tslint issue. VS code applies the fixes
correctly, but I can't able to find the logic. Atom[1] also manually
sorts the edits.

1: https://github.com/TypeStrong/atom-typescript/blob/master/lib/main/pluginManager.ts#L349-L352

### DIFF
--- a/tide.el
+++ b/tide.el
@@ -1776,7 +1776,13 @@ code-analysis."
 (defun tide-apply-edits (edits)
   (save-excursion
     (-map (lambda (edit) (tide-apply-edit edit))
-          (nreverse edits))))
+          (-sort
+           (tide-compose-comparators
+            (lambda (a b)
+              (< (tide-plist-get b :start :line) (tide-plist-get a :start :line)))
+            (lambda (a b)
+              (< (tide-plist-get b :start :offset) (tide-plist-get a :start :offset))))
+           (nreverse edits)))))
 
 (defun tide-format-region (start end)
   (let ((response (tide-send-command-sync


ananthakumaran:
@jupl could you try this branch?

jupl:
I can still replicate the issue unfortunately. :(

jupl:
https://github.com/jupl/tide-thingy

I can fix the `var` keyword with `tide-fix`, but attempting to fix curly spacing with `s` or `d` gets wacky. It works as expected in VSCode.

ananthakumaran:
Thanks for the example, it was helpful in debugging. I have added two more fixes
1) fixes for same file was sent as two different code-edits, I have to do a group by filename to fix this
2) we apply format-region on the region where the new code-edits are applied. This was a hack to prevent code-edit messing up the format, but in your case, tide-format was adding the spaces back. I have removed the tide-format.

I am kind of not sure about merging these changes. AFAIK, these things are not thoroughly documented anywhere, so I am kind of not sure what is the correct fix and who is at fault here.

ananthakumaran:
```typescript
    /** Apply each textChange in order, updating future changes to account for the text offset of previous changes. */
    function forEachTextChange(changes: ReadonlyArray<ts.TextChange>, cb: (change: ts.TextChange) => void): void {
        // Copy this so we don't ruin someone else's copy
        changes = JSON.parse(JSON.stringify(changes));
        for (let i = 0; i < changes.length; i++) {
            const change = changes[i];
            cb(change);
            const changeDelta = change.newText.length - change.span.length;
            for (let j = i + 1; j < changes.length; j++) {
                if (changes[j].span.start >= change.span.start) {
                    changes[j].span.start += changeDelta;
                }
            }
        }
    }
```

from this logic it looks like applying the changes in reverse order should work, as there won't be any more span after the current one.

ananthakumaran:
@jupl could you test it once more

PR #290, opened 2018-12-12
upgrade tsserver to 3.2.2

hateonion:
Typescript is released to 3.2.2 5 days ago. I just copy all the stuff under typescript packages lib folder and replace the existing old one in tide/tsserver folder.

Verified it's working for

Emacs: GNU Emacs 26.1 macOS

Hope I made something helpful.


josteink:
These days (by using the project's own `tsserver` when available) we are less reliant on a `tide`-shipped version of `tsserver`. That said, we *do* ship one and it might as well be current.

I believe @ananthakumaran automated and standardized the process of updating `tsserver`, so that we should be getting consistent results. He did so through make:

https://github.com/ananthakumaran/tide/blob/7c5ee3052153b6ab6e56bba569f792101725e307/Makefile#L25-L28

He also wrote a short tutorial on how to use that make recipe here: https://github.com/ananthakumaran/tide/issues/114#issuecomment-297923493

Can you verify that your current PR gives the same output/results?

If so, I'll be happy to merge.

hateonion:
Hi @josteink 

I checked out the Typescript repo to branch release-3.2(cause the master branch is for development, and the tsserver has bumped to 3.3.0-dev already on master).
Then running make update-tsserver and I got the latest change committed. 
Plz take a look.

josteink:
Thanks for the patch. Feel free to send more patches in the future! 😄

PR #293, opened 2018-12-27
Formalize the range of Emacs versions supported

lddubeau:
Taking a page out of [magit](https://github.com/magit/magit), this PR adds code so that tide formally checks whether the Emacs it is running on is at least as new as the oldest version supported.

When the check fails, the user gets a warning. This is what magit does. I've not spent any time figuring out the pros and cons of showing a warning like magit does vs slamming the brakes with a hard failure.

This also removes 24.3 from testing, as 24.4 is the oldest version supported.

### DIFF
--- a/tide-tests.el
+++ b/tide-tests.el
@@ -126,8 +126,8 @@
   (should-error (tide-load-tsconfig "test/notfound.json" '())))
 
 ;; Adapted from jdee-mode's test suite.
-(defmacro test-with-temp-buffer (content &rest body)
-  "Fill a temporary buffer with `CONTENT' and eval `BODY' in it."
+(defmacro mode-with-temp-buffer (content &rest body)
+  "Fill a temporary buffer with `CONTENT', turn on `tide-mode' and eval `BODY' in it."
   (declare (debug t)
            (indent 1))
   `(with-temp-buffer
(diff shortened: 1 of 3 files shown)


tide-tests.el
```diff
@@ -126,8 +126,8 @@
   (should-error (tide-load-tsconfig "test/notfound.json" '())))
 
 ;; Adapted from jdee-mode's test suite.
-(defmacro test-with-temp-buffer (content &rest body)
-  "Fill a temporary buffer with `CONTENT' and eval `BODY' in it."
+(defmacro mode-with-temp-buffer (content &rest body)
+  "Fill a temporary buffer with `CONTENT', turn on `tide-mode' and eval `BODY' in it."
   (declare (debug t)
```
josteink: This rename doesn't really seem to add anything to the overall change. That is, I don't think the new name makes things much cleraer.

The improved docstring is nice though.

josteink:
Left a tiny nitpicky comment, but apart from that this looks good to me.

Nice to have things formalized properly. in code, not just in some random github issue discussion. :+1:

tide-tests.el
lddubeau: The rename makes sense due to the addition of [`setup-with-temp-buffer`](https://github.com/ananthakumaran/tide/pull/293/files/c5649b969e5feac3a5ca3864d4ded34205427a71#diff-12e759e9335508acde68d9ed9a19f958R141). Whereas `setup-with-temp-buffer` runs `tide-setup`, `mode-with-temp-buffer` runs `tide-mode`. 

The name `test-with-temp-buffer` was not specially problematic before `setup-with-temp-buffer` was added but after I added `setup-with-temp-buffer` it seemed to me that the name `test-with-temp-buffer` was not more naturally applicable to a macro that invokes `tide-mode` than a macro that invokes `tide-setup`: after all, both macros, allow you to "[run a] test with [a] temp[orary] buffer". So I renamed it so that for both macros, the first word of the macro name indicates what is distinctive about it relative to the other macro.

I'll address a question that was not asked, but is related to the topic here. Why two macros? `tide-mode` just turns on the Emacs mode but does not start a `tsserver`. For some tests, this is enough. On the other hand, `tide-setup` turns on the mode and does more work (including starting a `tsserver`), which some tests need. We could just invoke `tide-setup` for all tests, even those that would be satisfied by `tide-mode`, but as a matter of principle if it is easy to do so then I prefer to set the test up to use less resources. Here, I prefer not to start `tsserver` for those tests that don't need it.

josteink: Fair enough. As such, maybe the inverse naming convention would make more sense?

I.e. ‘temp-buffer-with-tide-mode’ and ‘temp-buffer-with-tide-setup’?

That would imo make the distinction more explicit.

lddubeau: I'm just not seeing it. Since you termed the issue here a "nitpick" I'm going to just go ahead and merge this.

PR #302, opened 2019-02-10
Update links to more recent format options

jeremymeng:


jeremymeng:
It took me quite a while to find out the `insertSpaceBeforeFunctionParenthesis` option. Hopefully linking to latest options saves other people's time.  I noticed that the tsserver has been updated to 3.3.1 a couple of days ago so linking to v3.3.1 tag should be good.

Issue #294, opened 2019-01-09
Allow configuring tide to avoid starting tsserver immediately


lddubeau:
This is a suggestion for a new feature.

### The Problem

I use session management to record my Emacs session and restore it when I start Emacs again. I've used this setup for years. The only time I've had a problem before was with flymake but I fixed that by switching to flycheck. However, recently tide has started giving me issues, indirectly. I have a hand in a lot of different TypeScript projects. Restoring a session has become really slow due to all the `tsserver` instances that are immediately started and compete for CPU. It has been a problem for a long time but recently it's become unbearable. Maybe TS 3 has features that cause `tsserver` to need more processing power. I don't know. I've definitely been doing things that require me to restart Emacs more often lately. Right now, when I restart Emacs I have `pkill -f tsserver` at the ready to kill all `tsserver` instances. (Fortunately, Emacs is the only tool I use that starts `tsserver`.) Then I start them manually as needed.

### The Proposal

I'd like to be able to set a configuration variable such that the variable can be set to tell `tide-setup` to skip starting `tsserver` immediately. I'd be perfectly happy with having to issue `tide-restart-server` manually (like I'm doing now) whenever I need it. I thought about starting `tsserver` lazily: whenever tide needs it and `tsserver` is absent, then tide would start it automatically, but I'd be happy even without this functionality.

I volunteer for implementing this, but if there are concerns or flat out objections I'd rather hear about them ahead of time.

josteink:
I would assume there are various kinds of invocations which I would assume only occurs on actual end-user interaction:

* completion at point
* fetching eldoc
* looking up documentation

Etc.

Maybe we can try to lazy-load tsserver (and "configure" the buffer) when these kind of functions get invoked?

And that works out well... Maybe we won't need a preference at all?

lddubeau:
Oops... forgot to close this when #303 was merged.

PR #303, opened 2019-02-11
allow not starting tsserver immediately when starting tide

lddubeau:
This is the companion WIP to issue #294 

I've been using this for a few weeks now without running into actual issues (commit dates are deceptive: I used it before the code was committed) so I figure I should put it up for public view.

Going into a buffer that does not have `tsserver` started causes some things to not work but *that's expected*. When I finally start a `tsserver` manually it works just as well as if the server had been started automatically.

### DIFF
--- a/tide-tests.el
+++ b/tide-tests.el
@@ -147,8 +147,7 @@
        ;; We need this so that tests that simulate user actions operate on the right buffer.
        (switch-to-buffer (current-buffer))
        (save-current-buffer ,@body)
-       (-when-let (,server (tide-current-server))
-         (delete-process ,server)))))
+       (tide-kill-server))))
 
 (defmacro wait-for (&rest body)
   "Wait until BODY executes without error.  There's an arbitrary 5 second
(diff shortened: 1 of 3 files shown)


tide.el
```diff
@@ -1869,28 +1898,15 @@ code-analysis."
...
 
-  (tide-start-server-if-required)
-  (tide-mode 1)
   (set (make-local-variable 'eldoc-documentation-function)
        'tide-eldoc-function)
   (set (make-local-variable 'imenu-auto-rescan) t)
   (set (make-local-variable 'imenu-create-index-function)
        'tide-imenu-index)
 
-  (tide-configure-buffer)
```
ananthakumaran: any reason for removing the configure-buffer call for immediate mode? Note that we can't depend on `hack-local-variables-hook` see https://github.com/ananthakumaran/tide/commit/fbdef5d7af25786d2eddd691a8d6a00225c4202d

ananthakumaran:
rest of the changes looks straight forward to me.

tide.el
lddubeau: Ah, yes. When someone creates a new TS buffer, then `tide-configure-buffer` has never been called to configure this buffer. If `hack-local-variables-hook` fails to run, then the buffer won't be configured at all. I did not run into problems because presumably, with my configuration, `hack-local-variables-hook` always runs and took care of configuring the buffer.  I'll fix that.

lddubeau:
I've added a [third commit](https://github.com/ananthakumaran/tide/pull/303/commits/26c208a24cbfcd7516c4fade1aec5c7578090080) that takes care of the issue pointed out in the earlier [code review](https://github.com/ananthakumaran/tide/pull/303#pullrequestreview-202084433).

And I've rebased onto the current `master`.

tide.el
```diff
@@ -1871,28 +1895,27 @@ code-analysis."
...
+      ;; cannot rely on hack-local-variable-hook in tide-mode because the hook
+      ;; may not run at all, or run too late.
+      ;;
+      ;; It may happen for some use-case scenarios that tide-configure-buffer
+      ;; runs more than once with the same data for the same buffer, but that's
+      ;; not a big deal.
+      ;;
+      (tide-configure-buffer)
+    (when (eq tide-tsserver-start-method 'immediate)
+      (tide-start-server)))
```
ananthakumaran: when we start the server for the first time, do we expect the tide-start-server to call configure-buffer? I am assuming that would miss the current buffer because tide-mode is not yet activated?

lddubeau: Yeah, there's a problem there. I think at this point since we're discussing details rather than general objections to the approach, I should just get the tests together (which would most likely have revealed the problem here).

lddubeau:
I've added formal tests and fixed the execution path that resulted in a buffer not getting configured.

tide-tests.el
```diff
@@ -219,6 +266,64 @@ a test failure."
...
+  "Test that when a server is started after an initial buffer was
+opened in manual mode, a 2nd buffer opened in the same project is
+automatically configured."
+  (let ((tide-tsserver-start-method 'manual))
+    (with-spy configure-spy tide-command:configure
+      (with-open-file "test/trivial.ts"
+        (should (eq (tide-current-server) nil))
+        (should (not (spy-called configure-spy)))
+        (tide-restart-server)
+        (should (spy-called configure-spy))
```
josteink: Should we have a `(should (not (eq (tide-current-server) nil)))` here for consistency?

josteink:
I've looked over the overall changes and I see nothing *obviously* wrong, nor any changes which looks particularly dangerous.

If you say this has been working for you "for months", I see no reason for not merging this.

lddubeau:
@josteink Thanks. I'm going to merge this now.

PR #304, opened 2019-02-15
Fix tide-list-servers when cpu is nil

aaronjensen:
Either on Emacs 27 or macOS, cpu is nil and `isnan` throws when you give it nil. Either way, this fixes it for me.

### DIFF
--- a/tide.el
+++ b/tide.el
@@ -2441,7 +2441,8 @@ timeout."
                      ;; Sometimes the CPU usage value is NaN (which Emacs represents
                      ;; as 0.0e+NaN), for whatever reason. We cannot pass this value
                      ;; to round so we put "--" for the column value.
-                     (if (isnan cpu)
+                     ;; Other times, it is nil, so do the same then.
+                     (if (or (not cpu) (isnan cpu))
                          "--"
                        (number-to-string (round cpu)))
                      (case tide--server-list-mode-last-column


josteink:
LGTM. 

While we’re at it... Are there other places where we are using “isnan” which could face similar issues?

aaronjensen:
Let me take a gander.

aaronjensen:
Nope, that's the only one.

josteink:
Awesome. Thanks for the patch!

aaronjensen:
no problem, thank you for the quick merge.

lddubeau:
Thanks for the fix. The `pcpu` field of `process-attributes` apparently can contain unexpected values. [In this comment](https://github.com/ananthakumaran/tide/pull/279#issuecomment-450639700) pointed out the change to handle NaN. It was pure luck that I even caught this problem. I happened to change laptops and got NaN values. Before that I had run the code for a long time without ever seeing NaN. Back then, I searched the documentation to try to figure what the deal was with NaN but I did not find anything concrete.

I just now did a search in the `master` branch of Emacs and found some notes about how on MacOS, non-root processes are prevented from getting information about other processes. I suspect this explains the `nil` you get on MacOS. I also suspect the other fields from `process-attributes` to be unusable on MacOS.

Issue #312, opened 2019-05-04
Tide ignores tsconfig.json settings


jclangst:
When submitting issues to this project, please include the following information.

## Checklist

- [x] I have searched both open and closed issues and cannot find a duplicate.
- [X] I can [reproduce the problem](https://github.com/ananthakumaran/tide/wiki/Explanations-about-the-Issue-Reporting-Checklist#use-the-latest-version-of-the-relevant-packages) with the latest version of the relevant packages.
- [x] The problem still occurs after I issued [`M-x tide-restart-server`](https://github.com/ananthakumaran/tide/wiki/Explanations-about-the-Issue-Reporting-Checklist#try-m-x-tide-restart-server) in the buffer where I had the problem.
- [X] I verified that the version and the configuration file path reported by [`M-x tide-verify-setup`](https://github.com/ananthakumaran/tide/wiki/Explanations-about-the-Issue-Reporting-Checklist#check-the-information-reported-by-m-x-tide-verify-setup) are correct.
- [X] If [tide *is reporting*](https://github.com/ananthakumaran/tide/wiki/Explanations-about-the-Issue-Reporting-Checklist#the-behavior-of-cli-tools-agrees-with-your-expectations) an error or warning I think *should not be reported*, I can run `tsc` (and `tslint`, if applicable) without the error or warning I'm seeing in tide.
- [x] If [tide is *not reporting*](https://github.com/ananthakumaran/tide/wiki/Explanations-about-the-Issue-Reporting-Checklist#the-behavior-of-cli-tools-agrees-with-your-expectations) an error or warning I think *should be reported*, `tsc` (or `tslint`, if applicable) reports the error or warning I was expecting to see.
- [ ] I am [positive](https://github.com/ananthakumaran/tide/wiki/Explanations-about-the-Issue-Reporting-Checklist#how-to-determine-which-project-the-problem-belongs-to) the problem does not belong to `typescript-mode` or `tsserver`.

## Relevant Version Numbers

- Tide: 20190321.1611
- TypeScript: 3.4.5 (also tried on 3.3)
- Emacs: 26.2

(The list above is minimal. Make sure to include any other version numbers relevant to your report.)

## Steps to Reproduce the Bug

(1) Set up typescript project
(2) Define function with implicit any behavior like such:

```typescript
function (id) {
  return id;
}
```
[...]

PR #315, opened 2019-05-05
Disabled checking for all node modules and fixes suggestion bug for t…

### LINKED ISSUE #314: Tide should ignore errors/suggestions in files in "node_modules/@types"

## Checklist

- [x] If [tide *is reporting*](https://github.com/ananthakumaran/tide/wiki/Explanations-about-the-Issue-Reporting-Checklist#the-behavior-of-cli-tools-agrees-with-your-expectations) an error or warning I think *should not be reported*, I can run `tsc` (and `tslint`, if applicable) without the error or warning I'm seeing in tide.
- [x] If [tide is *not reporting*](https://github.com/ananthakumaran/tide/wiki/Explanations-about-the-Issue-Reporting-Checklist#the-behavior-of-cli-tools-agrees-with-your-expectations) an error or warning I think *should be reported*, `tsc` (or `tslint`, if applicable) reports the error or warning I was expecting to see.
- [x] I am [positive](https://github.com/ananthakumaran/tide/wiki/Explanations-about-the-Issue-Reporting-Checklist#how-to-determine-which-project-the-problem-belongs-to) the problem does not belong to `typescript-mode` or `tsserver`.

## Relevant Version Numbers

- Tide:  20190321.1611
- TypeScript: 3.4.5
- Emacs: 26.2

## Summary

When importing definition files from node_modules, these definition files are being treated as files of the project by `tsserver`. As a result, the `tide-project-errors` buffer shows suggestions and errors in these third-party files. You can see an example of this inclusion in the left hand side of this image: https://user-images.githubusercontent.com/13301098/57183295-44224580-6e70-11e9-8ec6-c5d34749f203.png

## Suggested Fix

In `tide-display-erros` [sic], we are excluding some files like those that have "node_modules/typescript" in them. I think this should be shortened to just "node_modules" because we don't want to show errors or suggestions for third-party projects.

jclangst:
This would close #314 and fix the issues with the original PR's fix for #312.

The relevant changes are:

(1) Changes to ignore all errors/suggestions from files in "node_modules"
(2) Properly does not do any processing at all for suggestion errors IFF `tide-disable-suggestions` is set to anything but nil.

The PR diff is a bit blocky due to auto-indentation.

### DIFF
--- a/tide.el
+++ b/tide.el
@@ -2122,7 +2122,7 @@ code-analysis."
       (erase-buffer))
     (display-buffer (current-buffer) t)
     (let* ((project-files (-reject (lambda (file-name)
-                                     (or (string-match-p "node_modules/typescript/" file-name)
+                                     (or (string-match-p "node_modules" file-name)
                                          (string-match-p "tsconfig.json$" file-name)))
                                    file-names))
            (syntax-remaining-files (cl-copy-list project-files))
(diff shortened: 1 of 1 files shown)


tide.el
```diff
@@ -2138,41 +2138,43 @@ code-analysis."
...
-                (progn
-                  (setq semantic-remaining-files (remove file-name semantic-remaining-files))
-                  (cl-incf semantic-errors (length diagnostics))))
-               ("suggestionDiag"
-                (unless tide-disable-suggestions
+           (let* ((inhibit-read-only t)
+		  (file-name (tide-plist-get response :body :file))
+		  (diagnostics (tide-plist-get response :body :diagnostics))
+		  (eventType (plist-get response :event))
+		  (shouldIgnore (and (string-equal eventType "suggestionDiag") tide-disable-suggestions)))
```
ananthakumaran: The logic looks correct, could you just inline these as the vars are used in only one place. Also, camelCase is not the standard in emacs lisp.

jclangst: Duh! Fixed the camelCase. `event-type` is used multiple times so I kept that in there.

PR #316, opened 2019-05-11
use other backend when tide can't completion.

zedhugh:
sometimes, I'd like use "company-file" to complete file name,
but I find that "company-tide" never passed the control to
make company use other backend. so I fix this problem with
this commit.

### DIFF
--- a/tide.el
+++ b/tide.el
@@ -4,7 +4,7 @@
 
 ;; Author: Anantha kumaran <ananthakumaran@gmail.com>
 ;; URL: http://github.com/ananthakumaran/tide
-;; Version: 3.3.4
+;; Version: 3.5.1
 ;; Keywords: typescript
 ;; Package-Requires: ((dash "2.10.0") (s "1.11.0") (flycheck "27") (typescript-mode "0.1") (cl-lib "0.5"))
 
--- a/tide.el
+++ b/tide.el
@@ -1361,7 +1361,7 @@ This function is used for the basic completions sorting."
       ;; Rank declarations lower than variables
       (string-equal modifier-b "declare"))))
 
-(defcustom tide-allow-other-backend-for-string t
+(defcustom tide-allow-other-backend-for-string nil
   "Whether allow other backends provide completion for string."
   :type 'boolean
   :group 'tide)
--- a/tide.el
+++ b/tide.el
@@ -1369,7 +1369,12 @@ This function is used for the basic completions sorting."
                     (and "import(" (or ?\" ?') (0+ (not (any ?\" ?'))))
                     (and "require(" (or ?\" ?') (0+ (not (any ?\" ?'))))))))
       (cons (company-grab (rx (or ?/ ?\" ?') (group (0+ (not (any ?\" ?'))))) 1) t)
-    (company-grab-symbol-cons "\\." 1)))
+    (let ((comp (company-grab-symbol-cons "\\." 1)))
+      (if (and (tide-in-string-p)
+               (stringp comp)
+               (string-empty-p comp))
+          nil
+        comp))))
 
 (defun tide-member-completion-p (prefix)
   (save-excursion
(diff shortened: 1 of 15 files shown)


josteink:
Hey there and thanks for the PR. 

Just skimming the diff it looks good to me, but i haven’t been able to test it yet. 

Also it might be good if whoever wrote this code in the first place approved these changes, since they might have had a reason for doing things as they did. 

Might be best if @anantakumaran approved this one.

zedhugh:
OK. I'm just confused why not passed to use other backend when tide can not provide any completion.

ananthakumaran:
If I remember correctly tide used to pass the control to next backend if the cursor is inside a string. But tsserver later added support for completion of string literals, file path, etc https://github.com/ananthakumaran/tide/pull/132/files

zedhugh:
Thanks for reply, but I found tsserver just work well for module name, cannot complete other file name like a image file name, and I need complete common file name.

ananthakumaran:
but the current changes would disable the string completion by tsserver? Do you know any other way? I am also ok with adding custom config like `tide-completion-enable-inside-string`

zedhugh:
add a custom config maybe a better way than my patch. My patch make `tide-completion-prefix` return nil when the old version return an empty string. I don't know detail of tsserver, but I think it can't privode any completion when `tide-completion-prefix` return empty string. At this time, use other backend for completion seems right.

ananthakumaran:
Tsserver itself doesn't require the prefix (it's optional)
```
Info 477  [22:30:37.778] request:
    {"command":"completions","seq":"49","arguments":{"file":"../tide/example/Another.ts","line":3,"offset":21,"includeExternalModuleExports":true,"includeInsertTextCompletions":true}}
Info 479  [22:30:37.778] response:
    {"seq":0,"type":"response","command":"completions","request_seq":"49","success":true,"body":[{"name":"hello","kind":"string","kindModifiers":"","sortText":"0"}]}
```

zedhugh:
Got it now. so this is a feature miss for tsserver?
for example, files like this:
|--- module.ts
|--- test.ts
|--- a.png
in test.ts, it works well when I import anything from module.ts,
but when I typed this code:
```
const imageFileName = "./a.|"
```
I want it complete "a.png" as a file name.
But, I tried call `company-tide` and `company-complete` by `M-x`, both of them can not work as I wish.

ananthakumaran:
>  Got it now. so this is a feature miss for tsserver?
I won't say it's a missing feature. I don't think tsserver is going to support the completion of non ts/js files well and it's not even in the context of import/require.

This looks too specific in my view. I am ok with adding a custom config to disable completion within string so other backends can kick in and do file completion.

zedhugh:
Glad to here this. or just need I change my PR as an custom config?

tide.el
```diff
@@ -1469,7 +1480,7 @@ This function is used for the basic completions sorting."
              (-any-p #'derived-mode-p tide-supported-modes)
              (tide-current-server)
              (not (nth 4 (syntax-ppss)))
```
ananthakumaran: could we just add `(not (and (not tide-completion-enable-inside-string) (tide-in-string-p)))`

zedhugh: It's not what I wanted. With this change, I don't need change any code of tide, make `company-files` in front of `company-tide` can get the same effect.

I want other backends provide completion only in regular string when `tide` cannot.

with you code, tide cannot provide completion in template string and import statement. And this is not what I want.

ananthakumaran:
@josteink feel free to share your opinions. I am kind of not sure what's the right approach here.

josteink:
I’ve pulled in both images and CSS via bundler/loader magic before. 

Having local file-names being offered as completion options would be nice when you need it, but I’m not sure if I’d like it to always be enabled. 

Having it as a config (with an easy toggle-command, and keybinding?) might be ok. 

Another approach might be to only enable it if completion is triggered with prefix-arguments. Would that be possible at all? And if so,  are company prefix-arguments already used for other options?

zedhugh:
If there is no good approach for this.
maybe call `company-files` manually or binding this function with a keybinding is the best way.
I want find a way which can implement this without involve anything in.

PR #324, opened 2019-06-24
Highlight fewer references in large buffers

sandersn:
tide-location-to-point is slow (linear?) in large buffers. When you highlight something with lots of references like `boolean` or some other common data type, tide will freeze while highlighting references.

This change highlights only 20 nearby references in large buffers, which are defined as >10,000 lines.

The right fix is probably to make `tide-location-to-point` fast, but I don't know enough about emacs to do that.


sandersn:
We could also not highlight `boolean`, `true`, `undefined`, etc, but in Typescript's source code, trying to highlight `Node` in the 30,000-line checker.ts is also very slow.

lddubeau:
Thank you. As discussed on #325, closing in favor of #325

PR #330, opened 2019-08-10
Make all tsserver file line endings consistent

schellj:
Previously, the `tsserver` files had a mix of unix and dos line endings.  Most lines were dos style (\r\n) with the primary exception being the Microsoft license header, which was unix style (\n).  This PR makes all line endings dos style.


josteink:
This is the cheapest, biggest ever commit in this repo, but I’m ok with it :smiley:

You got anything against it, @anathakumaran?

schellj:
Heh, I didn't realize the diff count was over 250k lines.

All I did was run `unix2dos` on all of the relevant files.

ananthakumaran:
I prefer not to make any changes to these files. These files are the result of the compilation process and directly copied over without any changes. Even if we commit this change, it's going to get changed next time when we upgrade the tsserver version. I guess the mixed line endings are probably due to how things are done in Typescript repo.

schellj:
Understood.

The files in https://github.com/microsoft/TypeScript/blob/master/lib/ don't have this issue.  I just tried `make update-tsserver` and the gulp-built files do have the mixed line endings issue.  Hmm.

schellj:
Looks like the issue stems from the fact that https://github.com/microsoft/TypeScript/blob/master/CopyrightNotice.txt uses LF line endings, while some or all of the files in https://github.com/microsoft/TypeScript/tree/master/src/lib use CRLF line endings.  Thus, when `gulp local` concatenates the copyright notice and lib sources, the line endings per built lib file are inconsistent.

I looked through https://github.com/microsoft/TypeScript/blob/master/Gulpfile.js and the gulp cli options and didn't see a way to force line endings to be one way or the other.  I'm not familiar with gulp, so there might be a way to do that that I missed.

In any case, this PR is not the correct way to handle this issue, so I'll close it.

josteink:
> Looks like the issue stems from the fact that https://github.com/microsoft/TypeScript/blob/master/CopyrightNotice.txt uses LF line endings, while some or all of the files in https://github.com/microsoft/TypeScript/tree/master/src/lib use CRLF line endings. 

So... Maybe you could make a PR in the main TypeScript repo and fix the inconsistency there instead?

schellj:
@josteink yeah, I was planning to open an issue or PR if I could figure out a proper fix (which probably isn't just to change the line endings on the copyright notice).

PR #332, opened 2019-08-17
allow fuzzy completion

ananthakumaran:


kijowski:
Hi @ananthakumaran. I've taken this feature for a test ride and it works fine with one small caveat - setting `(setq tide-completion-fuzzy t)` ignores `tide-completion-ignore-case` value and makes completion case-insensitive by default (at least in my setup). This is because `string-match-p` case sensitivity is controlled by `case-fold-search`. It can be easily fixed by following patch:
```
 @@ -1669,7 +1669,8 @@ This function is used for the basic completions sorting."
   (-filter (lambda (completion)
              (and
               (if tide-completion-fuzzy
-                  (string-match-p (regexp-quote prefix) (plist-get completion :name))
+                  (let ((case-fold-search tide-completion-ignore-case)) 
+                    (string-match-p (regexp-quote prefix) (plist-get completion :name)))
                 (string-prefix-p prefix (plist-get completion :name) tide-completion-ignore-case))
               (or (not tide-filter-out-warning-completions)
                   (not (equal (plist-get completion :kind) "warning")))))
@@ -1818,7 +1819,8 @@ This function is used for the basic completions sorting."
      (let* ((completion (get-text-property 0 'completion arg))
             (prefix (get-text-property 0 'prefix arg))
             (start (if tide-completion-fuzzy
-                       (string-match-p (regexp-quote prefix) (plist-get completion :name))
+                       (let ((case-fold-search tide-completion-ignore-case)) 
+                         (string-match-p (regexp-quote prefix) (plist-get completion :name)))
                      0)))
        `((,start . ,(+ start (length prefix))))))
     ((annotation) (tide-completion-annotation arg))
```
Other than that I think that fuzzy matching is working as intended. Please let me know if you want me to check some specific use case.

ananthakumaran:
Thanks for testing the PR and your patch. It looks like I missed the case flag. I will merge soon

PR #340, opened 2019-10-18
Minor improvements

elibarzilay:
Some random minor improvements, split into different commits in case some are not wanted.

### DIFF
--- a/tide.el
+++ b/tide.el
@@ -1571,7 +1571,7 @@ This function is used for the basic completions sorting."
   (let ((map (make-sparse-keymap)))
     (define-key map (kbd "n") #'tide-find-next-reference)
     (define-key map (kbd "p") #'tide-find-previous-reference)
-    (define-key map (kbd "C-m") #'tide-goto-reference)
+    (define-key map (kbd "RET") #'tide-goto-reference)
     (define-key map [mouse-1] #'tide-goto-reference)
     map))
 
@@ -2059,7 +2059,7 @@ code-analysis."
    'keymap (let ((map (make-sparse-keymap)))
              (define-key map [mouse-2] 'tide-goto-error)
              (define-key map [mouse-1] 'tide-goto-error)
-             (define-key map (kbd "C-m") 'tide-goto-error)
+             (define-key map (kbd "RET") 'tide-goto-error)
              (define-key map [follow-link] 'mouse-face)
              map)
    'tide-error filespan))
@@ -2265,7 +2265,7 @@ code-analysis."
   (let ((map (make-sparse-keymap)))
     (define-key map (kbd "n") #'tide-find-next-error)
     (define-key map (kbd "p") #'tide-find-previous-error)
-    (define-key map (kbd "C-m") #'tide-goto-error)
+    (define-key map (kbd "RET") #'tide-goto-error)
     map))
 
 (define-derived-mode tide-project-errors-mode special-mode "tide-project-errors"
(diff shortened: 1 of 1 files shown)


josteink:
Hey there and thanks for the PR!

This looks like good, simple refactorings and I appreciate that!

Some of them are straight forward and no discussion needed. Others I’ll have to take a deeper look into, but I won’t have time for that until after the weekend. 

In the meantime, if you have more improvements you’d like to add, feel free!

tide.el
```diff
@@ -1755,21 +1755,22 @@ number."
...
+          ;; check. So make sure all the other files are saved
+          ;; before saving current file.
 
-            (-each (nconc (-reject #'current-file-p locs)
-                          (-select #'current-file-p locs))
-              (lambda (loc)
-                (cl-incf count (tide-rename-symbol-at-location loc new-symbol))))
+          (-each (nconc (-reject current-file-p locs)
+                        (-select current-file-p locs))
+            (lambda (loc)
```
josteink: You've removed the `#'` function-specific quoting which while not required, provides the byte-compiler with a hint that the quoted symbol should be a function.

So as far as I know, this change may provide use with less sturdy code than we used to have. Is there any particular reason for this change?

tide.el
```diff
@@ -1975,16 +1976,16 @@ code-analysis."
...
+             (if err
+                 (progn
+                   (setq resolved t)
+                   (funcall cb err))
+               (when (and (plist-member result :syntaxDiag)
+                          (plist-member result :semanticDiag)
+                          (plist-member result :suggestionDiag))
+                 (setq resolved t)
+                 (funcall cb `(:body (,result) :success t)))))))
       (tide-send-command
```
josteink: From what I can tell, this is just change in indentation, right?

josteink:
Hey there, and sorry about the delay!

I've looked over the code, and tested the changes. Apart from the stripping of `#'`, I think the rest of these changes look good and on overall helps make the code simpler to read. I'm all for them. 👍 

Could you re-add the `#'` to the function-references, or alternatively elaborate on why you think it makes sense to remove them?

tide.el
elibarzilay: No, I made the function be bound with the simpler `let` instead of `cl-flet`, which means that the plain value is used rather than the function value, so `#'` is not needed (and is wrong).  Simplified demonstration: I changed

    (cl-flet ((a1 (x) (1+ x))) (mapcar #'a1 '(1 2 3)))

to

    (let ((a1 (lambda (x) (1+ x)))) (mapcar a1 '(1 2 3)))

Add a `#'` to the latter `a1`, and it'll break.

(And it's interesting to use `macroexpand` on the former to see why it's more complex, and as a sidenote, see also that the `#'` doesn't even make its way to the compiler...)

tide.el
elibarzilay: Yes; the default Emacs indentation for `cl-flet` etc is overly naive which makes this a common indentation mistake...  (See [this bug](https://debbugs.gnu.org/cgi/bugreport.cgi?bug=9622) too.)

elibarzilay:
Replied to both comments.  (I originally marked them resolved, but it looks like that hides the comments, so I unresolved them and will leave it to you to do so.)

josteink:
Thanks for the feedback. That absolutely makes sense.

Since I haven't seen any other objections in here the last ... 12 days, I guess nobody else have any objections.

Thanks for the patch. Consider it merged!

elibarzilay:
Thanks!

PR #342, opened 2019-10-30
Improve resource leftovers

elibarzilay:
There are two things here that are both aimed at reducing the amount of
resource clutter:

1. I've seen many tmp files left.  Looking at the code, it seems to me
   like a reasonable thing to get rid of them when a file is saved.  I
   did this via adding `tide-remove-tmp-file` on `after-save-hook`.

2. I also saw that there is a general tendency to accumulate many
   servers since they're never removed (actually, more than just the
   servers -- the whole project resources are kept, but the server is
   the main problem wrt resources).  This is also mentioned in #256.

   So I implemented a function that scans all buffers and cleanup all
   projects that have no live buffers.  It looks to me like a good idea
   to do this, since you can just kill old buffers to reduce resource
   usage.  (And killing old buffers is more obvious than explicitly
   openning the server list to kill old ones, especially since there's
   no way to tell if a server is used by some buffer or not.)

   (I added this function onto `kill-buffer-hook`, but if that's too
   extreme, then a more mild option is to not do that and just let
   people add it themselves.)

### DIFF
--- a/tide.el
+++ b/tide.el
@@ -382,12 +382,12 @@ ones and overrule settings in the other lists."
 (defmacro tide-on-response-success (response &optional options &rest body)
   "BODY must be a single expression if OPTIONS is not passed."
   (declare (indent 2))
-  (when (not body)
+  (unless body
     (setq body `(,options))
     (setq options '()))
   (tide-plist-map
    (lambda (key _value)
-     (when (not (member key '(:ignore-empty :min-version)))
+     (unless (member key '(:ignore-empty :min-version))
        (error "Invalid options %S" options)))
    options)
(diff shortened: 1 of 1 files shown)


lddubeau:
Thanks for the PR. Two changes are being proposed here: one for improving the cleanup of tmp files, and one for killing superfluous `tsserver` instances. Both are improvements and both aim at cleaning up left-overs, but cleaning tmp files and managing `tsserver` instances are two very different aspects of tide. IMO, this calls for two separate PRs.

lddubeau:
Thank you for splitting the PR.

PR #344, opened 2019-10-30
More temp file cleanups

elibarzilay:
I've seen many tmp files left.  Looking at the code, it seems to me like
a reasonable thing to get rid of them when a file is saved.  I did this
via adding `tide-remove-tmp-file` on `after-save-hook`.

### DIFF
--- a/tide.el
+++ b/tide.el
@@ -202,7 +202,7 @@ errors and tide-project-errors buffer."
 
 (defcustom tide-project-cleanup-delay 60
   "The number of idle seconds to wait before cleaning up unused tsservers.
-Use `nil` to disable automatic cleanups.  See also `tide-cleanup-dead-projects`."
+Use `nil` to disable automatic cleanups.  See also `tide-do-cleanups`."
   :type '(choice (const nil) integer)
   :group 'tide)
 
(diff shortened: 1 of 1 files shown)


josteink:
LGTM.

ananthakumaran:
> I've seen many tmp files left

is tide leaving garbage files around? Each buffer might have a tmp file associated with it, but it should be cleared once the buffer is killed.

As for clearing file on save-hook, I am afraid this might introduce race conditions. We store the buffer contents to tmp file and send the file name to tsserver. This could happen in an async manner during auto-complete etc. Clearing file on save could introduce the possibility of delete operation interfering with tsserver read file operation.

elibarzilay:
Yes.  Most of them are likely left when you exit emacs (which doesn't invoke the kill buffer hooks).

I haven't seen any such bad side effect so far, and I'd be very surprised if something like autocompletion is really async wrt emacs operations: can it perform an autocomplete after I move the cursor to a different place?  (I hope that the answer is not because the interaction would cancel any pending async operations -- and on the same token saving the buffer would cancel them too.)

ananthakumaran:
Just to explain the current mechanism, we don't send each and every file edits to tsserver. What we do is, just set a dirty buffer flag and when some other command is triggered (like autocomplete), we sync the buffer via `reload` command. Tide has two types of way to send command `tide-send-command` and `tide-send-command-sync`. The former returns right away without waiting for ack from tsserver. A callback could be supplied to get the ack & results.

As for autocomplete, company api itself is async, we get a callback function that should be called when we have candidates list.

```elisp
(candidates (cons :async
                      (lambda (cb)
                        (tide-command:completions arg cb))))
```

The completions command would trigger `reload` command (which is async) and also issue `completions` command (async as well). The candidates would be sent to company when the second callback is fired.

> can it perform an autocomplete after I move the cursor to a different place?

What happens once the call is triggered is not controlled by company. What it can do is throw away the candidates result returned by old callbacks. The is no cancel api to cancel existing callback. I mentioned completion as an example. There are other async commands as well.

> on the same token saving the buffer would cancel them too
saving the file typically triggers flycheck which in turn will trigger syntax check and so on.

I would be very hesitant to do the cleanup on save-hook unless somebody convinces me.

elibarzilay:
Thanks for the explanation.  IIUC, the concern is not about editing problems, but tsserver problems when the file disappears.  Wouldn't this concern apply just the same for the temporary file disappearing when a buffer is killed?

(Or in fact, given that it's common to do async operations, then there could be problems when files disappear in general not just with Emacs, and therefore there shouldn't be any bad results from such things happening--?)

ananthakumaran:
> Wouldn't this concern apply just the same for the temporary file disappearing when a buffer is killed?

Strictly speaking yes. We issue a `close` operation right before cleaning the file on buffer kill. It signals that tide won't sync the file manually anymore and tsserver should just watch the real file. There might be a small window for race conditions, but the close operation should eventually bring the tsserver to a correct state.

elibarzilay:
I thought about this again, and I still think that it's fine -- after all, if I didn't have this extension in my copy, I would probably start a habit of killing buffers and re-opening them from time to time just to avoid the `/tmp` junk.

Also, FWIW, I've been using this and didn't see any problems.

josteink:
I'm OK with this as it is. We're no longer going over *all* open buffers.

Any objections? If not, I'll merge this after the weekend.

@elibarzilay: For future PRs, it would be nice if you didn't pre-squash your commits and force push after we've had comments or done code reviews, as it makes it harder to track what changes are new since last time you looked into the PR.

So while the commits are cleaner, it actually makes the review harder, and probably slower too.

If we feel the need to squash before merging, we can do that on our own :smiley:

elibarzilay:
@josteink, to be clear, I think that you might be confusing this with the dead-projects PR...  This one has never went over all open buffers, and also it always had a single commit (I didn't squash it, just updated it as other PRs got in).

(And yes, re the concerns raised above -- I can say that I continue using this code and nothing broke...)

ananthakumaran:
The concern I have raised earlier is still not cleared. "Works for me" is not a good argument when we discuss the possibility of race conditions. I might have different flow compared to others (like autosave) and the race condition might only manifest there.

elibarzilay:
My comment should have been "*all* I can say": not an argument.  To clarify it further:

1. I still don't see *how* it could go wrong, vs killing the buffer and re-opening the file.

2. And I don't see a *good* alternative way to avoid the accumulation of temporary files.

The *least bad* way that I could think of is an extension to the dead project cleanup thing: make it also crawl over all buffers with a temp file and call `tide-remove-tmp-file` on it.  I think that it's an overkill, but at least it will get something that works.  (I plan to change to that after the other one is merged.)

elibarzilay:
@ananthakumaran: I re-implemented this using (a generalized version of) the cleanup thing, so it should happen only in idle time.

elibarzilay:
ping @ananthakumaran

elibarzilay:
Thanks!

PR #345, opened 2019-10-30
Project cleanup when buffers are killed

elibarzilay:
There is a general tendency to accumulate many servers since they're
never removed (actually, more than just the servers -- the whole project
resources are kept, but the server is the main problem wrt resources).
This is also mentioned in #256.

So I implemented a function that scans all buffers and cleanup all
projects that have no live buffers.  It looks to me like a good idea to
do this, since you can just kill old buffers to reduce resource
usage.  (And killing old buffers is more obvious than explicitly
openning the server list to kill old ones, especially since there's no
way to tell if a server is used by some buffer or not.)

(I added this function onto `kill-buffer-hook`, but if that's too
extreme, then a more mild option is to not do that and just let people
add it themselves.)

### DIFF
--- a/tide.el
+++ b/tide.el
@@ -739,6 +739,9 @@ in the npm global installation.  If nothing is found use the bundled version."
   (tide-each-buffer project-name
                     (lambda ()
                       (tide-cleanup-buffer-callbacks)))
+  (tide-cleanup-project-data project-name))
+
+(defun tide-cleanup-project-data (project-name)
   (remhash project-name tide-servers)
   (remhash project-name tide-tsserver-unsupported-commands)
   (remhash project-name tide-project-configs))
@@ -764,7 +767,7 @@ In particular, kill their tsserver processes."
       (message "Cleaning up %d projects..." (length to-kill))
       (dolist (proj to-kill)
         (delete-process (process-buffer (gethash proj tide-servers)))
-        (tide-cleanup-project proj))
+        (tide-cleanup-project-data proj))
       (sit-for 5)
       (message nil))))
 
--- a/tide.el
+++ b/tide.el
@@ -762,7 +762,9 @@ In particular, kill their tsserver processes."
         (cl-pushnew proj live-projects)))
     (-when-let (to-kill (-difference (hash-table-keys tide-servers) live-projects))
       (message "Cleaning up %d projects..." (length to-kill))
-      (dolist (proj to-kill) (tide-cleanup-project proj))
+      (dolist (proj to-kill)
+        (delete-process (process-buffer (gethash proj tide-servers)))
+        (tide-cleanup-project proj))
       (sit-for 5)
       (message nil))))
 
--- a/tide.el
+++ b/tide.el
@@ -200,6 +200,12 @@ errors and tide-project-errors buffer."
   :type 'function
   :group 'tide)
 
+(defcustom tide-project-cleanup-delay 60
+  "The number of idle seconds to wait before cleaning up unused tsservers.
+Use `nil` to disable automatic cleanups.  See also `tide-cleanup-dead-projects`."
+  :type '(choice (const nil) integer)
+  :group 'tide)
+
 (defcustom tide-default-mode "TS"
   "The default mode to open buffers not backed by files (e.g. Org
 source blocks) in."
@@ -737,17 +743,28 @@ in the npm global installation.  If nothing is found use the bundled version."
   (remhash project-name tide-tsserver-unsupported-commands)
   (remhash project-name tide-project-configs))
 
-(defun tide-cleanup-dead-projects* ()
-  (tide-cleanup-dead-projects (current-buffer)))
-(defun tide-cleanup-dead-projects (ignore)
+(defvar tide--project-cleanup-timer nil)
+(defun tide-schedule-dead-projects-cleanup ()
+  (when (and tide-project-cleanup-delay (not tide--project-cleanup-timer))
+    (setq tide--project-cleanup-timer
+          (run-with-idle-timer tide-project-cleanup-delay nil
+                               #'tide-cleanup-dead-projects))))
+(defun tide-cleanup-dead-projects ()
+  "Cleanup projects that don't have any associated live buffers.
+In particular, kill their tsserver processes."
+  (interactive)
+  (setq tide--project-cleanup-timer nil)
   (let ((live-projects '()))
-    (dolist (b (-remove-item ignore (buffer-list)))
+    (dolist (b (buffer-list))
       (-when-let (proj (with-current-buffer b
                          (and (bound-and-true-p tide-mode)
                               (tide-project-name))))
         (cl-pushnew proj live-projects)))
-    (dolist (proj (-difference (hash-table-keys tide-servers) live-projects))
-      (tide-cleanup-project proj))))
+    (-when-let (to-kill (-difference (hash-table-keys tide-servers) live-projects))
+      (message "Cleaning up %d projects..." (length to-kill))
+      (dolist (proj to-kill) (tide-cleanup-project proj))
+      (sit-for 5)
+      (message nil))))
 
 (defun tide-start-server-if-required ()
   (unless (tide-current-server)
@@ -2031,7 +2048,7 @@ code-analysis."
         (add-hook 'after-save-hook 'tide-auto-compile-file nil t)
         (add-hook 'after-change-functions 'tide-handle-change nil t)
         (add-hook 'kill-buffer-hook 'tide-cleanup-buffer nil t)
-        (add-hook 'kill-buffer-hook 'tide-cleanup-dead-projects* nil t)
+        (add-hook 'kill-buffer-hook 'tide-schedule-dead-projects-cleanup nil t)
         (add-hook 'hack-local-variables-hook 'tide-configure-buffer nil t)
         (when (commandp 'typescript-insert-and-indent)
           (eldoc-add-command 'typescript-insert-and-indent)))
@@ -2039,7 +2056,7 @@ code-analysis."
     (remove-hook 'after-save-hook 'tide-auto-compile-file t)
     (remove-hook 'after-change-functions 'tide-handle-change t)
     (remove-hook 'kill-buffer-hook 'tide-cleanup-buffer t)
-    (remove-hook 'kill-buffer-hook 'tide-cleanup-dead-projects* t)
+    (remove-hook 'kill-buffer-hook 'tide-schedule-dead-projects-cleanup t)
     (remove-hook 'hack-local-variables-hook 'tide-configure-buffer t)
     (tide-cleanup-buffer)))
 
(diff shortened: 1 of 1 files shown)


elibarzilay:
Well, the main question re the "too extreme" thing is whether this kind of functionality is something that you *do* want for some reason, regardless of performance.

The inefficiency was mainly because I was trying to get something that doesn't require any other changes, and also could be used interactively when needed.  If this kind of cleanup is wanted, then it's easy to make it efficient: for each server (or project) maintain a list of buffers, add buffers to this list when they're created, remove them when they're killed -- and on the removal, kill the project if it was the last one.

But there is another option in a different direction.  How much time does it take you to run that function with your 2.5k buffers?  -- If it's short enough to not be noticeable when running once, then I could move the work to an idle timer.  This could also improve things if you do something like open just one file and work on it, kill it and immediately open another file from the same project.  If it's done on an idle timer, then in such cases the server wouldn't need to be killed and restarted.

josteink:
I agree with @lddubeau here:

* that's a nice general function!
* but the function we attach to the `kill-buffer-hook` only needs to check the state of one particular server, so it really shouldn't do anything more.

If you could implement that, I don't think there would be any need to debate the "extremeness" of the solution, and everyone would be happy :smile:

elibarzilay:
I'm happy that it's wanted, so I'll make it better.

However, I kinda convinced myself there that the idle-timer way is better.  Two modes of work that are better with that: open a directory, enter a file and edit, kill and enter a different one, repeat.  This is something that I often do, and I expect to be annoyed by constantly paying a price for starting the server again and again and will probably keep a buffer open just to keep it alive.  Another mode is having a long running Emacs, and you might not want to have servers go away immediately, just during the night when you're not working -- and an idle timer would make it possible by using a long timer.

But combining this with the minimal work thing will be a bit more complicated (need to find and cancel timers when you enter a buffer), and possibly conflict with the nature of idle-timers (they're always global, so mixing them with a project-specific action can be confusing).

So @lddubeau: can you see how much time it takes to scan your huge list of buffers?

elibarzilay:
Revised this to use the idle timer as in the last comment, with a custom option that can disable it completely.

elibarzilay:
@josteink, @lddubeau: ping about this one.  I changed the implementation to an idle timer which can set to whatever so many buffers shouldn't be a problem, and it can also disable it completely (which I can set as the default which would make it a no-op extension for people who won't set it explicitly).

josteink:
Sorry about the delay in getting this expedited. I honestly don't feel this is an area where I have the greatest in depth knowledge about how things currently work in tide, and even less why things work the way they do.

To avoid introducing errors/regressions, if it's OK, I'd rather have someone else review and merge this one.

tide.el
```diff
@@ -737,6 +743,29 @@ in the npm global installation.  If nothing is found use the bundled version."
   (remhash project-name tide-tsserver-unsupported-commands)
   (remhash project-name tide-project-configs))
 
+(defvar tide--project-cleanup-timer nil)
+(defun tide-cleanup-dead-projects* ()
```
ananthakumaran: why do we use `*` here? if it's just of naming concern could we use something like `tide-schedule-dead-projects-cleanup`?

tide.el
```diff
@@ -737,6 +743,29 @@ in the npm global installation.  If nothing is found use the bundled version."
...
+    (dolist (b (buffer-list))
+      (-when-let (proj (with-current-buffer b
+                         (and (bound-and-true-p tide-mode)
+                              (tide-project-name))))
+        (cl-pushnew proj live-projects)))
+    (-when-let (to-kill (-difference (hash-table-keys tide-servers) live-projects))
+      (message "Cleaning up %d projects..." (length to-kill))
+      (dolist (proj to-kill) (tide-cleanup-project proj))
+      (sit-for 5)
+      (message nil))))
```
ananthakumaran: what is this `message` for?

tide.el
```diff
@@ -737,6 +743,29 @@ in the npm global installation.  If nothing is found use the bundled version."
...
+  (setq tide--project-cleanup-timer nil)
+  (let ((live-projects '()))
+    (dolist (b (buffer-list))
+      (-when-let (proj (with-current-buffer b
+                         (and (bound-and-true-p tide-mode)
+                              (tide-project-name))))
+        (cl-pushnew proj live-projects)))
+    (-when-let (to-kill (-difference (hash-table-keys tide-servers) live-projects))
+      (message "Cleaning up %d projects..." (length to-kill))
+      (dolist (proj to-kill) (tide-cleanup-project proj))
```
ananthakumaran: I am trying to understand what's the expectation here.
```elisp
(defun tide-cleanup-project (project-name)
  (tide-each-buffer project-name
                    (lambda ()
                      (tide-cleanup-buffer-callbacks)))
  (remhash project-name tide-servers)
  (remhash project-name tide-tsserver-unsupported-commands)
  (remhash project-name tide-project-configs))
```

We are calling tide-cleanup-project which goes through each buffer again (we know for a fact it won't find any) and then it cleans up the global data structures. Where do we exactly stop the server process? and what is the need for `sit-for` in the next line. Maybe I am missing some logic here.

tide.el
elibarzilay: There are two functions, `tide-cleanup-dead-projects` which performs the cleaning *now*, and is also an interactive function.  `tide-cleanup-dead-projects*` is indeed scheduling the first, but I followed a common CL pattern where `*` is used for a related-but-slightly-different version of a function (specifically, it's not interactive, so the name is not user-friendly).

This is just a convention, so I don't mind changing it to whatever...

tide.el
elibarzilay: The first `message` indicates doing the work, the second one with the `nil` argument will clear out the messages area.  Without it, if you have a cleanup that is done during the night, the "Cleaning" message will stick around annoyingly for something that should be happening in the background.

tide.el
elibarzilay: Bah, you're right -- I missed the fact that that `tide-cleanup-project` does not kill the actual process!  I added that now to fix it.

The `sit-for` is purely cosmetic like the `(message nil)` that it precedes; just waits a bit for the message to be visible if you happen to look at it when it does the cleaning.  (It's a `sit-for`, which means that it will continue immediately if there is any interaction.)

tide.el
elibarzilay: (Leaving this open, in case you'd rather it's not done.)

elibarzilay:
@josteink, re force-pushing, I did the last fix (due to @ananthakumaran's review) and did not squash it; but the thing that makes me feel uneasy about it is that if you were to click merge now, then the commit would stick there -- and I personally like to keep commits clean when they're merged...

(And unfortunately I don't see a good way out of this: I can ask you to let me rebase & squash things before merging but then it's something extra for you to remember; or if you let GH do the squashing then the result will be a bad commit message...)

josteink:
> then the commit would stick there -- and I personally like to keep commits clean when they're merged...

No worries. Github supports squash-and-commit as a standard way of merging PRs with a lot of rounds back and forwards:

Screenshot-ma  02  des  19:31:15 +0100 2019

No  need for you to do anything.

elibarzilay:
The problem is the commit message is created.  I'm not sure if it was GH in the past, or maybe VSTS, but it used to just concatenate all of the commit messages.  I see that GH is a bit better than this now, as it lets you edit the squashed commit message -- but if you just click the button and accept the default text you end up with the same bogus text.

(And to clarify the bogosity of such concatenated text, in this PR I now have two commits that I intended to keep, and one is an "oops, fixed blah" -- if it would be squashed with the default concatenated commit message, that "oops" text will be there, though it has no value for anyone looking at the history later.  Note that I'm not arguing *against* squashing (or rebaasing) -- since without it, the "oops" commit would be there as is and therefore relevant, but it's relevant for my own process in making the PR, and the fact that I did 200 commits implementing something is most likely uninteresting for anyone other than me -- and more importantly -- for things like understanding the current code, or bisecting, or whatever...)

tide.el
```diff
@@ -737,6 +743,31 @@ in the npm global installation.  If nothing is found use the bundled version."
...
+    (dolist (b (buffer-list))
+      (-when-let (proj (with-current-buffer b
+                         (and (bound-and-true-p tide-mode)
+                              (tide-project-name))))
+        (cl-pushnew proj live-projects)))
+    (-when-let (to-kill (-difference (hash-table-keys tide-servers) live-projects))
+      (message "Cleaning up %d projects..." (length to-kill))
+      (dolist (proj to-kill)
+        (delete-process (process-buffer (gethash proj tide-servers)))
+        (tide-cleanup-project proj))
```
ananthakumaran: We might not have to call this again. Killing the process should trigger `tide-net-sentinel` and it will cleanup the project. You could confirm this by checking the `tsserver exits` messages.

elibarzilay: Yes, but this would be unreliable if the process is in an unexpected state because of some problem (and in such a state a cleanup is even more desirable).  So I preferred it to be safer in that case...

ananthakumaran: What do you mean by unreliable? My understanding is that the sentinel function should get called regardless of how the process exits. Emacs and tsserver are running as two separate processes, so any kind of issue in tsserver shouldn't affect what get executed in Emacs process.

The reason I prefer to avoid `tide-cleanu-project` is that it loops through the entire list of buffers.

elibarzilay: Well, it's the part that worries me, if you look at the documentation, is "*changes status for any reason, including signals*"...  Right now, the implicit assumption is that it's called only when the process dies, but what if other signals & behaviors get added later?  Also, I worry about situations where things get unstable and Emacs doesn't get the subprocess signal, or fails to act on it or whatever.

> The reason I prefer to avoid `tide-cleanu-project is` that it loops through the entire list of buffers.

Good point.  I'm calling it only if there are no buffers associated with a project, but it still scans the whole buffer list.  How about splitting the `remhash` calls into a `tide-cleanup-project-data`, which `tide-cleanup-project` (and my code) will call?

elibarzilay: Alternative of the above, and unrelated to this PR in particular: refactor the whole project info thing to simplify such issues: make a project object that holds properties: `:name`, `:root`, `:server`, `:config`, `:unsupported-commands`, and probably more.  The result would be more "ELisp-y", and it would make it easier to implement the cleanup code from this PR, as well as improving things like `tide-cleanup-project` since there won't be a need to scan all buffers.

(But this is much more ambitious and I'm guessing that such a refactor would scare people...)

ananthakumaran: > Good point. I'm calling it only if there are no buffers associated with a project, but it still scans the whole buffer list. How about splitting the remhash calls into a tide-cleanup-project-data, which tide-cleanup-project (and my code) will call?

agree on this.

> Alternative of the above, and unrelated to this PR in particular: refactor the whole project info thing to simplify such issues: make a project object that holds properties: :name, :root, :server, :config, :unsupported-commands, and probably more. The result would be more "ELisp-y", and it would make it easier to implement the cleanup code from this PR, as well as improving things like tide-cleanup-project since there won't be a need to scan all buffers.

But that should probably done as a separate PR and as you mentioned there would be more concern about introducing regression

elibarzilay: @ananthakumaran, I finally did that.

elibarzilay:
@ananthakumaran, Thanks!

(@josteink, this is what I talked about re merges...)

PR #346, opened 2019-11-14
Some fixes

elibarzilay:
Cleaning up some more code (including chopping huge lines); and get rid of all compiler warnings.
See commit messages for details.

### DIFF
--- a/tide.el
+++ b/tide.el
@@ -202,7 +202,7 @@ errors and tide-project-errors buffer."
 
 (defcustom tide-default-mode "TS"
   "The default mode to open buffers not backed by files (e.g. Org
-  source blocks) in."
+source blocks) in."
   :type '(choice (const "TS") (const "TSX") (const "JS")(const  "JSX"))
   :group 'tide)
 
@@ -272,8 +272,8 @@ this variable to non-nil value for Javascript buffers using `setq-local' macro."
 
 (defun tide-buffer-file-name ()
   "Returns the path to either the currently open file or the
-  current buffer's parent. This is needed to support indirect
-  buffers, as they don't set `buffer-file-name' correctly."
+current buffer's parent. This is needed to support indirect
+buffers, as they don't set `buffer-file-name' correctly."
   (buffer-file-name (or (and (bound-and-true-p edit-indirect--overlay)
                              (overlay-buffer edit-indirect--overlay))
                         (and (bound-and-true-p org-src--overlay)
@@ -358,8 +358,8 @@ ones and overrule settings in the other lists."
 
 (defun tide-get-file-buffer (file &optional new-file)
   "Returns a buffer associated with a file. This will return the
-  current buffer if it matches `file'. This way we can support
-  temporary and indirect buffers."
+current buffer if it matches `file'. This way we can support
+temporary and indirect buffers."
   (cond
    ((equal file (tide-buffer-file-name)) (current-buffer))
    ((file-exists-p file) (find-file-noselect file))
@@ -427,9 +427,9 @@ ones and overrule settings in the other lists."
 (defun tide-first-buffer (project-name fn)
   "Callback FN for the first buffer within PROJECT-NAME with tide-mode enabled."
   (-when-let (buffer (-first (lambda (buffer)
-                             (with-current-buffer buffer
-                               (and (bound-and-true-p tide-mode)
-                                    (equal (tide-project-name) project-name))))
+                               (with-current-buffer buffer
+                                 (and (bound-and-true-p tide-mode)
+                                      (equal (tide-project-name) project-name))))
                              (buffer-list)))
     (with-current-buffer buffer
       (funcall fn))))
@@ -446,7 +446,7 @@ ones and overrule settings in the other lists."
 (defun tide-current-offset ()
   "Number of characters present from the begining of line to cursor in current line.
 
-offset is one based."
+Offset is one based."
   (1+ (- (point) (line-beginning-position))))
 
 (defun tide-offset (pos)
@@ -899,7 +899,7 @@ implementations.  When invoked with a prefix arg, jump to the type definition."
       (pop-to-buffer (tide-get-file-buffer file)))
     (tide-move-to-location (plist-get filespan :start))
     (when should-recenter-p
-        (recenter))))
+      (recenter))))
 
 (defalias 'tide-jump-back 'pop-tag-mark)
 
@@ -1415,9 +1415,9 @@ always be formatted as described above."
 (defun tide-compose-comparators (cmp1 cmp2)
   "A helper function that composes two comparators CMP1 and CMP2."
   (lambda (a b)
-    (or (funcall cmp1 a b)
-        (if (not (funcall cmp1 b a))
-            (funcall cmp2 a b)))))
+    (cond ((funcall cmp1 a b) t)
+          ((funcall cmp1 b a) nil)
+          (t (funcall cmp2 a b)))))
 
 (defun tide-compare-completions-basic (completion-a completion-b)
   "Compare COMPLETION-A and COMPLETION-B based on their `sortText' property.
@@ -1688,9 +1688,7 @@ number."
             (if (or (eq usage nil) (tide-is-identical-reference usage reference))
                 (setq usage reference)
               (setq multiple t)))))
-    (if (and (not multiple) usage definition)
-        usage
-      nil)))
+    (and (not multiple) definition usage)))
 
 (defun tide-references ()
   "List all references to the symbol at point."
@@ -2010,18 +2008,15 @@ code-analysis."
   (let* ((result '())
          (resolved nil)
          (err nil))
-    (cl-flet
-        ((resolve ()
-           (unless resolved
-             (if err
-                 (progn
-                   (setq resolved t)
-                   (funcall cb err))
-               (when (and (plist-member result :syntaxDiag)
-                          (plist-member result :semanticDiag)
-                          (plist-member result :suggestionDiag))
-                 (setq resolved t)
-                 (funcall cb `(:body (,result) :success t)))))))
+    (cl-flet ((resolve ()
+                (cond (resolved nil)
+                      (err (setq resolved t)
+                           (funcall cb err))
+                      ((and (plist-member result :syntaxDiag)
+                            (plist-member result :semanticDiag)
+                            (plist-member result :suggestionDiag))
+                       (setq resolved t)
+                       (funcall cb `(:body (,result) :success t))))))
       (tide-send-command
        "syntacticDiagnosticsSync"
        `(:file ,(tide-buffer-file-name))
@@ -2229,9 +2224,9 @@ code-analysis."
          (save-excursion
            (goto-char (point-max))
            (let ((inhibit-read-only t)
-		  (file-name (tide-plist-get response :body :file))
-		  (diagnostics (tide-plist-get response :body :diagnostics))
-		  (event-type (plist-get response :event)))
+                 (file-name (tide-plist-get response :body :file))
+                 (diagnostics (tide-plist-get response :body :diagnostics))
+                 (event-type (plist-get response :event)))
 	     (unless (and (string-equal event-type "suggestionDiag") tide-disable-suggestions)
 	       (pcase event-type
 		 ("syntaxDiag"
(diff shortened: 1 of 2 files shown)


tide.el
```diff
@@ -568,8 +572,8 @@ LINE is one based, OFFSET is one based and column is zero based"
 
 (defun tide-dispatch (response)
   (cl-case (intern (plist-get response :type))
-    ('response (tide-dispatch-response response))
-    ('event (tide-dispatch-event response))))
+    ((response) (tide-dispatch-response response))
+    ((event) (tide-dispatch-event response))))
```
josteink: Could you elaborate on this one? I'm not sure I understand this change, or why this should be a change for the better.

tide.el
```diff
@@ -1408,7 +1410,8 @@ This function is used for the basic completions sorting."
             (rx (or (and "import" (1+ space) (or ?\" ?') (0+ (not (any ?\" ?'))))
                     (and "from" (1+ space) (or ?\" ?') (0+ (not (any ?\" ?'))))
                     (and "import(" (or ?\" ?') (0+ (not (any ?\" ?'))))
-                    (and "require(" (or ?\" ?') (0+ (not (any ?\" ?'))))))))
+                    (and "require(" (or ?\" ?') (0+ (not (any ?\" ?'))))))
+            nil))
```
josteink: What's the value of the `nil` added here? Silencing warnings?

tide.el
```diff
@@ -1486,7 +1489,7 @@ This function is used for the basic completions sorting."
         (prefix (get-text-property 0 'prefix name)))
 
     (-when-let (insert-text (plist-get completion :insertText))
-      (when (looking-back (rx-to-string name))
+      (when (looking-back (rx-to-string name) nil)
```
josteink: Same or this one, I guess. When was the `LIMIT` parameter introduced and/or made mandatory anyway? 

I guess as long as the build is "green", this change doesn't cause any compatibility issues, and should be OK.

tide.el
```diff
@@ -2379,6 +2379,22 @@ highlights from previously highlighted identifier."
...
+      (progn
+        (tide--hl-set-timer)
+        ;; Unhighlight if point moves off identifier
+        (add-hook 'post-command-hook #'tide--hl-identifiers-post-command-hook nil t)
+        ;; Unhighlight any time the buffer changes
+        (add-hook 'before-change-functions #'tide--hl-identifiers-before-change-function nil t))
+    (remove-hook 'post-command-hook #'tide--hl-identifiers-post-command-hook t)
+    (remove-hook 'before-change-functions #'tide--hl-identifiers-before-change-function t)
+    (tide-unhighlight-identifiers)))
+
```
josteink: What does moving this autoload around solve? Not against it, but I can't see it documented or mentioned anywhere 😅

josteink:
Nevermind the comments so far. I failed to read the commit-message detailing the things I was wondering about.

(But that's what you get for making so big/long commits! 😄  )

tide.el
```diff
@@ -1415,9 +1415,9 @@ always be formatted as described above."
 (defun tide-compose-comparators (cmp1 cmp2)
   "A helper function that composes two comparators CMP1 and CMP2."
   (lambda (a b)
-    (or (funcall cmp1 a b)
-        (if (not (funcall cmp1 b a))
-            (funcall cmp2 a b)))))
+    (cond ((funcall cmp1 a b) t)
+          ((funcall cmp1 b a) nil)
+          (t (funcall cmp2 a b)))))
```
josteink: Is this re-implementation strictly speaking doing the same thing? I'm not sure I would want to trust that without any unit-tests proving so.

tide.el
```diff
@@ -1688,9 +1688,7 @@ number."
             (if (or (eq usage nil) (tide-is-identical-reference usage reference))
                 (setq usage reference)
               (setq multiple t)))))
-    (if (and (not multiple) usage definition)
-        usage
-      nil)))
+    (and (not multiple) usage definition usage)))
```
josteink: I realize this is more an *opinion* than an objective measure, but I think this is crossing the line from "overly verbose" code into "clever" code.

To the tired/lazy/unskilled eye it's not obvious that this function will return `usage` upon all conditions being met.

If the aim is to slim down the `(if (..) value nil`-form to something more compact, I'd rather have a `(when (...) usage)` form to be explicit about what we plan to return.

tide.el
```diff
@@ -2402,26 +2397,25 @@ highlights from previously highlighted identifier."
...
+                     (oend   (progn (unless (= start-line (plist-get end :line))
+                                      ;; Identifiers shouldn't span lines.
+                                      (error "identifier unexpectedly spans lines"))
+                                    (forward-char (- (plist-get end :offset)
+                                                     (plist-get start :offset)))
+                                    (point)))
+                     (overlay (make-overlay ostart oend)))
+                (setq current-line start-line)
+                (overlay-put overlay 'tide-overlay 'sameid)
+                (overlay-put overlay 'face 'tide-hl-identifier-face)))))))))
```
josteink: I was never a fan of `cl-loop`. I think this is much better 😃

tide.el
```diff
@@ -960,7 +977,8 @@ Noise can be anything like braces, reserved keywords, etc."
                           (-map (lambda (navto-item) (plist-get navto-item :name))
                                 navto-items)))))
                   t) nil t default))
-      (let ((navto-item (-find (lambda (navto-item) (string-equal completion (plist-get navto-item :name))) last-completions)))
+      (let ((navto-item (-find (lambda (navto-item) (string-equal completion (plist-get navto-item :name)))
+                               last-completions)))
```
josteink: Personally I'd shorten this line by introducing a line-break after the lambda-declaration to make a clearly visible scope for that function.

tide.el
```diff
@@ -2165,7 +2205,7 @@ code-analysis."
 (defun tide-project-errors-buffer-name ()
   (format "*%s-errors*" (tide-project-name)))
 
-(defun tide-display-erros (file-names)
+(defun tide-display-errors (file-names)
```
josteink: Subtle! Good catch!

josteink:
Hey there and thanks for another patch! 

All in all lots of small changes which I honestly think helps improve readability, but as usual I do have some comments here and there.

Please note I accidentally asked for some clarifications, which I later actually found clarifications for (in the commit message no less!). My bad!

I've deleted those comments from Github, so please just check what comments remain, and ignore whatever you've received by email if it's no longer up here 😃

tide.el
elibarzilay: 1. First, I'm assuming that the change is indeed good -- it's a subjective one, but I think that my version is easier to follow.
2. My usual way to make such changes is to do it once, then start from the previous code and do it again, then compare the two edits to see that the result is the same.  (Sounds strange, but I just mechanically reduce the different options in the same way so the result should be the same...)
4. `cmp2` is called when both of the `cmp1` calls returned `nil`: which means that there's no decision for `a` being smaller or bigger than `b` wrt `cmp1`.  It's a common thing to do when composing comparators in lisps since it's common to find binary boolean comparators rather than ternary -1/0/+1 ones.  Also, it's IMO clearer when `cmp2` is called in my version than the original...
5. Finally, here's code that tests equality of both versions, first an exhaustive one, and then a random one:

```emacs-lisp
(defun compose1 (cmp1 cmp2)
  (lambda (a b)
    (cond ((funcall cmp1 a b) t)
          ((funcall cmp1 b a) nil)
          (t (funcall cmp2 a b)))))
(defun compose2 (cmp1 cmp2)
  (lambda (a b)
    (or (funcall cmp1 a b)
        (if (not (funcall cmp1 b a))
            (funcall cmp2 a b)))))

(dolist (cmp1 (list #'< #'<= #'> #'>= #'=))
  (dolist (cmp2 (list #'< #'<= #'> #'>= #'=))
    (let ((cc1 (compose1 cmp1 cmp2))
          (cc2 (compose2 cmp1 cmp2)))
      (dotimes (x 5)
        (dotimes (y 5)
          (unless (equal (funcall cc1 x y) (funcall cc2 x y))
            (error "poof")))))))

(defun choose (&rest l) (nth (random (length l)) l))
(dotimes (i 1000)
  (let* ((cmp1 (choose #'< #'<= #'> #'>= #'=))
         (cmp2 (choose #'< #'<= #'> #'>= #'=))
         (cc1  (compose1 cmp1 cmp2))
         (cc2  (compose2 cmp1 cmp2)))
    (dotimes (j 1000)
      (let ((x (choose 1 2 3 4 5 6 7 8))
            (y (choose 1 2 3 4 5 6 7 8)))
        (unless (equal (funcall cc1 x y) (funcall cc2 x y))
          (error "poof"))))))
```

tide.el
elibarzilay: Yes, it's inherently a subjective opinion since the previous code is correct too.  So if you have a strong preference for the former then I should remove it, but I hope to convince otherwise...

The thing is that `and` (and `or`) in lisps are defined to return the last non-nil (or first non-nil) result, which makes usages of `nil` / `t` in conditionals suspicious.  For example, if you look at [the style guide rules](https://pl.barzilay.org/style-guide.html) I give students in my class, you'll see it listed under the "Avoid using redundant `if`s with literal..." bullet (it's about Racket though, a different Lisp dialect).

As for `when` -- it would be *worse* here, since it signifies doing a side-effect rather than returning a boolean.  In ELisp it's of course still correct to use it, but in Schemes `(when #f ...)` usually return some void/undefined value instead of false.

(And like I said, if this doesn't convince, then I'll drop it.)

tide.el
elibarzilay: Yeah, it's a source of many flamewars in CL, therefore more rare in other Lisps...

tide.el
elibarzilay: Well, if you do that, you'll still need a newline before the `last-completions`...  In any case that's even more subjective so I'll yield to personal preferences of other people.

(As a sidenote, I think that such preference for always breaking a line inside `lambda`s is an influence of curly languages (like JS) -- in traditional idiomatic Lisps there's no reason to so since there are many very-short `lambda`s.)

tide.el
josteink: 1. Ofcourse. So does everyone 😄 

3. Ok. That's fair enough. My bad for not doing good enough research.

4. That's a test for checking that they both behave the same way. What would be even better was if you instead of doing that, made a test for how `tide-compose-comparators` should behave in general. That way we will be protected from future regressions if someone makes a bad change to this function. Could you add that?

tide.el
josteink: OK. Fair enough. But in that case, we could remove the duplicate check for usage, by instead making this: `(and (not multiple) definition usage)))`. Right now `usage ... usage` looks like a bug :)

tide.el
josteink: Ok. Some good points. We can leave it as is. No change needed on my part.

josteink:
Hey there. Sorry about the delay in following up on this review.

You make lots of good points, so consider most of the nitpicks from previous review nullified.

That said: I've posted a few final comments, and I hope you're OK with those. If you can make those changes, I'm sure we can get this thing merged 👍

tide.el
elibarzilay: Oh, good catch -- yes, I didn't notice that it's repeated...  Fixed and will push when I push the requested test.

tide.el
elibarzilay: Well, I have this quick sketch for two different styles of tests, which one would you prefer to keep?  (Or keep both, or something different; I have no strong opinions on tests...)

```emacs-lisp
(ert-deftest tide-compose-comparators-tests ()
  "Tests that confirm the implementation of `tide-compose-comparators'."
  (let ((<< (tide-compose-comparators (lambda (x y) (< (car  x) (car  y)))
                                      (lambda (x y) (< (cadr x) (cadr y)))))
        (01-pairs '((0 0) (0 1) (1 0) (1 1))))
    (dolist (perm (-permutations 01-pairs))
      (should (equal (sort perm <<) 01-pairs))))
  (let ((<< (tide-compose-comparators (lambda (x y) (< (car  x) (car  y)))
                                      (lambda (x y) (< (cadr x) (cadr y))))))
    (should (equal nil (funcall << '(0 1) '(0 1))))
    (should (equal nil (funcall << '(0 1) '(0 0))))
    (should (equal t   (funcall << '(0 0) '(0 1))))
    (should (equal nil (funcall << '(1 0) '(0 0))))
    (should (equal t   (funcall << '(0 0) '(1 0))))
    (should (equal nil (funcall << '(1 0) '(0 1))))
    (should (equal t   (funcall << '(0 1) '(1 0))))))
```

josteink: While the first approach would be the most long term "scalable" way to handle the test, it still requires me to *think* about what is actually being tested. For me, that's never a good sign.

The second version, while "dumber", uses less magic and IMO makes the test itself much more obvious. I'll prefer that one, unless anyone else objects.

elibarzilay: Meanwhile, I pushed an update with that second test included.

josteink:
LGTM. Thanks for the patch, and thanks for hanging out in there!

Merged :)

PR #347, opened 2019-11-14
Improve reference buffer

elibarzilay:
As mentioned in #278 (which this supersedes).

* Rewrite much of `tide-insert-references`, so the code is simpler and
  the rendering is better.  Use `line-prefix` to display line numbers,
  add `tide-line-reference` on the whole line to make it easy to jump
  from anywhere on a line.

* Add `tide-goto-line-reference` that uses the above to jump to the
  referenced file at the location that we were at.  Use it as a binding
  for `RET`, leaving `tide-goto-reference` the same for mouse clicks.

### DIFF
--- a/tide.el
+++ b/tide.el
@@ -1616,13 +1616,21 @@ Move back when used from a shifted key binding."
   (-when-let (reference (get-text-property (point) 'tide-reference))
     (tide-jump-to-filespan reference nil t)))
 
+(defun tide-goto-line-reference ()
+  "Jump to the corresponding location in the referenced file."
+  (interactive)
+  (-when-let* ((ref (get-text-property (point) 'tide-line-reference))
+               (col (- (point) (line-beginning-position))))
+    (tide-jump-to-filespan ref nil t)
+    (goto-char (+ (line-beginning-position) col))))
+
 (defvar tide-references-mode-map
   (let ((map (make-sparse-keymap)))
     (define-key map (kbd "n") #'tide-find-next-reference)
     (define-key map (kbd "p") #'tide-find-previous-reference)
     (define-key map (kbd "TAB") #'tide-cycle-next-reference)
     (define-key map (kbd "<backtab>") #'tide-cycle-previous-reference)
-    (define-key map (kbd "RET") #'tide-goto-reference)
+    (define-key map (kbd "RET") #'tide-goto-line-reference)
     (define-key map [mouse-1] #'tide-goto-reference)
     ;; taken from grep.el
     (define-key map (kbd "SPC") #'scroll-up-command)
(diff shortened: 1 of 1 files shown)


josteink:
Hey there and thanks for another patch.

This looks like a big change, and unfortunately, I don't have that much time to do a deep in-depth review of this right now.

Maybe someone else could take a look? @ananthakumaran ? @lddubeau ?

lddubeau:
@josteink  No time for this right now, I'm afraid. A wire powering the oven portion of our kitchen stove started melting down. Luckily, I noticed the smell, opened the stove, located the wire, and disconnected it right away. It was a fire hazard. The range is certainly over 20-year old, maybe 30. (Bought by previous owners of the house, so I don't know exactly.) I've repaired appliances many times, but it makes no sense to try repairing *this one*. So we need a new stove. The old stove was a JennAir downdraft stove. Downdraft is crap. So we're converting to updraft, which is a major job. The whole operation involves duct work, electrical work, drywall repairing and patching work, painting, modifying the kitchen counter in place (because *of course* the old stove was a bit smaller than the standard stove size in the US) and roof work (for the new duct vent). I'm responsible for everything, except painting, which my wife will do, and except the roof work, which we have a contractor do. And we have a complete master bathroom remodel (complete gutting) which is currently on hold due to the emergency with the range, and for which I'm also responsible, except for electrical and plumbing.

So my days right now are split between my day job and working on the house, with pretty much no time for anything else.

josteink:
@lddubeau : Not having time is a very human condition. No need to excuse yourself. I was just asking, and I never ask a yes/no question where I can't live with both answers 😄

josteink:
@lddubeau : Not having time is a very human condition. No need to excuse yourself. I was just asking, and I never ask a yes/no question where I can't live with both answers 😄 

Good luck on your house!

elibarzilay:
FWIW, I can do a skype/hangouts/whatever call to go over this if it'll help.  (But I found a nice way to organize my directories to use my own patches so I can just wait.)

ananthakumaran:
I can't promise any timeline but will try to review when time permits. I usually give more preference to bug fix than refactoring. So if there are any bug fixes, you could split it into a separate PR and I will get to it faster.

elibarzilay:
I could in theory try to split it into chunks, but it would take a while to do that -- if only because each PR would take a while to go in.  Also, the main part -- the edits in `tide-insert-references` -- are likely to stay as one PR...  If that sounds useful, I can try to do that.

elibarzilay:
Actually, an almost equivalent alternative is to take *this* PR, and split it into multiple commits.  Might make it easier to follow if each step is documented in its own commit (but TBH the GH UI really wants a review of a whole PR).

elibarzilay:
Moved a bunch of stuff out to separate smaller PRs.

josteink:
Great stuff. Will see if I can get some of those smaller PRs expedited swiftly.

josteink:
Thanks for splitting up the PR!

With all the other small changes moved into separate PRs, this one just became much more manageable to test and review.

Having just tried these changes, I think they work pretty well, much in line with how people might expect them to work, based on experience with `grep` or `occur`.

If nobody else has any objections, I'm inclined to merge this as is. Do you have any issues with this PR in its current form, @ananthakumaran ?

josteink:
Thanks for another good patch. Keep them coming!

elibarzilay:
Thanks!

PR #349, opened 2019-11-23
Better `tide-file` face definition

elibarzilay:
Make `tide-file` inherit directly from `font-lock-type-face` instead of
`dired-header`, since that doesn't exist until `dired` is used (so
opening the references before dired gets no highlights).

### DIFF
--- a/tide.el
+++ b/tide.el
@@ -146,7 +146,7 @@ errors and tide-project-errors buffer."
   :group 'tide)
 
 (defface tide-file
-  '((t (:inherit dired-header)))
+  '((t (:inherit font-lock-type-face)))
   "Face for file names in references output."
   :group 'tide)
 


josteink:
So the rationale for this change is pretty good. I wasn't aware of that bug. As such, good catch! 😄 

About the actual change itself... I'm not sure I think this is the best solution. I personally think having a path with several subfolders highlighted like a type looks kinda "weird".

Wouldn't it be just much simpler to `(require 'dired)` to ensure the face is available?

elibarzilay:
I didn't change the actual face -- `dired-header` is defined as `font-lock-type-face`, and I just made it use the latter directly.

You could argue that loading `dired` and then using it is better in case the face definition changes in the future, but (a) I doubt that can ever happen, (b) if you really want to ensure the same face, then there is also `dired-header-face` which is a variable defined as `dired-header` and I'm guessing that *that* is what's intended to be customized instead of the face itself.  (This is coming from old times before faces became more of an actual thing.)

Because of that, I thought that just using `font-lock-type-face` directly is overall simpler.

josteink:
> I didn't change the actual face -- dired-header is defined as font-lock-type-face, and I just made it use the latter directly.

Ok that's funny. If so that's a difference introduced in *my* local config, where I probably customize `font-lock-type-face` after `dired` is loaded. I guess that actually makes it 2 bugs in this one situation 😄 

If that's the case, I really can't see any big argument against this. It's probably just me which will encounter a visual change (based on a load-order error on my config's part!), and most users will only experience a consistent face in the references buffer.

Consider it merged.

PR #350, opened 2019-11-23
Add some grep-like bindings

elibarzilay:
Copied from "grep.el".

### DIFF
--- a/tide.el
+++ b/tide.el
@@ -1607,6 +1607,10 @@ This function is used for the basic completions sorting."
     (define-key map (kbd "p") #'tide-find-previous-reference)
     (define-key map (kbd "RET") #'tide-goto-reference)
     (define-key map [mouse-1] #'tide-goto-reference)
+    ;; taken from grep.el
+    (define-key map (kbd "SPC") #'scroll-up-command)
+    (define-key map (kbd "S-SPC") #'scroll-down-command)
+    (define-key map (kbd "DEL") #'scroll-down-command)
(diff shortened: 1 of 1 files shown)


tide.el
```diff
@@ -1607,6 +1607,10 @@ This function is used for the basic completions sorting."
     (define-key map (kbd "p") #'tide-find-previous-reference)
     (define-key map (kbd "RET") #'tide-goto-reference)
     (define-key map [mouse-1] #'tide-goto-reference)
+    ;; taken from grep.el
+    (define-key map " " 'scroll-up-command)
+    (define-key map [?\S-\ ] 'scroll-down-command)
+    (define-key map "\^?" 'scroll-down-command)
```
josteink: We're mostly using `(kbd " ")` macros throughout this mode. It would be nice if you did that here too.

Please also prefix function-quotes/refernces with `#'` instead of `'`.

josteink:
Hey there and thanks for another patch :)

I approve of the overall changes here, but I have some issues with the syntax. If you can get that fixed, I'll get this merged in no time.

tide.el
elibarzilay: Sure, I'll do both now.  (BTW, the motivation for *not* doing it in the first place was to keep it identical to the snippet in `grep.el`.)

elibarzilay: Done now.

josteink:
Looks good to me. Merged! 😃

PR #351, opened 2019-11-23
Improve reference navigation functions

elibarzilay:
* Abstract moving forward/backward in a separate function, which also
  fixes a bug when moving back from the middle of an identifier.

* Make `tide-next-reference-function` use `set-window-point` when a
  window is shown, otherwise things get confusing since the point can
  reset itself back to the one displayed in a window.  (When navigating
  to a new buffer, the reference buffer can be hidden, and at that time
  the value of `(point)` reverts to wherever the visible point was.)

* Add `tide-cycle-next-reference` that cycles back at the buffer end(s).
  Bind it to `TAB`.

### DIFF
--- a/tide.el
+++ b/tide.el
@@ -1561,39 +1561,54 @@ This function is used for the basic completions sorting."
 
 ;;; References
 
+(defun tide--next-reference (pos arg &optional cyclep)
+  "Move to the ARG-th next reference from pos, negative values go to previous
+ones.  Cycle around if CYCLEP is non-nil."
+  (let* ((nextp (>= arg 0))
+         (next  (if nextp #'next-single-property-change #'previous-single-property-change))
+         (n     (* 2 (abs arg))))
+    (when (get-text-property pos 'tide-reference)
+      (setq pos (or (funcall next pos 'tide-reference) pos)))
+    (when (or nextp (and (> pos (point-min))
+                         (get-text-property (1- pos) 'tide-reference)))
+      (setq n (1- n)))
+    (dotimes (_i n)
+      (setq pos (funcall next pos 'tide-reference))
+      (unless pos
+        (unless cyclep
+          (error "Moved %s reference" (if nextp "past last" "before first")))
+        (setq pos (funcall next (if nextp (point-min) (point-max))
+                           'tide-reference))))
+    (goto-char pos)))
+
 (defun tide-next-reference-function (n &optional reset)
   "Override for `next-error-function' for use in tide-reference-mode buffers."
   (interactive "p")
-
   (-when-let (buffer (get-buffer "*tide-references*"))
     (with-current-buffer buffer
-      (when reset
-        (goto-char (point-min)))
-      (if (> n 0)
-          (tide-find-next-reference (point) n)
-        (tide-find-previous-reference (point) (- n)))
+      (when reset (goto-char (point-min)))
+      (tide--next-reference (point) (or n 1))
+      (-when-let (window (get-buffer-window buffer))
+        ;; actually move to the point if the refs are shown
+        (set-window-point window (point)))
       (tide-goto-reference))))
 
 (defun tide-find-next-reference (pos arg)
   "Move to next reference."
   (interactive "d\np")
-  (setq arg (* 2 arg))
-  (unless (get-text-property pos 'tide-reference)
-    (setq arg (1- arg)))
-  (dotimes (_i arg)
-    (setq pos (next-single-property-change pos 'tide-reference))
-    (unless pos
-      (error "Moved past last reference")))
-  (goto-char pos))
-
+  (tide--next-reference pos arg))
+(defun tide-cycle-next-reference (pos arg)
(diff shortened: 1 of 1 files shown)


josteink:
Thanks for yet another patch.

I like this. This is very nice.

I only have one objection: For consistency/symmetry, shouldn't we also bind shift-TAB to going *back* one identifier?

elibarzilay:
In a modern editor that would make sense, but I think that in Emacs there are a bunch of places where `TAB` is bound and `S-TAB` isn't.  (Some random examples that I can think of: `man`, `info`, and `magit`.

One thing that could be a compromise is to make it move back when `this-command-keys-shift-translated` is on, so it will move back when using a shifted binding.  So, if you don't have `S-tab` bound explicitly, using it will get translated to `tab`, which will call the function, which will notice the shift and move back instead.  I pushed that version now.

josteink:
> One thing that could be a compromise is to make it move back when this-command-keys-shift-translated is on, so it will move back when using a shifted binding. So, if you don't have S-tab bound explicitly, using it will get translated to tab, which will call the function, which will notice the shift and move back instead. I pushed that version now.

I like that. That's a really nice solution.

tide.el
```diff
@@ -1561,39 +1561,53 @@ This function is used for the basic completions sorting."
...
-      (error "Moved past last reference")))
-  (goto-char pos))
-
+  (tide--next-reference pos arg))
+(defun tide-cycle-next-reference (pos arg)
+  "Move to next reference, cycling back when reaching the last.
+Move back when used from a shifted key binding."
+  (interactive "d\np")
+  (tide--next-reference pos
+                        (if this-command-keys-shift-translated (-arg) arg)
```
josteink: Shouldn't this be `(- arg)`, that is with a space in between?

elibarzilay: Sorry, fell asleep before I could try it -- it's fixed now.

josteink:
Thanks for the update.

I've done some testing in three different environments (Windows/GUI, Linux/X11, Linux/CLI), and I may have wasted your time.

In all three environments. I've found shift-tab to be bound to `<backtab>`, so the check you're doing in code here might never actually trigger (at least on those platforms).

I guess if you've tested it and it works on OSX, we can keep it as it is. But if it doesn't work there either, maybe all this is just excess code we shouldn't be pulling in.

What do you think?

elibarzilay:
Yeah, I was hoping to not get into it, but that's indeed not a trivial issue...  This is going to be long, but see the bottom for the TL-DR question.

I actually have `S-tab` globally bound to some function I use a *lot* (switching between the two most recent buffers), and throughout the years I collected a bunch of bindings to keep it working -- mostly using `local-function-key-map` (which is the same mechanism that translates escape sequences on a terminal to keys).  It's navigating the few different names that it can have (`backtab` or `iso-lefttab`), and keeps it as `S-tab` instead.  (It's [this bit](https://github.com/elibarzilay/eliemacs/blob/master/include/keys.el#L76-L81), in case you're interested.)  (Oh, and I'm mostly on Linux or WSL.)  Also, the builtin code that explains the mess shows that it's because sometime (no clue where) there is a dedicated key, and the default setup maps `S-tab` to that key:

```emacs-lisp
;; Many keyboards don't have a `backtab' key, so by convention the user
;; can use S-tab instead to access that binding.
(define-key function-key-map [S-tab] [backtab])
```

This is unrelated to how to make it possible to use it, since even with a direct binding you'd run into the question of which key name to use in the binding.  I now see that there are some builtin files that use `backtab`, which I don't see since I keep the `S-tab` thing.

And finally, to try my code, you can manually bind it to some other key where it is straightforward to use the shifted version.

To summarize all of this, there are the following options:

1. Keep the current reverse-on-shift behavior "FWIW".
2. Go back to just having the plain behavior.
3. Go back as the above, and add a function for `backtab` that is reversed by default.

I don't have a strong opinion (and my setup will not be affected anyway) so I'll be happy to go with whichever you think is best.

MarcusRiemer:
Just as a short voice from the outside: I personally never got used to `n` and `p`, so having `TAB` and `S-TAB` working out of the box sounds great to me.

josteink:
@MarcusRiemer ... So lets just leave it as it is then? Option 1?

As long as it solves a real problem for real people out there, I'm happy with that.

And if so, and there's no other objections... I guess we're to merge? :smiley:

elibarzilay:
I'm guessing that option 3 (binding `backtab` to an in-reverse function) would be the closest to making it work out-of-the-box for most people, but I don't have a good way to verify this claim.  (Even if I get rid of my tweaks, I'm not working in a conventional setting...)

josteink:
@elibarzilay  If you're happy with that approach, I will be happy to test it.

elibarzilay:
@josteink -- I force-pushed it, so should be ready now.

josteink:
Works perfectly! Consider it merged!

elibarzilay:
Thanks!

PR #352, opened 2019-11-27
Find tsserver and tsc via exec-path

sandinmyjoints:
For people using nvm and https://github.com/rejeep/nvm.el, this will find the typescript executables installed globally for the particular version of node that is currently activated, if any. For example, when I've used nvm.el to activate v10.16.3, then it will find `/Users/william/.nvm/versions/node/v10.16.3/bin/tsserver`, which none of the existing methods would have found.

### DIFF
--- a/tide.el
+++ b/tide.el
@@ -666,6 +666,7 @@ npm local package directory, in the project root as defined by projectile, and
 in the npm global installation."
   (or
    (and tide-tscompiler-executable (expand-file-name tide-tscompiler-executable))
+   (executable-find "tsc")
    (tide-tscompiler-locater-npmlocal-projectile-npmglobal)))
 
 (defconst tide--tsserver "tsserver.js"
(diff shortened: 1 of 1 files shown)


josteink:
LGTM. Anyone else have any objections?

lddubeau:
The search logic has been so far that a local install trumps a global install. (This is also how `npm` works. A project that has `foo` installed locally will use the local version, no matter what may have been installed globally.) The proposed change diverges from that logic. Looking at the order of the checks, if the executable that is being sought is to be found in `PATH`, then that will be used, irrespective of what may be locally installed in the project being edited. I'd like to preserve the logic that local trumps global.

josteink:
> I'd like to preserve the logic that local trumps global.

That's a good point. I second that.

ananthakumaran:
I see couple of options

1) You can add it as a last option in `tide-tsserver-locater-npmlocal-projectile-npmglobal` and `tide-tscompiler-locater-npmlocal-projectile-npmglobal` so we will give it more preference than the bundled tsserver.

2) If #1 is not suitable for your case, you could provide your own function for `tide-tsserver-locator-function`. We don't have locator-function for tscompiler, but we would be happy to accept a PR which makes it configurable.

sandinmyjoints:
Thanks for the feedback, all. I'll look into making these changes soon.

PR #353, opened 2019-11-30
Add optional tide-jump-to-fallback

gamb:
Occasionally tide won't be able to find a definition. In cases like this it'd be nice to fallback to another jump method (e.g. [dumb-jump](https://github.com/jacktasia/dumb-jump)).

Here's an approach that I'm finding useful:
```elisp
(use-package tide
  :ensure t
  :after (typescript-mode company)
  :config
  (setq tide-jump-to-fallback 'dumb-jump-go)
  :hook ((typescript-mode . tide-setup)
         (typescript-mode . tide-hl-identifier-mode)
         (before-save . tide-format-before-save)))
```

### DIFF
--- a/tide.el
+++ b/tide.el
@@ -211,6 +211,11 @@ source blocks) in."
   :type 'boolean
   :group 'tide)
 
+(defcustom tide-jump-to-fallback #'tide-jump-to-fallback-not-given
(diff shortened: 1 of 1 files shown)


josteink:
LGTM!

PR #356, opened 2019-12-17
Fix writing temp file for a buffer

elibarzilay:
Need to write the whole file, otherwise things won't work when viewing a
narrowed part of the buffer.

### DIFF
--- a/tide.el
+++ b/tide.el
@@ -1212,7 +1212,9 @@ Noise can be anything like braces, reserved keywords, etc."
     (setq tide-buffer-dirty nil)
     (unless tide-buffer-tmp-file
       (setq tide-buffer-tmp-file (make-temp-file "tide")))
-    (write-region (point-min) (point-max) tide-buffer-tmp-file nil 'no-message)
+    (save-restriction
+      (widen)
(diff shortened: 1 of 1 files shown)


josteink:
Hey there, and thanks for another patch :)

Fundamentally I don't see anything *wrong* with this patch, but feel it lacks a little information about context.

What issues do you encounter which this fixes, and how do you trigger that condition?

Basically anything I can do to test that 1. the bug exists and 2. this fixes the bug.

Thanks.

ananthakumaran:
I think we use `(widen)` in most of the places, looks like we missed it here. @josteink to reproduce, you could select a region and run `M-x narrow-to-region`.

elibarzilay:
Yes, what @ananthakumaran said.  To be more specific, once you're working on a narrowed region, and it gets written to the temp file, that's what the tsserver gets to work with.

josteink:
Yeah I see. That construct is used all over.

So... I guess there's no objections with getting this merged then? :sweat_smile:

josteink:
No objections. Consider it merged!

Issue #308, opened 2019-03-21
add support for eslint


lddubeau:
Currently, tide provides support for linting with `tslint` but not for linting with `eslint`. It turns out that linting TypeScript code with `eslint` is already a thing. (I did not realize that. I use `eslint` but only for JS code, and I exclusively use `tslint` for TS code.)

I think adding support for `eslint` *now* would be an enhancement, but it appears that `tslint` will eventually be deprecated. (See https://github.com/Microsoft/TypeScript/issues/29288.) So eventually the lack of support for `eslint` will be a bug.

I see 3 areas of concern:

1. When `eslint` is to be used for linting, sequence the `eslint`-based checker to run after `typescript-tide` and `tsx-tide` like we do with `typescript-tslint`.

2. Provide an `eslint` equivalent to `tide-add-tslint-disable-next-line`. (I actually already have code for this.)

3. Update the documentation to explain how to use `eslint` with tide. (The equivalent of what we already have for `tslint`.)

ananthakumaran:
I am ok with 2 and 3.

about 1 could you explain how would you figure out the user is planning to use `eslint`? via custom variable config?

Personally, I don't like the fact that tide has to be responsible for setting the correct next checker. We had the discussion about the same issue before and finally agreed to set next checker.

> Yeah, it seems tide having better performance, but I've been having problem that it keeps creating a flycheck_myfilename.tsx file, and my react server keeps breaking :( , I am not sure if it's flycheck's problem or tide's. That's why I switched to lsp, it's been working better than tide tbh, just a bit slow :(

but apparently it's confusing others as well. But, I don't have any good alternative solution either for the issue.

josteink:
> That's why I switched to lsp, it's been working better than tide tbh

Disregarding everything else about this issue, I can really sympathize with this, and it's been a thought I've been having for a good while.

Once Emacs and LSP works properly, why should we bother maintaining a separate LSP client, just specifically for TypeScript?

When I checked the state of affairs with Emacs and LSP last year, it was quite a joke. Recently though I've taken a look at it again (from a Rust-perspective) and it has improved immensely, to the point of me suspecting that whatever I was missing was actually due to incomplete support in the Rust Language Server, not Emacs.

And once things get really good and just work out of the box, I would really expect most Emacs-users to prefer a "standard" LSP experience over some non-standard thingie like `tide`, also for TypeScript.

[...]

PR #359, opened 2020-01-14
Rework the linting support to support disabling eslint errors.

lddubeau:
We also gain support for adding the @ts-ignore flag that the tsc compiler understands.

This change adds logic for identifying the general type of checker from which flycheck errors are reported, and so be able to add the right flag when the user requests that an error be turned off.

This tackles point 2 of [this post](https://github.com/ananthakumaran/tide/issues/308#issue-423772384).

I've been running this for a while in projects that use eslint for all linting work, and projects that still use tslint, and it's been working for me.

### DIFF
--- a/tide-tests.el
+++ b/tide-tests.el
@@ -209,20 +209,28 @@ of `BODY'."
          (tide-kill-server)
          (kill-buffer ,buffer-sym)))))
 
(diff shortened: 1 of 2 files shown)


josteink:
Sorry about the late response.

If it's OK, I'll hold on investigating this PR, until the questions I've raised in #308 are put down or resolved :smile:

PR #363, opened 2020-02-17
Add LICENSE file.

phst:
This makes it easier for humans and tools to determine the license of the code.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,674 @@
+                    GNU GENERAL PUBLIC LICENSE
+                       Version 3, 29 June 2007
+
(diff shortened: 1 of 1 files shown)


josteink:
Unless I'm mistaken, the package meta-data suggest plain GPL. Is that just "upgradable" to GPLv3 like that? If so, great.

@ananthakumaran: As the actual project-owner, do you have any specific opinion in these matters? I think it would be best if you handled this PR.

ananthakumaran:
> either version 3 of the License, or (at your option) any later version.

The license header in the source file specifies GPL v3 as well.

Issue #292, opened 2018-12-20
Deciding when to drop support for old Emacs versions.


lddubeau:
@josteink and I began talking about it on #279 but I think this deserves its own discussion. Specifically, I'd like @ananthakumaran to chime in.

When a contributor produces a PR, and the PR fails on an old version of Emacs, then in order to get the PR to run on the old version, the contributor has to research the failure and pour in more time into amending the PR to run on the old version of Emacs.

If we can find a limit to how old a version of Emac `tide` supports, then we can limit the time cost of supporting old versions. *And it would be nice to have a simple rule that collaborators can apply, rather than have to discuss it over and over.*

The initial discussion in #279 centered on checking what version of Emacs various distributions ship. Ubuntu 16.04 LTS ships 24.5. Debian 9 (the latest stable Debian release) ships 24.5 too, but I had forgotten that Debian 8 is a LTS, which ships 24.4. However, Ubuntu 14.04 is also an LTS that has not reached end of life yet (this happens in April 2019) and it ships 24.3, as far as I can tell.

We could have for rule that 

> `tide` is guaranteed to run on an Emacs version as old as the oldest version currently shipped by default by the latest two active LTSes of `[some distribution(s)]`. (An "active LTS" being one that has not yet reached end-of-life.) 

We have to decide on `some distribution(s)`. As a practical matter we cannot check all distributions out there so we have to limit the list. It could be just Ubuntu though I'm biased and would like Debian included because I use Debian everywhere. Though I think practically I'd be able to get by if we track only Ubuntu, through installing Emacs from the `...-backports` distribution.

"the latest two active LTSes" has for effect to limit how far back we look. Looking at Ubuntu today, it would limit the check to 18.04 and 16.04, and exclude 14.04.

Or maybe there's a better rule than the one above. Or maybe no rule and we have to raise an issue every time we consider dropping an old version, but I'd rather have some basic rule that we can also put in the README so users know what to expect.

----

Since the issue came up in #279, I'll mention that my desire to fix the backward compatibility issue there would be to drop testing on 24.3. I'm expecting that modifying the code for 24.5 will also make it work for 24.4, but 24.3 requires more work, and requires uglier modifications than for 24.5.

ananthakumaran:
[...]

PR #369, opened 2020-03-17
add partial support for xref

thanhvg:
This commit adds support for `xref-find-references` command. 

image

`xref` is only available for emacs 25 and later.

### DIFF
--- a/tide.el
+++ b/tide.el
@@ -1219,7 +1219,7 @@ Noise can be anything like braces, reserved keywords, etc."
            (tide-eldoc-maybe-show (tide-doc-text (plist-get response :body))))))))
   nil)
 
-(defun tide-eledoc-display-message-p()
+(defun tide-eldoc-display-message-p()
   (if (fboundp #'eldoc-display-message-no-interference-p)
       (eldoc-display-message-no-interference-p)
     (eldoc-display-message-p)))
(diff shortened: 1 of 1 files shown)


tide.el
```diff
@@ -39,6 +39,7 @@
 (require 'thingatpt)
 (require 'tide-lv)
 (require 'tabulated-list)
+(require 'xref)
```
ananthakumaran: I think we need to decide whether to drop support for emacs 24. I am +1 on dropping support, given that emacs 27 in is prerelease stage

lddubeau: We've established rules as to what versions we support in #292, which I've extracted and written down [here](https://github.com/ananthakumaran/tide/wiki/Supported-Emacs-Versions). By those rules, we should still support Emacs 24.

I'd expect the rule we came up with to allow dropping Emacs 24 by the end of June. (Debian 8 will reach end of life by June 30th, and I think an Ubuntu LTS will be released in April, which will mean Ubuntu 16.04 will no longer be one of the two latest active LTS of Ubuntu.)

One of the goals of agreeing on rules was to avoid having to debate over and over how to determine how far back to provide support for Emacs.

ananthakumaran: agree on the rules. That means we wait till June end or add support in such a way that it doesn't break emacs 24.

lddubeau: I'm fine with either solution. @thanhvg let us know what solution you pick.

thanhvg: I think we should wait till summer.

PR #375, opened 2020-04-27
Use native JSON parsing when available

### LINKED ISSUE #374: Feature request: Use jansson-backed json de/serialization

## Checklist

- [X] I have searched both open and closed issues and cannot find a duplicate.
- [X] I can [reproduce the problem](https://github.com/ananthakumaran/tide/wiki/Explanations-about-the-Issue-Reporting-Checklist#use-the-latest-version-of-the-relevant-packages) with the latest version of the relevant packages.
- [X] The problem still occurs after I issued [`M-x tide-restart-server`](https://github.com/ananthakumaran/tide/wiki/Explanations-about-the-Issue-Reporting-Checklist#try-m-x-tide-restart-server) in the buffer where I had the problem.
- [X] I verified that the version and the configuration file path reported by [`M-x tide-verify-setup`](https://github.com/ananthakumaran/tide/wiki/Explanations-about-the-Issue-Reporting-Checklist#check-the-information-reported-by-m-x-tide-verify-setup) are correct.
- [ ] If [tide *is reporting*](https://github.com/ananthakumaran/tide/wiki/Explanations-about-the-Issue-Reporting-Checklist#the-behavior-of-cli-tools-agrees-with-your-expectations) an error or warning I think *should not be reported*, I can run `tsc` (and `tslint`, if applicable) without the error or warning I'm seeing in tide.
- [X] If [tide is *not reporting*](https://github.com/ananthakumaran/tide/wiki/Explanations-about-the-Issue-Reporting-Checklist#the-behavior-of-cli-tools-agrees-with-your-expectations) an error or warning I think *should be reported*, `tsc` (or `tslint`, if applicable) reports the error or warning I was expecting to see.
- [X] I am [positive](https://github.com/ananthakumaran/tide/wiki/Explanations-about-the-Issue-Reporting-Checklist#how-to-determine-which-project-the-problem-belongs-to) the problem does not belong to `typescript-mode` or `tsserver`.

## Relevant Version Numbers

- Tide: all
- TypeScript: all
- Emacs: >=27 

## Steps to Reproduce the Bug

- search tide elisp code for json-serialize, json-parse-string, json-parse-buffer, etc.

## Expected Behavior

- find code that uses these defuns

## Actual Behavior

- no code uses these defuns

Starting in Emacs 27, Emacs has defuns that use a json library written in C, yielding performance gains of up to 15x on json operations -- see https://github.com/emacs-lsp/lsp-mode/issues/210
lsp-mode has implemented feature detection, with fallback to json.el, for example: https://github.com/emacs-lsp/lsp-mode/blob/d89f403fbf629f4c6b95d63115a4959547cea531/lsp-mode.el#L5925-L5937
That seems like a sensible approach that tide could use as well!

jakejx:
I took a stab at making tide use the native json parser when it is available. There is a new variable `tide-native-json-parsing` that controls which JSON parser is used. I am quite new to writing elisp so I am happy to have feedback on how to improve the code.

Closes #374

### DIFF
--- a/tide.el
+++ b/tide.el
@@ -1290,7 +1290,7 @@ Noise can be anything like braces, reserved keywords, etc."
    `(:file ,(tide-buffer-file-name)
      :startLine ,(tide-line-number-at-pos) :startOffset ,(tide-current-offset)
      :endLine ,(tide-line-number-at-pos) :endOffset ,(+ 1 (tide-current-offset))
-     :errorCodes ,(tide-get-flycheck-errors-ids-at-point))))
+     :errorCodes ,(vconcat (tide-get-flycheck-errors-ids-at-point)))))
 
 (defun tide-command:getCombinedCodeFix (fixId)
   (tide-send-command-sync "getCombinedCodeFix"
--- a/tide.el
+++ b/tide.el
@@ -242,7 +242,8 @@ this variable to non-nil value for Javascript buffers using `setq-local' macro."
 (defcustom tide-native-json-parsing (and (>= emacs-major-version 27)
                                          (functionp 'json-serialize)
                                          (functionp 'json-parse-buffer)
-                                         (functionp 'json-parse-string))
+                                         (functionp 'json-parse-string)
+                                         nil)
   "Use native JSON parsing (only emacs >= 27)."
   :type 'boolean
   :group 'tide)
--- a/tide.el
+++ b/tide.el
@@ -346,7 +346,7 @@ buffers, as they don't set `buffer-file-name' correctly."
   (condition-case nil
       (if tide-native-json-parsing
           (with-temp-buffer
-            (insert-file-contents)
+            (insert-file-contents filename)
             (goto-char (point-min))
             (json-parse-buffer :object-type 'plist :null-object json-null :false-object json-false))
         (let ((json-object-type 'plist))
@@ -360,6 +360,19 @@ buffers, as they don't set `buffer-file-name' correctly."
         (let ((json-object-type 'plist))
           (json-read-from-string string)))
     (error '())))
+
+(defun tide-json-read-object ()
+  (if tide-native-json-parsing
+      (json-parse-buffer :object-type 'plist :null-object json-null :false-object json-false :array-type 'list)
+    (let ((json-object-type 'plist)
+          (json-array-type 'list))
+      (json-read-object))))
+
+(defun tide-json-encode (obj)
+  "Encode OBJ into a JSON string. JSON arrays must be represented with vectors."
+  (if tide-native-json-parsing
+      (json-serialize obj :null-object json-null :false-object json-false)
+    (json-encode obj)))
 
 (defun tide-plist-get (list &rest args)
   (cl-reduce
@@ -606,9 +619,7 @@ Offset is one based."
   (let* ((request-id (tide-next-request-id))
          (command `(:command ,name :seq ,request-id :arguments ,args))
          (json-encoding-pretty-print nil)
-         (encoded-command (if tide-native-json-parsing
-                              (json-serialize command :null-object json-null :false-object json-false)
-                            (json-encode command)))
+         (encoded-command (tide-json-encode command))
          (payload (concat encoded-command "\n")))
     (process-send-string (tide-current-server) payload)
     (when callback
@@ -831,9 +842,7 @@ Currently, two kinds of cleanups are done:
 
 (defun tide-decode-response (process)
   (with-current-buffer (process-buffer process)
-    (let ((length (tide-decode-response-legth))
-          (json-object-type 'plist)
-          (json-array-type 'list))
+    (let ((length (tide-decode-response-legth)))
       (when (and length (tide-enough-response-p length))
         (search-forward "{")
         (backward-char 1)
@@ -845,7 +854,7 @@ Currently, two kinds of cleanups are done:
                                    `(:success :json-false :type "response"
                                      :message ,tide-max-response-length-error-message
                                      :request_seq ,seq)))
-                          (json-read-object))))
+                          (tide-json-read-object))))
           (delete-region (point-min) (point))
           (when response
             (tide-dispatch response)))
--- a/tide.el
+++ b/tide.el
@@ -239,6 +239,14 @@ this variable to non-nil value for Javascript buffers using `setq-local' macro."
   :group 'tide
   :safe #'booleanp)
 
+(defcustom tide-native-json-parsing (and (<= emacs-major-version 27)
+                                         (functionp 'json-serialize)
+                                         (functionp 'json-parse-buffer)
+                                         (functionp 'json-parse-string))
+  "Use native JSON parsing (only emacs >= 27)."
+  :type 'boolean
+  :group 'tide)
+
 (defconst tide--minimal-emacs
   "24.4"
   "This is the oldest version of Emacs that tide supports.")
@@ -336,14 +344,21 @@ buffers, as they don't set `buffer-file-name' correctly."
 
 (defun tide-safe-json-read-file (filename)
   (condition-case nil
-      (let ((json-object-type 'plist))
-        (json-read-file filename))
+      (if tide-native-json-parsing
+          (with-temp-buffer
+            (insert-file-contents)
+            (goto-char (point-min))
+            (json-parse-buffer :object-type 'plist :null-object json-null :false-object json-false))
+        (let ((json-object-type 'plist))
+          (json-read-file filename)))
     (error '())))
 
 (defun tide-safe-json-read-string (string)
   (condition-case nil
-      (let ((json-object-type 'plist))
-        (json-read-from-string string))
+      (if tide-native-json-parsing
+          (json-parse-string string :object-type 'plist :null-object json-null :false-object json-false)
+        (let ((json-object-type 'plist))
+          (json-read-from-string string)))
     (error '())))
 
 (defun tide-plist-get (list &rest args)
@@ -591,7 +606,9 @@ Offset is one based."
   (let* ((request-id (tide-next-request-id))
          (command `(:command ,name :seq ,request-id :arguments ,args))
          (json-encoding-pretty-print nil)
-         (encoded-command (json-encode command))
+         (encoded-command (if tide-native-json-parsing
+                              (json-serialize command :null-object json-null :false-object json-false)
+                            (json-encode command)))
          (payload (concat encoded-command "\n")))
     (process-send-string (tide-current-server) payload)
     (when callback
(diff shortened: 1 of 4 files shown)


tide.el
```diff
@@ -239,6 +239,14 @@ this variable to non-nil value for Javascript buffers using `setq-local' macro."
...
 
+(defcustom tide-native-json-parsing (and (<= emacs-major-version 27)
+                                         (functionp 'json-serialize)
+                                         (functionp 'json-parse-buffer)
+                                         (functionp 'json-parse-string))
+  "Use native JSON parsing (only emacs >= 27)."
+  :type 'boolean
+  :group 'tide)
+
 (defconst tide--minimal-emacs
```
josteink, line 250: I like this. Make it configurable, at least as long as it's not properly tested.

tide.el
```diff
@@ -336,14 +344,21 @@ buffers, as they don't set `buffer-file-name' correctly."
...
-      (let ((json-object-type 'plist))
-        (json-read-file filename))
+      (if tide-native-json-parsing
+          (with-temp-buffer
+            (insert-file-contents)
+            (goto-char (point-min))
+            (json-parse-buffer :object-type 'plist :null-object json-null :false-object json-false))
+        (let ((json-object-type 'plist))
+          (json-read-file filename)))
     (error '())))
```
josteink, line 354: Looks reasonable, but would be even better with a test-case attached, testing for both modes.

tide.el
```diff
@@ -336,14 +344,21 @@ buffers, as they don't set `buffer-file-name' correctly."
...
 
 (defun tide-safe-json-read-string (string)
   (condition-case nil
-      (let ((json-object-type 'plist))
-        (json-read-from-string string))
+      (if tide-native-json-parsing
+          (json-parse-string string :object-type 'plist :null-object json-null :false-object json-false)
+        (let ((json-object-type 'plist))
+          (json-read-from-string string)))
```
josteink, line 361: Same for this one: tests would be nice.

tide.el
```diff
@@ -591,7 +606,9 @@ Offset is one based."
   (let* ((request-id (tide-next-request-id))
          (command `(:command ,name :seq ,request-id :arguments ,args))
          (json-encoding-pretty-print nil)
-         (encoded-command (json-encode command))
+         (encoded-command (if tide-native-json-parsing
+                              (json-serialize command :null-object json-null :false-object json-false)
+                            (json-encode command)))
          (payload (concat encoded-command "\n")))
```
josteink, line 612: It would be nice if this was also isolated in its own `defun`, and then it would be easy to add tests covering it.

tide.el
```diff
@@ -1561,7 +1578,7 @@ This function is used for the basic completions sorting."
   (let* ((source (plist-get (get-text-property 0 'completion name) :source))
          (entry-names (if source
                           `(:entryNames [(:name ,name :source ,source)])
-                        `(:entryNames (,name))))
+                        `(:entryNames [,name])))
          (arguments (-concat (get-text-property 0 'file-location name)
```
josteink, line 1582: This change stands out. Was it intentional? Could you elaborate on why it was required?

tide.el
```diff
@@ -2439,12 +2456,11 @@ current buffer."
    t))
 
 ;;; Identifier highlighting
-
 (defun tide-command:documentHighlights (cb)
   (tide-send-command
    "documentHighlights"
    `(:file ,(tide-buffer-file-name) :line ,(tide-line-number-at-pos) :offset ,(tide-current-offset)
-     :filesToSearch (,(tide-buffer-file-name)))
+     :filesToSearch [,(tide-buffer-file-name)])
```
josteink, line 2463: Same for this change: It stands out. Was it intentional? Why it was required?

josteink:

I've taken a look and it mostly looks good. No major objections to the implementation as it is, but I've left a few comments none the less.

If we got those resolved, I'm pretty sure we can get this merged without too many issues :+1:

tide.el
jakejx: Yeap it was intentional. The two changes are due to `json-serialize` (native) only serialises vectors to JSON arrays (AFAIK), while`json-encoded` (elisp) seraliases both vectors and lists, so I decided to use a vector for compatibility.

jakejx:
Thanks for the swift review! I'll look into adding the tests 😄

tide.el
ananthakumaran: should we think about adding `tide-json-encode, tide-json-decode` instead of inlining the native support check everywhere? It would be easier to debug as well if there are any differences (just encode/decode via both method and compare the values)

ananthakumaran:
Most of the performance benefit is going to come from switching the `json-read-object` inside `tide-decode-response`, which I don't see in this PR. This is the only place where we decode large json responses (for autocomplete mostly) from tsserver.

tide.el
josteink: I was considering that too, but decided that we could live with 2 repetitions of the same boilerplate, without going crazy with intermediate functions.

If however, as you say, there are even more usage (`json-read-object` inside `tide-decode-response`), abstracting it into its own function definitely makes sense.

josteink: I was considering that too, but decided that we could live with 2 repetitions of the same boilerplate, without going crazy with intermediate functions.

If however, as you say, there are even more usage (`json-read-object` inside `tide-decode-response`), abstracting it into its own function definitely makes sense.

And that will also help testability.

jakejx:
> Most of the performance benefit is going to come from switching the `json-read-object` inside `tide-decode-response`, which I don't see in this PR. This is the only place where we decode large json responses (for autocomplete mostly) from tsserver.

I must have missed it, will work on using the native decoder for that.

jakejx:
I have updated the code to use the native JSON parser for reading objects. Currently there are 4 functions that are entry points (and thus where the check for the native flag occurs): `tide-safe-json-read-file`, `tide-safe-json-read-string`, `tide-json-read-object` and `tide-json-encode`.

I have added a single test file (just a random tsconfig) that covers the common JSON structures. The tests basically reads this file and ensures that both JSON parsers produce the same output.

One thing to note is that when reading JSON, JSON arrays are represented as lists. However, when serialising to JSON, the arrays must be vectors (as the native parser only accepts vectors). While it is possible to read JSON arrays as vectors, I think this would require other changes to other parts of the code, as `car` and `cdr` are used for list manipulation.

tide.el
```diff
@@ -239,6 +239,14 @@ this variable to non-nil value for Javascript buffers using `setq-local' macro."
   :group 'tide
   :safe #'booleanp)
 
+(defcustom tide-native-json-parsing (and (<= emacs-major-version 27)
```
bimbiltu, line 242: Shouldnt this line be using `>=` instead of `<=`?

jakejx: yes that was a mistake, updated.

josteink:
Thanks for the updates. The build seems to be failing for 2 reasons:

* The build for `emacs-git-master-snapshot` seems not to be running an old snapshot, and has had test-failures before. Not a big deal (for this PR). I'd like to look into that independently though.
* You are running your new tests on older Emacs-versions where the new JSON engine isn't present.

Could you preface your JSON-tests with `(when (>= emacs-major-version 27) ... )` ? That should fix the build and make us ready to merge, unless there are other objections.

josteink:
If you rebase this PR on master (or just merge latest changes from master now), the CI should be updated to include an up-to-date git master Emacs version.

This version should hopefully be able to run and verify your tests, and if that build is green, I think we can agree the tests are properly good.

tide.el
ananthakumaran: agree, I think we should keep this disabled by default for some time, and then once well tested, we can enable based on emacs version and function availability.

jakejx:
It looks like the snapshot is not sufficiently up to date. This is from travis: `/home/travis/build/ananthakumaran/tide/tide.el:2806:1:Warning: the following
    functions are not known to be defined: json-parse-buffer,
    json-parse-string, json-serialize`

josteink:
> It looks like the native functions are not being loaded from some reason in the CI. This is from travis: `/home/travis/build/ananthakumaran/tide/tide.el:2806:1:Warning: the following functions are not known to be defined: json-parse-buffer, json-parse-string, json-serialize`

I agree. From the build-outpout:

````
$ emacs --version

GNU Emacs 27.0.50

Copyright (C) 2019 Free Software Foundation, Inc.
````

That does not seem to be an up to date build, and having that build failing does not block the merging of this PR.

File: tide.el [hunk 8]
```diff
@@ -239,6 +239,15 @@ this variable to non-nil value for Javascript buffers using `setq-local' macro."
   :group 'tide
   :safe #'booleanp)
 
+(defcustom tide-native-json-parsing (and (>= emacs-major-version 27)
+                                         (functionp 'json-serialize)
+                                         (functionp 'json-parse-buffer)
+                                         (functionp 'json-parse-string)
+                                         nil)
```
josteink, line 246: What's the point of this whole setup, if we terminate with `nil` ? Is it for making it easier to correctly switch default in the future?

josteink:
Trying to run this locally, based on Emacs `git master` I couldn't get any of the new tests to pass.

Looking into it, after looking quite a few places, I finally found I had a missing prerequisite: `libjansson` was not installed on my system, so obviously Emacs could not use it.

If it can be "hard" for someone even involved in `tide` to correctly implement/enable a crucial performance-related component like this, would it make sense to have this more clearly highlighted in the `README`?

Basically just a section about performance, which informs that `tide` performs better when running on Emacs-builds which has `libjansson`-use enabled?

Any opinions?

josteink:
You know, I've tested it manually, and things `seems` to run faster. I'll be happy to merge this code as is.

If someone wants to improve the readme, w.r.t. to the optimal setup for performance and features, that can be done in a separate PR.

Consider this merged!

tide.el
jakejx: yeap the main reason is to keep the native parsing disabled by default (even if the user's system supports it). Eventually we can remove the `nil` and the value will be determined by the user's system.

jakejx:
> If it can be "hard" for someone even involved in `tide` to correctly implement/enable a crucial performance-related component like this, would it make sense to have this more clearly highlighted in the `README`?
> 
> Basically just a section about performance, which informs that `tide` performs better when running on Emacs-builds which has `libjansson`-use enabled?
> 
> Any opinions?

I agree with this. I will look into opening a PR to update the `README`.

josteink:
On a side-note: Installing `libjansson` and rebuilding Emacs, now makes `lsp-mode` fly compared to before as well. My entire Emacs feels snappier.

Thanks for helping me make my Emacs-setup even better! :smiley:

tide.el
sandinmyjoints: Maybe time to remove this?

ananthakumaran: removed it in master.

PR #379, opened 2020-05-17
add custom variable `tide-completion-source` to show completion source in dropdown

bolasblack:


josteink:
Name of customisation in documentation is in insistent with defcustom in code. Decide on one name please 🙂

ananthakumaran:
> Name of customisation in documentation is in insistent with defcustom in code. Decide on one name please 🙂

You could run `make readme` to autogenerate the readme

bolasblack:
> > 
> 
> You could run `make readme` to autogenerate the readme

Cool, thank you

ananthakumaran:
@bolasblack I have made some changes to the branch. Could you test it once?

README.md
```diff
@@ -325,26 +328,47 @@ Maximum allowed response length from tsserver. Any response greater than this wo
...
 completions whose `:kind` property is "warning" for Javascript code. You can fix this behavior by setting
 this variable to non-nil value for Javascript buffers using `setq-local` macro.
 
+##### tide-native-json-parsing `(and
+ (>= emacs-major-version 27)
+ (functionp 'json-serialize)
+ (functionp 'json-parse-buffer)
+ (functionp 'json-parse-string)
+ nil)`
+
```
josteink, line 368: Documentation generated here is clearly wrong.

josteink:
I've found 1 error in the generated documentation, which may or may not be related to the changes of this PR.

This error could have easily been overlooked though, because there was lot of "noise" in the updated documentation, coming from earlier changes which had yet to have its documentation generated.

I think it would make sense to update documentation in Git master first, and rebase this PR on latest changes there, so it would be easier to review these changes (and the impact on documentation it has) more easily.

Or does anyone have a problem with that?

README.md
ananthakumaran: The script just inlines the code and we can remove the nil at the end once we are confident about the native JSON. Anyhow, this shouldn't affect this PR from getting merged.

bolasblack:
> @bolasblack I have made some changes to the branch. Could you test it once?

@ananthakumaran Thanks very much, it works good 👍

bolasblack:
Something wired, the source under the `node_modules` directory shows the absolute path:

图片

ananthakumaran:
this is strange, not sure if it's a bug or expected result from tsserver. I can't think of any way other than trimming/taking basename of the file.

bolasblack:
I found `completionEntryDetails` command support request by array of `interface CompletionEntryIdentifier { name: string, source?: string }` (https://github.com/microsoft/TypeScript/blob/3e0d58f4d2353fd4c4c274d1aeafb8b6ae022605/lib/protocol.d.ts#L1510-L1519)

So maybe we can request `completionEntryDetails` once for all `CompletionEntry` before show dropdown, if the custom variable enabled?

josteink:
> So maybe we can request completionEntryDetails once for all CompletionEntry before show dropdown, if the custom variable enabled?

Maybe I'm a cynic, and too used to using such APIs over networks, but to me that sounds like it's going to be very, very slow.

Feel free to write a POC and test if works for you, though.

ananthakumaran:
I think we are discussing two different things
1) the one here is the source field, which is available in the normal completion struct and completionDetail struct.
2) The second thing about switching to bulk request to fetch completionEntryDetails, I haven't looked at the details, if it's possible we could do it. But note that, by default completion details is disabled. tsserver is known to send a lot of results back > 1Mb, so asking for details for all requests is going to send an even big response back which could cause issues. This is why we disable it by default.

As for the scope of this PR, let's just fix 1. If you want to solve 2, you can open a separate PR.

bolasblack:
Hi @ananthakumaran, I wrote a new function `tide-normalize-source` to transform `../../../../../node_modules/@types/babel/react/index` to `@babel/core`, so the source can be more readable :D

ananthakumaran:
I am going to merge this, likely time-normalize-source is not handling all the cases, but it's ok given that we have the customization disabled by default and we can fix new cases when we find it.

bolasblack:
@ananthakumaran Thank you 😁

PR #383, opened 2020-06-10
Fix edebug specification for ‘with-spy’.

phst:
The first two arguments are symbols, not arbitrary forms.

### DIFF
--- a/tide-tests.el
+++ b/tide-tests.el
@@ -174,7 +174,7 @@ a test failure."
 
 (defmacro with-spy (name function-symbol &rest body)
   "Execute `BODY' with `NAME' bound to a spy that spies on `FUNCTION-SYMBOL'."
-  (declare (debug t)
+  (declare (debug (symbolp symbolp body))
            (indent 2))
   ;; Avoid polluting the scope of the code using this macro.
   (let ((orig (make-symbol "orig"))


josteink:

Consider me dumb if you like, but I really don't understand what this change does :smile: 

Could you please provide some background as to why these changes are needed, and in what way they change tide behaviour?

If I can understand what this PR fixes, I'll be happy to merge.

phst:
Sure. You can reproduce the issue e.g. by doing the following:

1. Make sure the `spy` struct is defined (navigate to its definition and hit `C-M-x`)
2. Edebug `with-spy` (`C-u C-M-x`)
3. Define the rest of the macros used in the tests
4. Edebug some of the tests that uses `with-spy` (e.g. `tide-setup/immediate-server-start`)
5. Run the test using `M-x ert`

Once the execution hits the `with-spy` macro, you'll get an error

> Wrong type argument: symbolp, (edebug-after 0 2 configure-spy)

This is because of the broken `debug` specification of `with-spy`. The current `t` is equivalent to `(form form body)`, causing Edebug to replace the first two arguments with instrumenting forms. But that's incorrect: they are symbols and used as such in the macro body, e.g. by let-binding `name`. So using `form` here is wrong, the edebug specification needs to be `symbolp` or similar.

josteink:
Thanks for the clarification. To me this almost sounds like something which must "always" have been broken.

Is that so? Or was this error introduced by a change in the way `edebug` is implemented in Emacs?

Or more down to business: Are there any chances of this change causing unexpected regressions?

phst:
> To me this almost sounds like something which must "always" have been broken. Is that so?

I believe so. I haven't checked the history of Edebug, so it's possible that it was more lenient in the past. I think the more probable explanation is that this bug is only hit in rare edge cases: only if the `with-spy` macro *and* one of its callers is edebugged. I probably just happen to be the first one to hit this edge case.

> Or more down to business: Are there any chances of this change causing unexpected regressions?

No. First, this only changes the test code, so the user-visible behavior cannot change. Second, the `debug` declaration only affects how Edebug instruments the code, it doesn't affect the semantics of the code. Typically the test is run without Edebug instrumentation, in which case there's no effect at all.

josteink:
Since there’s seemingly no other objection s to this PR, I’ll merge it as is. 

Thanks for the patch, and thanks for hanging in there 🙂

PR #384, opened 2020-07-04
Leave company-backends for the user to setup

alanpearce:
Loading `tide.el` has the side-effect of adding the completion backend to `company-backends` with no configuration option. The user may however not wish to use `company-tide`, or not have it as the first backend.

Other `company` packages do not mangle `company-backends`, instead documenting the code to add their backend, to allow the user to add the backends in their preferred order (examples: [1](https://github.com/krzysztof-magosa/company-ansible#how-to-configure-company), [2](https://github.com/randomphrase/company-c-headers/tree/5e676ab0c2f287c868b1e3931afd4c78895910cd#setup), [3](https://github.com/tigersoldier/company-lsp/tree/f921ffa0cdc542c21dc3dd85f2c93df4288e83bd#usage)). I therefore propose this change to follow this convention.

### DIFF
--- a/tide.el
+++ b/tide.el
@@ -125,6 +125,11 @@ errors and tide-project-errors buffer."
   :type 'boolean
   :group 'tide)
 
+(defcustom tide-completion-setup-company-backend t
+  "Add `company-tide' to `company-backends'."
+  :type 'boolean
+  :group 'tide)
+
 (defcustom tide-completion-ignore-case nil
   "CASE will be ignored in completion if set to non-nil."
   :type 'boolean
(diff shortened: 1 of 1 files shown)


josteink:
This will be a breaking change/upgrade for most tide-users, which (for me personally) makes it a no-go. 

While I can't speak for the other contributors, but if you could come up with some suggestions for how we can eliminate this causing breaking changes, that would definitely increase the chance of this getting merged.

ananthakumaran:
https://github.com/company-mode/company-mode/issues/923#issuecomment-544263928 from this comment, it seems like updating `company-backends` from third party package is not the correct approach. But unfortunately, as @josteink pointed out, this code has been there from the beginning and going to break nearly all the existing users' setup which I prefer not to do. We can't merge this PR unless we come up with a solution that doesn't break the existing setup.

josteink (approved):
This looks good to me. What do you think, @ananthakumaran ?

ananthakumaran:
LGTM

josteink:
Merged!

alanpearce:
Thanks for re-reviewing and merging so quickly! Sorry I didn't get around to addressing your feedback for... a while.

josteink:
Np. A patch is a patch, and GitHub makes sure we can pick it up anytime, even 9 months later 😀

PR #387, opened 2020-07-20
Improve file ignoring in tide-project-errors view

kijowski:
This PR introduce new configuration option `tide-project-files-ignore-patterns` which allows for better control of what files are displayed in `tide-project-errors` buffer.

It also makes sure that the errors from filtered `project-files` are not displayed in the buffer - I had a problem where I was getting entries for files from my node_modules even thought they should be filtered out. I think that this should be fixed by https://github.com/ananthakumaran/tide/pull/315 , so maybe this was an unrelated error.

### DIFF
--- a/tide.el
+++ b/tide.el
@@ -252,6 +252,11 @@ this variable to non-nil value for Javascript buffers using `setq-local' macro."
   :type 'boolean
   :group 'tide)
 
+(defcustom tide-project-files-ignore-patterns '("node_modules" "tsconfig.*.json$")
+  "List of regular expressions to ignore while displaying errors with `tide-project-errors'."
+  :type '(repeat string)
+  :group 'tide)
+
 (defconst tide--minimal-emacs
   "24.4"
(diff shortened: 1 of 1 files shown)


ananthakumaran:
hm, strange, when we call `geterrForProject`, we usually don't get errors for node_modules. The logic is not there to filter out node_modules, but rather to print the footer `"\n%d syntax error(s), %d semantic error(s), %d suggestion error(s)\n"`, for that we need to know whether we have received errors for all the files from tsserver. if we wait for node_modules, we will wait forever because tsserver never send those.

if you can share a minimal project that can reproduce the issue, that would help with debugging the issue.

kijowski:
Thanks for looking into it - I've experimented a bit and it seems that this might be related to tsserver version. At  https://github.com/kijowski/tide-repro you can find the simplest repo that is displaying this behavior. I was able to replicate it on Typescript version newer then 3.8.0 - on 3.7 the issues from node modules are not displayed.

This is how it looks:
image

Issue #388, opened 2020-07-21
Support for Yarn PnP


gamb:
Just creating this as a stub. It'd be really neat if tide could support [Yarn PnP](https://yarnpkg.com/advanced/pnpapi), where files are zipped inside `.yarn/cache`.

Quick and dirty solution that works (but relies on you already having the zipped buffer open):
```diff
(defun tide-get-file-buffer (file &optional new-file)
  "Returns a buffer associated with a file. This will return the
current buffer if it matches `file'. This way we can support
temporary and indirect buffers."
  (cond
   ((equal file (tide-buffer-file-name)) (current-buffer))
   ((file-exists-p file) (find-file-noselect file))
   (new-file (let ((buffer (create-file-buffer file)))
               (with-current-buffer buffer
                 (set-visited-file-name file)
                 (basic-save-buffer)
                 (display-buffer buffer t))
               buffer))
+   ((string-match "\\$\\$virtual.*cache/" file)
+   (find-file-noselect (replace-regexp-in-string "\\$\\$virtual.*cache/" "cache/" (replace-regexp-in-string "\\.zip/" ".zip:" file))))
   (t (error "Invalid file %S" file))))
```

It is possible to `yarn unplug` dependencies to unzip them, but it seems to me that we should be able to leverage Emacs ability open buffers inside zips.

gamb:
Example response with PnP:
```
1 -> (tide-response-success-p (:seq 0 :type "response" :command "definition" :request_seq "589" :success t :body ((:file "/home/gamble/path/to/project/.yarn/$$virtual/recoil-virtual-ad5204a645/0/cache/recoil-npm-0.0.10-9811d22459-2.zip/node_modules/recoil/dist/index.d.ts" :start (:line 117 :offset 17) :end (:line 117 :offset 31) :contextStart (:line 117 :offset 1) :contextEnd (:line 117 :offset 89)))))
```
The response path is transformed from:
```
.yarn/$$virtual/recoil-virtual-ad5204a645/0/cache/recoil-npm-0.0.10-9811d22459-2.zip/node_modules/recoil/dist/index.d.ts
```
To an Emacs-friendly path (on disk):
```
.yarn/cache/recoil-npm-0.0.10-9811d22459-2.zip:node_modules/recoil/dist/index.d.ts
```

ramblehead:
[...]

PR #392, opened 2020-07-28
Adding Yarn 2 PnP support

ramblehead:
This patch allows tide to jump to files stored in zip packages (in Yarn 2 cache)
using Emacs arc-mode.

See #388


ramblehead:
Hard-reset in my fork :/ - will create another PR.

Issue #390, opened 2020-07-24
eldoc-display-message-p() renamed to eldoc-display-message-no-interference-p()


ramblehead:
tide-eldoc-maybe-show() fails in latest (development) emacs build due to Subj.
Renaming fixes the issue.

josteink:
We need some decent cross version compatibility, since not everyone runs latest version of Emacs.

I guess we should abstract this into a function like `tide-eledoc-display-message-p` where we dispatch to the appropriate function, based on what is available.

PRs welcome :)

ramblehead:
@josteink thank you for the reply. Happy to PR. Would aliasing be acceptable:

```elisp
(when (< 28 emacs-major-version)
  (defalias 'eldoc-display-message-no-interference-p 'eldoc-display-message-p))

;;; Copied from eldoc.el
(defun tide-eldoc-maybe-show (text)
  (with-demoted-errors "eldoc error: %s"
    (and (or (eldoc-display-message-no-interference-p)
             ;; Erase the last message if we won't display a new one.
             (when eldoc-last-message
               (eldoc-message nil)
               nil))
         (eldoc-message text))))
```

josteink:
I think we should aim for something simpler, and avoid aliasing (which may cause issues in other modules!):

* Just a tide-specific function which dispatches to either `eldoc-display-message-no-interference-p` or `eldoc-display-message-p`
* We use this instead of the version-specific function in our code, where we previously used `eldoc-display-message-p` directly.
* We should probe for the existence of `eldoc-display-message-no-interference-p` using `fboundp` instead of checking Emacs-version, since version 28 can be any git snapshot with or without this function.

That should be more or less just 4 lines of code, and it should be safe across versions and packages, with minimal changes/risk to `tide` itself.

ramblehead:
It must be my old C++ thinking pattern kicks in :) Even there, with new C++20, I should be switching to constexpr if and concepts...

josteink:
Closed via linked PR.

PR #393, opened 2020-07-28
Fixing typo in tide-eledoc-display-message-p() name

ramblehead:
Should be tide-eldoc-display-message-p()

Sorry, I copy'n'pasted the function name with typo in my previous commit/pull request.

See #390


josteink:
Good catch. My bad for not noticing. 

Thanks for following up.

PR #403, opened 2020-10-30
add strikethrough property for deprecated candidate

ananthakumaran:
image

### DIFF
--- a/tide.el
+++ b/tide.el
@@ -1735,6 +1735,7 @@ This function is used for the basic completions sorting."
                                               "")))
                         (propertize name 'face '(:strike-through t))
                       name)))))
+
 (with-eval-after-load 'company
   (cl-pushnew 'company-tide company-backends))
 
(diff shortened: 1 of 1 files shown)


josteink:
Cool. LGTM!

￼￼Out of interest: are there any other properties we could extract and use as well?

Like being a recommended/most commonly used method/property (like we can see in VS full).

ananthakumaran:
> Out of interest: are there any other properties we could extract and use as well?

I just scanned through the realse announcement and found the [feature](https://devblogs.microsoft.com/typescript/announcing-typescript-4-0/#deprecated-support). There are others that might be useful if company supports icon feature like [vscode](https://code.visualstudio.com/docs/editor/intellisense#_types-of-completions)

> Like being a recommended/most commonly used method/property  (like we can see in VS full)

Could you point me to any docs or links? Not sure I understand the feature you are thinking about.

josteink:
> Could you point me to any docs or links?

Sorry. I was thinking about a Visual Studio feature which I think I've only seen with C#, and was conflating things.

PR #407, opened 2020-12-17
fix code edit handling on tide-rename-file

### LINKED ISSUE #406: Move file feature request / tide-rename-file bug (?)
There is `tide-rename-file`, but it does not work well when moving the file into a different directory. Would be dope if it did though. It would require:

- the moved file (TMF) must adjust its relative import paths.
- the files that import TMF must adjust their import paths too.

ananthakumaran:
tsserver used to? send code edits for the new file name. Looks like it
sends code edits for old file name now. If we get any edits for old
file, apply it before rename. Apply the rest of the code edits after rename.

fixes #406

### DIFF
--- a/tide.el
+++ b/tide.el
@@ -2027,7 +2027,7 @@ number."
     (tide-cleanup-buffer)
     (mkdir (file-name-directory new) t)
     (rename-file old new)
-    (rename-buffer new)
+    (rename-buffer (file-name-nondirectory new))
     (set-visited-file-name new)
     (set-buffer-modified-p nil)
     (tide-apply-code-edits after-rename-edits)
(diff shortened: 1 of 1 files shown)


ananthakumaran:
@Sleepful can you test this branch?

MarcusRiemer:
I also stumbled over this problem but deemed it a problem with `tsserver`. With this branch I have just moved a file to a different folder and had no issues. Thanks!

Sleepful:
Yup! It is working like a charm.

Sleepful:
I found a little bug, I renamed `/reports/money/Screen.tsx` to ` /moneys/Screen.tsx`

```
deleted    /reports/money/Screen.tsx
new file   /moneys/Screen.tsx
```

And it did this:
```
modified   /reports/money/Tooltip.tsx
-import { Row } from './Screen'
+import { Row } from '../../moneys/ScreenScreen'
```

it has double "Screen" on the new import

Sleepful:
It happens with a file rename without moving directory

Sleepful:
It seems to have something to do with sharing parts of the old name with the new name. It doesn't happen if names are very different.

Sleepful:
I am not able to replicate the issue again, this is so odd, don't mind me

PR #409, opened 2020-12-23
Fix tide-rename-file bug on new buffer name;

### LINKED ISSUE #408: Tide chokes when an open buffer shares the name of a renamed buffer
doing the following rename with `tide-rename-file` while there is an open buffer named "index" from a different file such as `/unrelated/index.ts`

```
old    /someFolder/index.ts
new    /anotherFolder/index.ts
```

it will rename the file just fine, but it will not be able to update imports

```
import { thing } from /someFolder/index.ts // no update to new path!
```

error on emacs buffer says: `apply: Buffer name 'index.ts' is in use`

If I close all buffers named `index.ts` before the rename, then the rename works fine.

However, if the renamed file is called `var/index.ts` and another file also called `foo/index.ts` imports it, then it will give an error when `tide` tries to modify the import in `foo/index.ts`, because tide will leave the buffer open for `var/index.ts`, and this gives the `already in use` error when tide tries to update `foo/index.ts`

## Possible solution?

Close buffers after modifying them, right now tide leaves them open.

Culprit is probably around here https://github.com/ananthakumaran/tide/blob/master/tide.el#L1307

## Update

I *think* I am close to the culprit, I think it is `find-file-noselect`, [related SO answer that mentions how this function does not kill the buffer](https://emacs.stackexchange.com/questions/2868/whats-wrong-with-find-file-noselect)

## Update Update

PR submitted.

Sleepful:
fixes #408:
If the new name of a buffer clashes with an opened buffer then
tide-rename-file stops and doesn't apply after-rename-edits.
Especially problematic because tide leaves buffers open as it applies
code edits.

### DIFF
--- a/tide.el
+++ b/tide.el
@@ -2027,7 +2027,7 @@ number."
     (tide-cleanup-buffer)
     (mkdir (file-name-directory new) t)
     (rename-file old new)
-    (rename-buffer (file-name-nondirectory new))
+    (rename-buffer (file-name-nondirectory new) t)
     (set-visited-file-name new)
     (set-buffer-modified-p nil)
     (tide-apply-code-edits after-rename-edits)


ananthakumaran:
Thanks for debugging the issue. I have uniquify buffer enabled, which seem to do this automatically, because I never ran into the problem while testing.

```
(require 'uniquify)
(setq uniquify-buffer-name-style 'forward)
```

ananthakumaran:
Thanks for debugging the issue. I have uniquify buffer enabled, which seem to do this automatically, because I never ran into the problem while testing.

```
(require 'uniquify)
(setq uniquify-buffer-name-style 'forward)
```

PR #410, opened 2021-01-04
Add tide-node-flags customizable variable

jomi-se:
Useful for large TypeScript codebases which need to set max-old-space-size to a higher value.

I have a large codebase that doesn't compile with the default value (😭 ) so the tsserver process crashes from time to time. This PR allows passing custom flags to the node process to change `max-old-space-size` (or whatever other flag).

Would you consider merging this?

Btw, massive thanks for tide 🙏

### DIFF
--- a/README.md
+++ b/README.md
@@ -266,6 +266,13 @@ paths are resolved against the project root directory.
 
 Name or path of the node executable binary file.
 
+### tide-node-flags nil `nil`
+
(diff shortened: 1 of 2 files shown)


josteink:
LGTM.

josteink (approved):

jomi-se:
Thank you!

Issue #333, opened 2019-09-04
tide-format-before-save slow, Sync request timed out format


mstorus:
Summary: tide-format-before-save is very slow in my repo, and times out when saving twice in a short period of time.

## Relevant Version Numbers

- Tide: 3.6.2
- TypeScript: 3.6.2
- Emacs: 26.2

## Steps to Reproduce the Bug

- Open a .ts source file in my repo
- enter a newline
- save the file
- enter another newline
- save the file

## Expected Behavior

I expected the 2nd save to complete quickly.

## Actual Behavior

After saving the 2nd time, emacs hangs on "Saving file ..." for about 7 seconds, and then reports the error
```
Debugger entered--Lisp error: (error "Sync request timed out format")
  signal(error ("Sync request timed out format"))
  error("Sync request timed out %s" "format")
  tide-send-command-sync("format" (:file "/Users/me/src/repo/filename.ts" :line 1 :offset 1 :endLine 21 :endOffset 1))
  tide-format-region(1 528)
  tide-format()
  tide-format-before-save()
  run-hooks(before-save-hook)
  basic-save-buffer(t)
  save-buffer(1)
  funcall-interactively(save-buffer 1)
  call-interactively(save-buffer nil nil)
  command-execute(save-buffer)
```

## More Details
It seems like the slowness only happens on every 2nd save. After the 2nd save times out, I'm able to save again (3rd try) and it completes quickly. Then if I attempt a 4th save, it times out again, and so on.

Note that invoking `tide-format` on a region behaves as expected (it completes quickly).

If I remove `(add-hook 'before-save-hook 'tide-format-before-save)` from my `.emacs` file, then I no longer experience any slowness upon save.

If I set `(setq tide-tsserver-process-environment '("TSS_LOG=-level verbose -file /tmp/tss.log"))`, then I get the following output in `/tmp/tss.log` after the 2nd save:

```
Info 1250 [19:2:9.95] Finishing updateGraphWorker: Project: /Users/me/src/repo/tsconfig.json Version: 15 structureChanged: true Elapsed: 7602ms                                                      
Info 1251 [19:2:9.95] Project '/Users/me/src/repo/tsconfig.json' (Configured)                                                                                                                        
Info 1252 [19:2:9.95]   Files (1644)  


... (list of files) ....


Info 1253 [19:2:9.95] -----------------------------------------------                                                                                                                                              
[...]

PR #413, opened 2021-02-10
remove tslint

ananthakumaran:
tslint tool is [deprecated](https://github.com/palantir/tslint#tslint) and the temp files created by it doesn't
play nice with tsserver as well.

see #412 #333


josteink:
LGTM.

PR #423, opened 2021-03-04
Add tide-eslint-disable-next-line

rymndhng:
This function is inspired by `tide-tslint-disable-next-line`.

See #308

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,6 +10,7 @@ jobs:
   build:
     runs-on: ubuntu-latest
     strategy:
+      fail-fast: false
       matrix:
         emacs_version:
           - 25.1
--- a/tide.el
+++ b/tide.el
@@ -1492,6 +1492,48 @@ always be formatted as described above."
                 (string-join error-ids " "))
         (typescript-indent-line)))))
 
+;;; Disable Eslint Warnings
+
+(defconst tide-eslint-disable-next-line-regexp
+  "\\s *//\\s *eslint-disable-next-line\\s *\\(.*\\)")
+
+(defun tide-add-eslint-disable-next-line ()
+  "Add a eslint flag to disable rules generating errors at point.
+
+This function adds or modifies a flag of this form to the
+previous line:
+
+  // eslint-disable-next-line [rule1], [rule2], [...]
+
+The line will be indented according to the current indentation
+settings.  This function generates rule1, rule2 to cover all the
+errors present at point.
+
+If the previous line does not already contain a disable-next-line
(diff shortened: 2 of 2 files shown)


rymndhng:
I'm uncertain what the cause of the test errors are on 25.1. The tests pass in the other emacs version, and it also works locally.

rymndhng:
@ananthakumaran thanks for trying to tweak that. It's really odd, any thoughts on what's going on?

josteink:
The error makes me think the build-error is caused by some unrelated change in an upstream dependency.

I've re-queued a new build for a previously green master-build here: https://github.com/ananthakumaran/tide/actions/runs/557567320

If that fails too, we can safely assume that our test-harness no longer is runnable on Emacs 25.1 without some additional intervention.

josteink:
Yup. That formerly good build fails too now. 

It’s the build which is broken, and probably not the PR.

josteink:
There's at least 2 angles we can attack this from. Or 2 theories:

From what I can tell, the difference between the 2 master-runs is that one succeeds at the Github actions task `purcell/setup-emacs@master`, while the other doesn't.

* Run 1 (Good): https://github.com/ananthakumaran/tide/runs/1648695854?check_suite_focus=true
* Run2 (Bad): https://github.com/ananthakumaran/tide/runs/2034324388?check_suite_focus=true

The one who fails, fails with with a HTTP 503 on an upstream build/artifact-dependency from `https://emacs-ci.cachix.org` and some Nix-stuff going on there, which I had no idea we depended on :smile: 

My guess is that this could be reported as an upstream bug for https://github.com/purcell/setup-emacs, and isn't an issue we can/should worry about right now. On the flip side, that sounds temporary and might resolve itself.

So theory number 2, which is my person favourite is this one:

When checking last successfully run build from git master, we see we're using `dash` version `2.17.0` during the test-phase, while these newer builds who are failing with invalid signature-checks are retrieving dash `2.18.0`. Dash has had backwards compatibility issues in the past, so if they've accidentally broken an old Emacs, it wouldn't be the first time. In which case, we should report it upstream to https://github.com/magnars/dash.el.

Either way, it does definitely *not* look like any bad changes have been introduced by this PR, so I see no reason to reject them over the build alone.

rymndhng:
@ananthakumaran @josteink Is there something I can do to help move this forward?

josteink:
It's unfortunate that the CI is (generally) unreliable, but since there have been no other objections I'm just going to merge this 🙂

josteink:
I re-ram the CI pipeline now and whatever issue was upstream is now fixed. A fully green merge 🙂

PR #424, opened 2021-03-27
Make no-server errors less naggy

elibarzilay:
Every once in a while I run into a situation where the server is either
dead or just misbehaving for whatever reason.  Since tide tends to send
a lot of messages to the server, this leads to an annoying state where
it keeps barfing over and over again.

This avoids throwing the no server error if it was just thrown.

---

I think that there are other situations where things get overly
annoying: I think that there are either other similar errors, and/or
other situations that lead to excessive nagging.  If so, then more
work will be needed, but I wanted to put it out now to see if
there's any opinions about doing this.

### DIFF
--- a/tide.el
+++ b/tide.el
@@ -653,19 +653,25 @@ Offset is one based."
     ((event) (tide-dispatch-event response))))
 
 (defun tide-send-command (name args &optional callback)
-  (unless (tide-current-server)
-    (error "Server does not exist. Run M-x tide-restart-server to start it again"))
+  (if (not (tide-current-server))
+    (let* ((message "Server does not exist, run M-x tide-restart-server to start it again")
+           (already-barfed
+            (with-current-buffer (messages-buffer)
+              (save-excursion (goto-char (point-max))
+                              (forward-line -1)
+                              (save-match-data (re-search-forward message nil t))))))
+      (unless already-barfed (error "%s" message)))
 
-  (tide-sync-buffer-contents)
+    (tide-sync-buffer-contents)
 
-  (let* ((request-id (tide-next-request-id))
-         (command `(:command ,name :seq ,request-id :arguments ,args))
-         (json-encoding-pretty-print nil)
-         (encoded-command (tide-json-encode command))
-         (payload (concat encoded-command "\n")))
-    (process-send-string (tide-current-server) payload)
-    (when callback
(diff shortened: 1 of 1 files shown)


tide.el
```diff
@@ -653,19 +653,25 @@ Offset is one based."
...
-         (json-encoding-pretty-print nil)
-         (encoded-command (tide-json-encode command))
-         (payload (concat encoded-command "\n")))
-    (process-send-string (tide-current-server) payload)
-    (when callback
-      (puthash request-id (cons (current-buffer) callback) tide-response-callbacks))))
+  (if (not (tide-current-server))
+    (let* ((message "Server does not exist, run M-x tide-restart-server to start it again")
+           (already-barfed
+            (with-current-buffer (messages-buffer)
```
ananthakumaran, line 659: could we get the same information via `(current-message)`

elibarzilay: That was my first attempt, but it looks like after an error is thrown
from the hook, it clears out the echo area --- and `current-message`
gets you just the *current* echo area.

ananthakumaran: Got it, then it seems like a decent approach. Two notes

1) we might want to move the error handling to a different method
2) Irrespective of whether we log error or not, we need to stop the execution. Earlier error would always be called, with the new change, we have to explicitly return and I don't see that in the current patch.

elibarzilay: 1. Yes, I completely agree.  I just started with a very focused error wrapper because I think that this was the main thing that was really bugging me (that is, when the server is broken), so I wanted to use it like this for a while to see if there are others.

2. Agreed: and in this patch I made it avoid sending the message & registering the callback when an error is not thrown.  I'm not sure I follow the second part of your comment though: there is no explicit return since that's not needed (since the actual work is in the else branch anyway).  BTW, this is also related to the above: if there's need to wrap other errors (and therefore generalize this change), then there might also be a need to use some `catch`ing to ensure that code that should be aborted is indeed skipped.

ananthakumaran: > so I wanted to use it like this for a while to see if there are others.

sure, feel free to merge when you are done with testing.

> there is no explicit return since that's not needed (since the actual work is in the else branch anyway

ignore my second point, looks like I missed the `if` at the top level

elibarzilay: I plan to take my patience with this so I can be sure that it does what I want it to do...

Issue #427, opened 2021-05-12
tide-format very slow 


bigteech:
```lsp
(require 'web-mode)
(add-to-list 'auto-mode-alist '("\\.tsx\\'" . web-mode))
(add-hook 'web-mode-hook
          (lambda ()
            (when (string-equal "tsx" (file-name-extension buffer-file-name))
              (setup-tide-mode))))
;; enable typescript-tslint checker
(flycheck-add-mode 'typescript-tslint 'web-mode
```
I used the config in tide's readme.   
Whenever I use  tide-format, my emacs  hung up.    
![image](https://user-images.githubusercontent.com/16490211/117937301-67720c00-b338-11eb-9c1b-71aa2abd1554.png)
the web-mode-fontify used 90% of the time;

ananthakumaran:
Refer [debugging](https://github.com/ananthakumaran/tide#debugging) section and find the response time for `format` command on tsserver. Since this is a synchronous command UI will get blocked till the format is done. If the tsserver is not taking much time, then please post the full profile report here.

bigteech:
### tsslog: 
![image](https://user-images.githubusercontent.com/16490211/118245914-e8b1d600-b4d3-11eb-9566-28521741066a.png)
### profiler screenshot:
![image](https://user-images.githubusercontent.com/16490211/118246014-0d0db280-b4d4-11eb-8ce9-60cc91c89a0d.png)
### profiler txt:
```txt
[...]

PR #428, opened 2021-05-14
Avoid triggering too many change hooks

ananthakumaran:
This commit attempts to reduce the number of times
`after-change-functions` is called when edits are applied. Otherwise,
with large enough buffer and complex font locking rules, emacs could
get blocked for long time.

Ideally the start and end of the region that is being changed should
be supplied to the function, but I am not sure how it could be
calculated easily as each edit affects successive edits' region.

For now, I have settled on using the min max point of the buffer and
apply the optimization only if there are more than two edits.

see #427

### DIFF
--- a/tide.el
+++ b/tide.el
@@ -1258,7 +1258,7 @@ Noise can be anything like braces, reserved keywords, etc."
   nil)
 
 (defun tide-eldoc-display-message-p()
-  (if (fboundp #'eldoc-display-message-no-interference-p)
+  (if (fboundp 'eldoc-display-message-no-interference-p)
       (eldoc-display-message-no-interference-p)
     (eldoc-display-message-p)))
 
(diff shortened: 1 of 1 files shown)


tide.el
```diff
@@ -2172,11 +2172,17 @@ code-analysis."
...
-(defun tide-apply-edits (edits)
+(defun tide-do-apply-edits (edits)
   (save-excursion
     (-map (lambda (edit) (tide-apply-edit edit))
           (nreverse edits))))
 
+(defun tide-apply-edits (edits)
+  (if (and (fboundp 'combine-change-calls)
+           (> (length edits) 2))
+      (combine-change-calls (point-min) (point-max) (tide-do-apply-edits edits))
```
josteink, line 2183: Could we improve this further by making more accurate bounds for this combined update?

josteink (approved):
A tiny nitpick or suggestion, but otherwise it looks good to me 🙂

tide.el
ananthakumaran: As explained in the commit message, it's not straightforward to calculate the region beforehand (unless we do two pass - first to calculate the regions, second for the actual changes). The main complexity is that each **edit** can introduce changes that will affect the next **edit's** region. We can calculate it easily after applying the edits, but the emacs API requires it beforehand.

At this point, I haven't spent much time on this though. If this seems to become the next bottleneck, probably we can try to optimize.

josteink: Yeah that makes sense. Fair enough.

Issue #341, opened 2019-10-22
Disable auto saving on action


ghost:
Is there any way of disabling auto saving functionality when I import an class from completions or when I rename an symbol ? Its pretty annoying that it auto saves and repositions my cursor every time I import an file or rename a symbol, I guess this happens for more actions.

ananthakumaran:
Currently, there is no way to disable save. Save also triggers flycheck syntax check.

We wrap the buffer edits with `save-excursion`, so ideally it should restore the cursor back to old position. But seems like it moves a bit

d4ncer:
Any update on this? I vastly prefer `tide` to the `lsp`, but this is the one pain point. I'm happy to help brainstorm/implement a solution for this, if it's a change you're happy with @ananthakumaran.

ananthakumaran:
The issue is about auto save or cursor movement

d4ncer:
Autosave, specifically around selecting a completion for an external dependency, which adds - or modifies an existing - import statement and applies before-save hooks. I run `prettier` as part of my `before-save-hook`, which reformats the buffer and repositions the cursor. Disabling prettier fixes the cursor problem, but ideally I'd like a way to turn off auto save on edit to prevent the `before-save-hook` running.

d4ncer:
I've opened a PR to specifically handle configuration of auto save when applying code edits. I can modify it to include renaming symbols too, if it would be valuable.

ananthakumaran:
Thanks for the PR. One of the concerns I have is whether it would introduce undesired behavior. For example, if we rename a symbol and the buffer save triggers a sync operation that syncs the file data with tsserver.

You could test for some time with the flag enabled and if no issues are found, we could include the flag (with some warnings in the doc perhaps)

d4ncer:
Sounds good. I have the flag running locally (only for code edits, not for renaming symbols) - I'll test for the next few weeks and report back.

d4ncer:
I've run this flag - `tide-save-buffer-after-code-edit` - as `t` for a little under two weeks now and it has been a godsend. Haven't noticed a single issue. Any additional actions you want me to take to merge the associated PR?

ananthakumaran:
> I've run this flag - tide-save-buffer-after-code-edit - as t for a little under two weeks now 

@d4ncer you mean with `nil`?

d4ncer:
🤦🏽‍♂️ sorry yes, I meant nil

ananthakumaran:
[...]

PR #435, opened 2021-09-16
Configure auto save when applying code edits

d4ncer:
Adds a new custom variable `tide-save-buffer-after-code-edit` to configure auto-save behaviour when applying code edits.

The variable defaults to `t` so there is no impact to current users.


d4ncer:
Addresses a part of #341

PR #440, opened 2022-03-12
add refresh support to project error buffer

### LINKED ISSUE #439: tide-project-errors cannot be refreshed
## Relevant Version Numbers
- OS: Manjaro (up to date)
- Tide: 20220102.441 (v4.5.4)
- TypeScript: 4.6.2 (using es6 as configured in `tsconfig.json`)
- Node: v16.9.1
- Emacs: 27.2

## Steps to Reproduce the Bug
1. Open a TS file
1. `M-x tide-project-errors`
1. `g`

## Expected Behavior
The key bindings summary listed in `tide-project-errors-mode`'s help page says that `g` is bound to `revert-buffer`. I think that the expected action to happen when pressing `g` therefore is that the buffer be refreshed (show the latest errors and not display outdated errors).

Re-binding `g` to `tide-project-errors` does not refresh the buffer either (even though the buffer is associated with the directory of the file it was opened from which is located inside the project directory) because it seems like this command may only be called from within TS files.

Of course, any suggestions of what to bind to `g` to make the buffer refresh would be welcome.

## Actual Behavior
The following message is printed:
```
revert-buffer--default: Buffer does not seem to be associated with any file
```

ananthakumaran:
Project error needs to be requested in the context of a project
file. The initial buffer file name from which the command was issued
is captured as origin and later used when refresh is requested.

fixes #439

### DIFF
--- a/tide.el
+++ b/tide.el
@@ -939,8 +939,8 @@ Currently, two kinds of cleanups are done:
    `(:hostInfo ,(emacs-version) :file ,(tide-buffer-file-name)
      :formatOptions ,(tide-file-format-options) :preferences ,tide-user-preferences)))
 
-(defun tide-command:projectInfo (cb &optional need-file-name-list)
-  (tide-send-command "projectInfo" `(:file ,(tide-buffer-file-name) :needFileNameList ,need-file-name-list) cb))
+(defun tide-command:projectInfo (cb &optional need-file-name-list file)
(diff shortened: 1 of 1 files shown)


ananthakumaran:
@m1cm1c could you try this branch. should solve the refresh issue

m1cm1c:
@ananthakumaran Works like a charm! Thank you very much! :)

PR #443, opened 2022-04-29
Fix handling of timestamps in new `(TICKS . HZ)` format

duxovni:
Emacs is rolling out a new timestamp format, to be used by time-handling functions such as `current-time` and `time-subtract`. Tide assumes timestamps are in the old `(HI LO US PS)` format, and attempts to extract a number of seconds by directly pulling out an element from this list with `cadr`.  In order to properly work with the changing timestamp format, we need to use Emacs' provided functions for handling timestamps, rather than poking at the internals directly.

### DIFF
--- a/tide-tests.el
+++ b/tide-tests.el
@@ -125,7 +125,7 @@
 (ert-deftest seconds-elapsed-gets-calculated ()
   (let ((start-time (current-time)))
     (sleep-for 5)
-    (let ((elapsed (seconds-elapsed-since start-time)))
+    (let ((elapsed (tide-seconds-elapsed-since start-time)))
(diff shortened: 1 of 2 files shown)


duxovni:
(In case anyone else is searching for issues, this is the cause of the `Wrong type argument: listp, 1000000000` errors produced by some tide commands on very recent emacs versions.)

josteink:
As something of a compatibility concern, would you be able to extract this into its own function and write a test for that?

(We may also want to update the CI to include more recent Emacs-versions too, but I guess that’s another matter)

duxovni:
Good idea, was this the sort of thing you had in mind?

josteink:
Looks great. Consider it merged. 

Thanks for contributing! 🙂

PR #456, opened 2022-10-26
Update to eldoc 1.1.0

sandersn:
1. Setup of eldoc is slightly different: add-hook to eldoc-documentation-functions instead of making a local variable.
2. eldoc-message is now deprecated, and the function gets a callback to call instead.

This lets newer eldoc-based frontends work, like eldoc-overlay.

I don't know how widely supported eldoc 1.1.0 is, or how to tell. It ships with emacs 28. If it's not widely installed, then this PR is not a good idea.

### DIFF
--- a/tide.el
+++ b/tide.el
@@ -1263,14 +1263,14 @@ Noise can be anything like braces, reserved keywords, etc."
   (tide-fallback-if-not-supported "quickinfo-full" tide-command:quickinfo-full tide-command:quickinfo-old cb))
 
 
-(defun tide-eldoc-function ()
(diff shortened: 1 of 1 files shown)


ananthakumaran:
We try to support emacs that ships with the last 2 [debian](https://github.com/ananthakumaran/tide/wiki/Supported-Emacs-Versions) LTS versions. if we strictly follow that, we have to support 26+, if we say only the latest LTS, we have to support 27+.

My personal opinion is, it's too early to drop support for <28.1, since 29 is not even released.

Looking at the patch, it might be possible to support both the old and new versions with minimal changes.

PR #458, opened 2023-01-27
Treesitter support

kijowski:
This PR removes hard dependency on `typescript-mode` and adds support to treesitter based `typescript-ts-mode` (it keeps support for `typescript-mode`)

`typescript-mode` has been used directly only for indentation and after changing calls from `typescript-indent-line` to `indent-according-to-mode` we can safely remove hard dependency and support treesitter based modes at the same time.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,10 @@
 ;; formats the buffer before saving
 (add-hook 'before-save-hook 'tide-format-before-save)
 
+;; if you use typescript-mode
 (add-hook 'typescript-mode-hook #'setup-tide-mode)
+;; if you use treesitter based typescript-ts-mode (emacs 29+)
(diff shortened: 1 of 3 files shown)


kijowski:
I can see that CI build for 27.2 failed, but it looks like there was a problem with melpa connectivity. I think that after rerunning it there should not be any errors.

josteink (approved):
LGTM

sandinmyjoints:
When I switch to typescript-ts-mode, tide-company no longer has any completions. Is anyone else seeing this problem?

josteink:
Have you added the new hook as outlined in the readme?

sandinmyjoints:
Yes, `tide-setup` is in my `typescript-ts-mode-hook`, and I can see that `tide-mode` is enabled in my TS buffers. Company as well -- I get completions from other backends.

PR #459, opened 2023-01-28
Update to eldoc 1.1.0 with support for emacs < 28.1

kijowski:
This PR is a continuation of work done in https://github.com/ananthakumaran/tide/pull/456 

It checks if emacs version is less than 28.1 (where `eldoc` 1.1 got introduced) and uses the old mechanism in that case.

### DIFF
--- a/tide.el
+++ b/tide.el
@@ -1262,16 +1262,19 @@ Noise can be anything like braces, reserved keywords, etc."
 (defun tide-command:quickinfo (cb)
   (tide-fallback-if-not-supported "quickinfo-full" tide-command:quickinfo-full tide-command:quickinfo-old cb))
 
-
 (defun tide-eldoc-function (&optional cb)
-  (unless (member last-command '(next-error previous-error))
-    (if (tide-method-call-p)
-        (tide-command:signatureHelp (lambda (text) (funcall #'tide-eldoc-maybe-show text cb)))
-      (when (looking-at "\\s_\\|\\sw")
-        (tide-command:quickinfo
-         (tide-on-response-success-callback response (:ignore-empty t)
-           (tide-eldoc-maybe-show (tide-doc-text (plist-get response :body)) cb))))))
-  nil)
+  (cond ((member last-command '(next-error previous-error))
+         nil)
+        ((tide-method-call-p)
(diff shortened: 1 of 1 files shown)


tide.el
```diff
@@ -1263,14 +1263,14 @@ Noise can be anything like braces, reserved keywords, etc."
   (tide-fallback-if-not-supported "quickinfo-full" tide-command:quickinfo-full tide-command:quickinfo-old cb))
 
 
-(defun tide-eldoc-function ()
+(defun tide-eldoc-function (&optional cb)
   (unless (member last-command '(next-error previous-error))
     (if (tide-method-call-p)
-        (tide-command:signatureHelp #'tide-eldoc-maybe-show)
+        (tide-command:signatureHelp (lambda (text) (funcall #'tide-eldoc-maybe-show text cb)))
```
ananthakumaran, line 1269: funcall is not necessary here, we can just call directly

tide.el
```diff
@@ -1263,14 +1263,14 @@ Noise can be anything like braces, reserved keywords, etc."
   (tide-fallback-if-not-supported "quickinfo-full" tide-command:quickinfo-full tide-command:quickinfo-old cb))
 
 
-(defun tide-eldoc-function ()
+(defun tide-eldoc-function (&optional cb)
```
ananthakumaran, line 1266: > If the computation of said doc string (or the decision whether
  there is one at all) is expensive or can’t be performed
  directly, the hook function should return a non-nil, non-string
  value and arrange for CALLBACK to be called at a later time,
  using asynchronous processes or other asynchronous mechanisms.

As per the doc, this function is supposed to return a non-nil if we plan to use the callback. This seems to happen by chance because tide-send-command returns the callback value (again by chance of the final put call). I think we need to make it more explicit by returning true `t`. What do you think?

kijowski: Sorry I've missed that part in the docs. I agree about having explicit return value - I've reworked the `tide-eldoc-function` a bit to simplify conditional routing at the same time.

tide.el
kijowski: I've removed unnecessary funcall

ananthakumaran (approved):

PR #465, opened 2023-05-21
Enable flycheck checker on .js and .ts files

vedant-influencer-garage:
The flycheck checker runs only on `.tsx` and `.jsx` files, but not all typescript files end with that extension. Files in the project without react components(HTML markup) in that end with plain `.ts` (eg. utility helpers, hooks) also need to be checked.

### DIFF
--- a/tide.el
+++ b/tide.el
@@ -2545,7 +2545,7 @@ current buffer."
   :modes '(web-mode js2-jsx-mode rjsx-mode)
   :predicate (lambda ()
                (and
-                (tide-file-extension-p "jsx")
+                (or (tide-file-extension-p "jsx") (tide-file-extension-p "js"))
                 (tide-flycheck-predicate))))
 
 (add-to-list 'flycheck-checkers 'jsx-tide t)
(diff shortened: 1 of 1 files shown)


ananthakumaran:
Tide already has 2 other flycheck checkers defined for typescript and jsx. Please check the code above your changes

Chipe1:
@ananthakumaran Didn't notice these checkers earlier but looks like they are not enabled for web-mode. What's the expected workflow where tide would work on both `.ts` and `.tsx` files of a project? Should it be configured to use web-mode only on `.tsx` files and typescript-mode on `.ts` files?
I use a single mode (web-mode) for all files (ts, tsx, js, jsx, html, css) with single config/shortcuts and I feel most people would also stick to a single mode. `.ts` is a subset of `.tsx` so enabling the checker would not cause any issues and `tide` would work seamlessly with `web-mode` on `.js` and `.ts` files with the config mentioned in README. Thoughts?

ananthakumaran:
> Should it be configured to use web-mode only on .tsx files and typescript-mode on .ts files?

This is the typical workflow (I use as well) used by others.

The flycheck checkers follow a naming pattern {language}-{framework}, so javascript-tide should get enabled on javascript file etc. Instead of changing the jsx/tsx, javascript and typescript checkers should be updated to accommodate the use case.

We need to add web-mode in the allowed modes. The tricky part is, if it's web mode, we should only enable it on a subset of files (js/ts), so that should be checked as well (only if it's web mode).

Chipe1:
Got it, thanks for for taking the time to explain the workflow. Really awesome package, makes my life a lot easier. Cheers!

Chipe1:
I don't see an option to close the PR for some reason. Can you please go ahead and close it.

